Issue #14, opened 2023-12-09
Class structure for the cards


Turtyo:
## Feature Description

What should be the structure of the classes for the cards to be able to have multiple effects, such as damage and poison or damage and heal for example ?

## Motivation

This issue is meant as a way to discuss our structure going forward. Right now, `CardBase` handles most of that, but as a parent class of other effect, it shouldn't have functions that are not useful to multiple children classes (for example, it shouldn't have a `drawCard` or `Heal` function in it, those should be defined in specific children classes.

## Acceptance Criteria

- `CardBase` should hold function that are useful to multiple children classes, and not specific effects
- It should be easy to add effects to a card

## Proposed Solution

We would have:
```GDscript
# CardBase.gd
extends Node

# Base class for cards
class_name CardBase

var card_effects = []

func _ready():
    # Initialize a new array for each instance
    card_effects = []

func apply_all_effects(card):
    for effect in card.card_effects:
        effect.apply_effect()
```
with:
```GDscript
# CardEffect.gd
extends Node

# Base class for card effects
class_name CardEffect

fund _ready(card):
    # to be used by child classes to add their effect to the card
    card.card_effects.append(self)

func apply_effect():
    pass
```
and the effects would extend from `CardEffect`:
```GDscript
# CardPoison.gd
extends CardEffect

class_name CardPoison

# @Override
func apply_effect():
    # do some poison damage
```

When you want to create a card, you can either use a JSON format like this:
```json
[
  {
    "name": "CardWithPoisonAndHeal",
    "description": "Apply 3 poison and heal 3",
    "effects": [
      {
        "type": "CardPoison",
        "parameters": {
          "duration": 3,
          "damage_per_turn": 3
        }
      },
      {
        "type": "CardHeal",
        "parameters": {
          "heal_amount": 3
        }
      }
    ]
  }
]
```

And we would have some code to display the card / setting the effects in the background by reading the file
But i would like to not have that be the structure, because JSON files are not very nice to work with. For example i could say 3 poison in the description but put 4 by mistake in the duration.

Ideally, it would be good to have something that's directly editable. For example we could have something in the text file of the card that reads:
"Apply PoisonEffect(3) and HealEffect(4)". That would setup a card with 3 poison and 4 heal, but when displayed it would read as "Apply 3 poison stacks and heals 4". This could probably be done by using "to_str" and by using a parser. This probably should be addressed in another issue

## Additional Context

The discussion was started in #11

[...]

Issue #7, opened 2023-12-06
Use an object to store modifiers associated with a type of card / item effect


Turtyo:
## Feature Description

An object that holds the modifications to the state of the game that change the numerical values of effects. Everything modifies and checks this object when the need to make a calculation that can be impacted by buff / debuff arises.

## Motivation

This allows for a centralized modifier checking system when calculating damages. If I have an item that allows me to `heal +5` on every heal, every card that heals me shouldn't have to know that I have said object. Instead the item modifies this object that holds the changes, and every card then checks the same dictionary to know if there is a modification.

## Acceptance Criteria

- The object can be modified by buff / debuff
- The object can hold new types of effects easily
- The object holds information for both permanent changes (changes due to objects that stay after the end of a combat) and changes from temporary effects (that end after a set number of turns / end of combat)

## Proposed Solution

A dictionary of effects and the associated buffs / debuff that goes with them. For example, if we have `slash` damage and `piercing` damage:
```
{
    "slash" :  { Add: 10, Multiply: 2}
    "piercing" :  { Add: -15, Multiply: 1}
}
```
Would make a card that deals 10 `slash` damages and 20 `piercing` damages a total of `(10 + 10)*2 = 40 slash` damages  and `(20 - 15)*1 = 5 piercing` damages.

The problem with the previous implementation is that it doesn't differentiate between permanent and temporary status changes. This can be solved by having the following structure for the value associated to a given key:
`(Add permanent: value_1, Add temporary: value_2, Multiply permanent: value_3, Multiply temporary: value_4)`
Where the temporary values would be reset to 0 after each combat and the permanent values would only change if something adds or removes an item that changes this value.

## Additional Context

This overrules [Use a `Dictionary` to store buffs instead of an `Array`](https://github.com/Saplings-Projects/1M_sub/issues/5) by being a more general approach to the problem of storing buffs / debuffs. This is also in link with [Make stats changes related to character instead of calculating them at attack time](https://github.com/Saplings-Projects/1M_sub/issues/4)

Mikotey:
Motivate:
> when calculating damages.
Just a nitpick. It's not exclusively for damage types. They could be used for anything including increasing summons spawned to increasing the # of items you get as a reward

Proposed Solution:
> A dictionary of effects and the associated buffs / debuff that goes with them
[...]

Issue #2, opened 2023-12-01
Add a Status class to be parent of Buff/Debuff


Turtyo:
**Is your feature request related to a problem? Please describe.**
Right now the debuff and buff are all under the same `BuffBase` class. This can lead to confusion as we might not know later what we are doing. We might also want to differentiate between the two for a question of priority / UI design.

**Describe the solution you'd like**
Creating a parent class called `Status` or `StatusEffect`, and make both `Buff` and `Debuff` extend this class. Most of code would be in the `Status` class since for the coding part of it, they do pretty much the same thing (they have an effect on the characters)

**Describe alternatives you've considered**
An alternative would have been to keep the `BuffBase` class and having a `bool is_debuff`. The issue with this is that it's probably less clear than having two distinct classes. It also doesn't allow for static typing checks at compilation.

**Additional context**
For the UI, it was suggested to have debuffs with a red border and buffs with a green border as an indicator (not clear if this is a flashing effect on the border of the screen when the effect is applied, or if this is the color of card border which gives this status).
This has been discussed as something that could be done with [this PR comment](https://github.com/Saplings-Projects/1M_sub/pull/1#discussion_r1412344670) and was decided to be implemented in another PR, since this is out of the scope of the initial PR.

cheesefrycook:
Makes sense

PR #8, opened 2023-12-06
Implement a github action for future tests with GUT

Turtyo:
# Description

This PR introduces a short CI.yml to automatically run tests that are in `tests/unit` with GUT.

## Related issue(s)

This PR doesn't have a related issue as it was started before the changes requiring an issue. Approved by project owner.

## List of changes

A more detailed list of the changes.
- added `CI.yml` in `.github/workflows/`
- added test in `tests/units/`

## Additional notes

Tests are currently in `Tests` and will need to be moved to `tests/unit` for it to work.

PR #16, opened 2023-12-09
Add a Status class to be parent of Buff/Debuff

Turtyo:
# Description

This pull request adds a Status class to be able to differentiate between Buff and Debuff types.

## Related issue(s)

#2 

## List of changes

-Changed BuffBase to StatusBase
- Created 2 folders for Buff and Debuffs
- Modified all buff and debuffs to have the correct name / extension / variable names

## Tests

No new tests provided.

## Additional notes

The specific functions inside those classes will probably change due to #14 and #7 but the De/Buff structure will probably stay, so this issues is still relevant.


Buffs/BuffBase.gd
Tomzkk, line 1: Putting this comment here since I am not sure how to add a comment on file that's not been changed but don't we also need to change buffcomponent ?

Tomzkk: Also some other places like cardbase which has buffs_to_apply which should now probably be statuses_to_apply, ignore_attacker_buffs and ignore_victim_buffs in DealDamageData ....
Also BuffBase is directly referenced in other places, such as HealthComponent, debuffs are still erferenced as buff_poison etc in tests for example

Tomzkk (commented):
Added some comments of things that might have been missed ? 

Additionally wondering about the additional context in the issue behind this -> if we wanted to do something that all buffs with share (like the mentioned UI border which changes depending on whether it's a buff or debuff would we need to add that to each class or is there some way in godot to determine if class is from buff folder ? 
Same if we for example wanted to get all buffs from a character (only buffs, not debuffs) - would we be able to do it like this ?

Buffs/BuffBase.gd
cheesefrycook: Other than the stuff Tom mentioned, it looks good to me. Should we go ahead and make the BuffBase and DebuffBase or address it in a future PR?

cheesefrycook (approved):

Buffs/BuffBase.gd
Turtyo: I will make it right now, so we don't need to touch it again after

Turtyo:
There was a lot more stuff to modify than i expected. Last time I thought the scene was still running fine, but I had forgotten to reload Godot. So it turns out the scene wasn't loading anymore due to resources not loading.
Had to modify resources files.
All tests are passing now, but this could have been checked much more rapidly with #8 (or with me just running the tests, but you get the point...)

I'll do the modification for the buff/debuff class now and we should be good to go.

cheesefrycook (approved):

Saphust (approved):

Issue #12, opened 2023-12-07
Deck animation and layout


cheesefrycook:
## Feature Description

Add a dealing and discarding animation for cards. Add fan-shaped layout for cards. Add animation for hovering cards.

## Motivation

I feel it's important to get card animation integrated early. Since the animation will have cards added and discarded one by one, this could lead to issues where we're trying to access cards that aren't properly initialized or destroyed yet.

The fan-shaped layout is a visual change. It will make large hand sizes easier to see.

## Concept art / Designs

Proposed card layout from Slay the Spire.
<img width="596" alt="Screenshot 2023-12-07 172524" src="https://github.com/Saplings-Projects/1M_sub/assets/151570126/ef5d0c28-a2f9-4f61-a1a2-226dc351f9b2">

## Acceptance Criteria

- Cards are dealt in a fan shape
- Cards are dealt one by one and discarded one by one
- Interacting with cards during dealing and discarding does not result in unexpected states
- Hovering a card scales it up and moves other cards out of the way

## Proposed Solution

Cards can be dealt and removed on a looping timer. When you want to add or discard a card, it can be added to a queue to be added/removed on the next timer tick. For discarding, adding to a queue prevents the card from remaining in the hand after the card is already played.

Cards have a `CardMovementComponent` which is a state machine. Each state can have its own movement code. `CardMovementComponent` has a `desired_position` which is set from the `CardContainer`. States can also be switched from the `CardContainer`. We want to do this in the `CardContainer` because it has a list of all the cards in the player's hand. This allows us to set the `desired_position` of the cards relative to all other cards.

The fan-shaped layout can use curves with the card's input x-position to modify either rotation or y-position.

## Implementation Steps (Optional)

- Implement `CardMovementComponent`
- Implement card drawing and discarding queue
- Implement fan-shaped layout
- Implement card hover states

## Additional Context

Dependent on #11

Turtyo:
What would the movement of the cards be ? Straight line from the draw to the hand and straight line from the hand to the discard ? I think it's a good idea to start with a simple motion like that.
How do you plan to make `CardMovementComponent` ?
I think timers for draw and discard should probably be different (discard faster than you draw ?)

cheesefrycook:
- Yes a simple animation like that to start with. Maybe also scale the card up from zero.
[...]

Issue #11, opened 2023-12-07
Draw pile and discard pile


cheesefrycook:
## Feature Description

Add a draw pile and discard pile. Add global drawing and discarding of cards. Allow CardBase to draw and discard cards.

## Motivation

There is no concept of draw pile and discard pile. Additionally, there is no way to have cards drawn or discarded from other cards. For example, a card with the effect "Draw 2 cards"

## Acceptance Criteria

- CardBase allows drawing and discarding of cards via parameters (amount_of_cards_to_draw, amount_of_cards_to_discard)
- When cards are played, they are added to a discard pile
- When the draw pile is empty, the discard pile is shuffled into the draw pile
- Cards are put into the player's hand from the draw pile

## Proposed Solution

The draw pile and discard pile are stored in CardContainer as arrays of CardBase, NOT CardWorld. When ready to draw we get the top CardBase from the draw pile array and initialize a new CardWorld with that data.

Modify CardManager to hold a reference to the current CardContainer. If you wish to draw cards from somewhere, call `CardManager.card_container.draw_cards()`

Turtyo:
> Add global drawing a discarding of cards

Little typo maybe ?

- Why have `CardBase` be the one that holds the ability to draw cards ?
- How many draw by turn ? Can this value be modified by the player in some way ?
- The action to shuffle the discard pile in the draw pile should be possible when implementing (for example you draw 2 but you have only 1 left in draw pile)
- Someone mentioned having the discard pile be called `compost`, I like the idea, what do you think (ofc we won't call it like that in the code, but for a visual indication it could be nice)

cheesefrycook:
Which typo are you referring to? 
- Yes, my idea is CardBase should allow drawing of cards. For example, a card with the effect "Draw 2 cards". We could do this in children of CardBase, but I feel card drawing will be something that a lot of cards will want, so it may be easier to do in the parent like I'm suggesting here. So we have params on the CardBase like `amount_of_cards_to_draw` `amount_of_cards_to_discard` that can be easily set. Similar to how CardBase can deal damage right now with the `damage_to_deal_to_target`
- My idea is to start with 5 cards drawn per turn
- I think I made a typo on the original. The discard pile should shuffle into the draw pile when there are no cards left to draw.
- Love the idea!

Turtyo:
"Add global drawing a discarding of cards" doesn't really make sense to me, I would've expected "Add global drawing **and** discarding of cards" for example

[...]

PR #17, opened 2023-12-11
Add draw pile and discard pile

cheesefrycook:
# Description

Adds a draw pile and discard pile.

## Related issue(s)

#11 

## List of changes

- Add draw pile and discard pile to CardContainer
- Add drawing and discarding of cards to CardBase
- Add RandomGenerator which sets a seed for deterministic randomness. Used for randomness when drawing cards.
- Modify CardManager to just hold a reference to the current CardContainer
- Add draw pile and discard pile to the UI

## Tests

I'm going to wait until I implement #12, because it will change how drawing and discarding tests work a bit.


Turtyo:
This one might get a merge conflict due to #16 since I modified some files you modified too, we'll need to see once the other one is merged.
Will review this tomorrow !

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
+
+func _create_card_in_world(card_data: CardBase):
+	var card_instance: CardWorld = card_scene.instantiate()
+	add_child(card_instance)
+	cards_in_hand.append(card_instance)
+
+	card_instance.init_card(card_data)
+	
+	# bind mouse events
+	var card_click_handler: ClickHandler = card_instance.get_click_handler()
```
Turtyo, line 133: maybe this could be a helper function if we need to reuse these 4 lines

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
+	discard_pile.append(card.card_data)
+	
+	# remove from world
+	cards_in_hand.remove_at(card_index)
+	card.queue_free()
+	
+	on_card_counts_updated.emit()
+
+
+func _discard_last_card() -> void:
```
Turtyo, line 153: we should be careful of how we use this function later on. For now it's only used to discard all cards, but it could create some issues if not handled properly, since we append cards but we also remove cards from last index (we could end up removing a card we just added)

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
 	if new_phase == Enums.Phase.PLAYER_ATTACKING:
-		deal_cards()
+		deal_to_hand_limit()
 	if new_phase == Enums.Phase.ENEMY_ATTACKING:
 		discard_all_cards()
 
 
 func _on_card_clicked(card: CardWorld) -> void:
-	if CardManager.is_card_queued():
+	if is_card_queued():
```
Turtyo, line 166: will check this function again tomorrow with more sleep, because I feel like there is probably a simpler way to write that.

Cards/CardBase.gd
```diff
@@ -39,6 +41,11 @@ func _on_card_play(caster: Entity, target: Entity) -> void:
 	# TODO add other functionality that lots of cards may share (eg: restore AP, draw cards)
 
 
+func on_post_card_played():
```
Turtyo, line 45: what is the goal of this ?

Cards/CardBase.gd
```diff
@@ -82,3 +89,11 @@ func _apply_buffs(caster: Entity, target: Entity) -> void:
 				party_member.get_buff_component().add_buff(buff, caster)
 		else:
 			target.get_buff_component().add_buff(buff, caster)
+
+
+func _draw_cards():
+	CardManager.card_container.draw_cards(amount_of_cards_to_draw)
+
+
+func _discard_cards():
```
Turtyo: maybe this should be called "discard_random_cards" as opposed to a function that would let you choose which cards to discard for example

Core/Battler.gd
```diff
@@ -68,6 +68,9 @@ func _on_enemy_start_turn() -> void:
 		var success: bool = _on_attack(enemy.get_behavior_component().attack, enemy, PlayerManager.player)
 		assert(success == true, "Enemy failed to attack.")
 	
+	# TODO: temporary delay so we can see the draw pile and discard pile working
+	await get_tree().create_timer(1.0).timeout
```
Turtyo, line 72: timer time should be a variable so it's more easily accessible

Managers/RandomGenerator.gd
```diff
@@ -1,13 +1,9 @@
...
+## Initializes randomness for the game.
 
 
 func _ready() -> void:
-	random = RandomNumberGenerator.new()
-	random.seed = randi()
-	
-	print("Random seed set to: " + str(random.seed))
+	var seed: int = randi()
+	seed(seed)
```
Turtyo, line 7: does this set the seed globally for all random generator instances ? or just in a given scope ?

Turtyo (commented):
After reading the code, i'm pretty sure there will be merge issues with things in #16 but it should not be too hard to choose I think (but there will probably be quite a lot of them).

Cards/CardBase.gd
cheesefrycook: This is so a card can draw or discard cards. This will change if we implement the CardEffect system mentioned in #14

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
+func remove_queued_card() -> void:
+	discard_card(queued_card)
+	set_queued_card(null)
+
+
+func is_card_queued() -> bool:
+	return queued_card != null
+
+
+func deal_to_hand_limit() -> void:
```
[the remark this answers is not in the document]
cheesefrycook: Good call, the name of this should probably be deal_to_starting_hand_size

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
+		# limit amount of cards to a maximum
+		if cards_in_hand.size() < max_hand_size:
+			_draw_card()
 
 
 func discard_all_cards() -> void:
-	for card: CardWorld in cards:
-		card.queue_free()
-	cards.clear()
+	while cards_in_hand.size() > 0:
```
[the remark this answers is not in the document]
cheesefrycook: Like the comment mentions, if we do it like that then the for-loop is going to have elements removed while it's iterating - which causes issues

Cards/CardContainer.gd
cheesefrycook: Yeah true. I can add _bind_card_input

Cards/CardContainer.gd
cheesefrycook: Don't think we will use it for anything except discarding all cards unless we end up needing it elsewhere internally. Externally you shouldn't ever call this

Cards/CardContainer.gd
cheesefrycook: Yeah, probably so. There's still some issues with card input as well if you spam click. I wanted to open a new issue related to this, where I also wanted to ask if we'd prefer a click or drag and drop input system.

Core/Battler.gd
cheesefrycook: Yeah fair. This is gonna get removed whenever we add enemy attack animations, but I can add the variable for now

Managers/RandomGenerator.gd
cheesefrycook: This is for the entire game. So every random thing is going to work off of this seed. We probably won't want this in the long run, because I imagine we're going to want seeded randomness for stuff like map generation, but unseeded randomness for stuff like random card shuffling. This is helpful for now though because we can reproduce random stuff while debugging

Managers/RandomGenerator.gd
```diff
@@ -0,0 +1,13 @@
+extends Node
+
+
+var random: RandomNumberGenerator = null
```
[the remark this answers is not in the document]
cheesefrycook: Seems like this comment is on an outdated version. This isn't being used anymore. Instead seeded randomness is getting set for the entire project.

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
+var queued_card: CardWorld = null
 
 
 func _ready() -> void:
 	PhaseManager.on_phase_changed.connect(_on_phase_changed)
-	CardManager.successful_card_play.connect(_on_card_successful_play)
+	CardManager.set_card_container(self)
+	
+	draw_pile = default_deck.duplicate()
+	draw_pile.shuffle()
```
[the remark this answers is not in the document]
cheesefrycook: See my comment below about the RandomGenerator. This should be seeded randomness with the current implementation.

Managers/CardManager.gd
```diff
@@ -1,22 +1,12 @@
...
-
-
-func set_queued_card(card: CardWorld) -> void:
-	queued_card = card
-
-
-func notify_successful_play() -> void:
-	successful_card_play.emit(queued_card)
-	set_queued_card(null)
+func set_card_container(in_card_container: CardContainer):
```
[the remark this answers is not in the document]
cheesefrycook: This doesn't add cards to the deck no, this is just used to set the CardContainer in the global CardManager AutoLoad class. So on ready, CardContainer calls `CardManager.set_card_container(self)` so you can then access the CardContainer via `CardManager.card_container`. This is the same type of thing as the `PlayerManager.player` pattern

Cards/CardContainer.gd
```diff
@@ -10,93 +10,184 @@ class_name CardContainer
...
 
 
-func remove_card(card: CardWorld) -> void:
-	var card_index: int = cards.find(card)
-	cards[card_index].queue_free()
-	cards.remove_at(card_index)
+func discard_random_card(amount: int) -> void:
+	var valid_cards_to_discard: Array[CardWorld] = []
+	
+	for card in cards_in_hand:
```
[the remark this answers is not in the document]
cheesefrycook: Yeah true, that's probably cleaner

Cards/CardBase.gd
cheesefrycook: Good call, will change

Paper-2 (commented):
Should we add a menu for the draw/discard pile now or later?

Cards/CardContainer.gd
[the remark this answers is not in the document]
cheesefrycook: Just realized we don't even need to check the queued_card anymore since we're calling discard_random_card from `on_post_card_played()` in the CardBase. Just pushed a change that simplifies even further

cheesefrycook:
You mean like a UI popup that shows all the cards in your deck when you click it? That's a good idea. Don't think we should implement it in this PR, but maybe one for the near future?

Paper-2 (commented):
> You mean like a UI popup that shows all the cards in your deck when you click it? That's a good idea. Don't think we should implement it in this PR, but maybe one for the near future? 

Yeah an UI. I guess this PR is already almost done and trying to do something else would make it a mess.

Cards/CardBase.gd
cheesefrycook: So for example, say we're at our hand limit of 10 cards. We play a "Draw 2 cards" card. What should happen is the card should discard and then we draw a new card. Then we will hit the hand limit of 10 cards. Without the `on_post_card_played`, the card we played won't be discarded yet. So we try to draw a new card before the queued card has been discarded. It will see that the hand is already full, and not draw the card.

cheesefrycook: Now that I think about it, I think we can just rearrange the logic a bit so we only need one `_on_card_play()` that fires after the card discards. We just need to add a new `can_play_card()` that we check in the Battler first

Managers/RandomGenerator.gd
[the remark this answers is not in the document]
Turtyo: Yes I saw that in later commits

Cards/CardContainer.gd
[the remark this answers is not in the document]
Turtyo: This seems to only be an issue because we use _discard_last_card() instead of discarding a specific card. This also wouldn't be an issue if we decided to discard the first card instead of discarding the last card (because the new Array will have  the correct first element)

Managers/CardManager.gd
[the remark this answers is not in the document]
Turtyo: Maybe adding a short comment for this would still be good

Cards/CardBase.gd
Turtyo: We can just have the card be discarded before its effects are played maybe ?

cheesefrycook: Yeah, my latest commit does this

Managers/RandomGenerator.gd
Turtyo: The seed is randomly set here, is there a way for us to decide the seed at launch ? Because we can't give a seed to the function so it sets this seed, it always uses `var seed: int = randi()`

Cards/CardContainer.gd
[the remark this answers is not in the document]
cheesefrycook: My understanding is that for-each-loops shouldn’t ever change the size of the thing they’re iterating. Since discard_card removes an index, I was running into issues when using a for-loop. I guess I can do _discard_first_card instead if that’s preferred? Although I feel like we want to discard from the end for animation purposes later

Managers/RandomGenerator.gd
cheesefrycook: Yeah, we could just replace randi() with whatever number we want. I can add logic to the RandomGenerator that does this if we think it’s useful

Core/Battler.gd
```diff
@@ -66,8 +66,13 @@ func _on_enemy_start_turn() -> void:
 	
 	# enemy attack
 	for enemy: Entity in _enemy_list:
-		var success: bool = _on_attack(enemy.get_behavior_component().attack, enemy, PlayerManager.player)
-		assert(success == true, "Enemy failed to attack.")
+		var enemy_attack: CardBase = enemy.get_behavior_component().attack
```
Turtyo, line 69: this seems to be out of the scope of the PR and the related issue

cheesefrycook: Yeah true. This needed to change because I changed the way cards are played, which also affects enemy attacks since they use a CardBase to attack. I can revert the change if you’d like, we’ll have to fix enemy attacks in a different PR though

Turtyo: leave it like that then

Turtyo:
I resolved the two merge conflicts, you should also pull it in local to be make sure everything is alright

cheesefrycook:
Should be good now, had to fix some loading issues

Cards/CardContainer.gd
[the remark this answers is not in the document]
Turtyo: Yeah you are right, the for loop shouldn't change the size of the data it is dealing with. Let's keep it like that.

Cards/CardBase.gd
Turtyo: Which commit are you talking about btw ?

Managers/RandomGenerator.gd
Turtyo: Well that would imply having to modify the code, it's probably better to have a structure like:
`func(given_seed: int = -1) -> void:
    var set_seed: int = given_seed
    if set_seed == -1:
        set_seed = randi()
    seed(set_seed)
    # print the set seed here
```
So `given_seed` defaults to -1 and we can give it a seed to put the one we want.

File: Managers/CardManager.gd (hunk 12 above)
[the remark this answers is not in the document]
Comment (Turtyo) [reply Managers/CardManager.gd]: I didn't see a comment for this in the latest commit but I might have missed it. If it's not there, could you add one ?

**Comment by Turtyo** [comment]
I was thinking it would be good to have tests for this PR even if they will be changed for the next PR related to this. In order to confirm the code at least.

File: Managers/CardManager.gd (hunk 12 above)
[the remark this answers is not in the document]
Comment (cheesefrycook) [reply Managers/CardManager.gd]: I'll add it

File: Managers/RandomGenerator.gd (hunk 7 above)
Comment (cheesefrycook) [reply Managers/RandomGenerator.gd:7]: Okay, I can add this

**Comment by cheesefrycook** [comment]
I can write a few tests then

File: Cards/CardBase.gd (hunk 4 above)
Comment (Turtyo) [reply Cards/CardBase.gd]: I see that we don't `_on_card_play(caster, target)` inside `try_play_card`, but I don't see where we play the card after discard

Comment (Turtyo) [reply Cards/CardBase.gd]: ah nvm it's in Battler.gd, all good

**Comment by Turtyo** [comment]
> I can write a few tests then

yes, we won't run them with the CI since it's not working currently, but if you can confirm all the tests pass locally at least then it's all good

File: Managers/RandomGenerator.gd [hunk 15]
```diff
@@ -2,8 +2,17 @@ extends Node
 ## Initializes randomness for the game.
 
 
+const GAME_SEED = -1
```
Turtyo, line 5: if this is `const` it means the `GAME_SEED` will always be at -1 when entering `_ready`, which means it will always be randomly set in `set_game_seed`. So we can't actually choose the seed.

Tests/test_cards.gd
```diff
@@ -0,0 +1,99 @@
...
+
+func test_draw_cards():
+	_card_container.draw_cards(5)
+	
+	assert_eq(_card_container.cards_in_hand.size(), 5)
+	assert_eq(_card_container.draw_pile.size(), 45)
+	assert_eq(_card_container.discard_pile.size(), 0)
+
+
+func test_draw_two_cards():
```
Turtyo, line 39: would rename to `test_draw_card_2()` to not be confused with the action of drawing two cards

Tests/test_cards.gd
```diff
@@ -0,0 +1,99 @@
...
+func test_discard_cards():
+	_card_container.draw_cards(5)
+	_card_container.discard_random_card(1)
+	
+	assert_eq(_card_container.cards_in_hand.size(), 4)
+	assert_eq(_card_container.draw_pile.size(), 45)
+	assert_eq(_card_container.discard_pile.size(), 1)
+
+
+func test_discard_two_cards():
```
Turtyo, line 57: same as above, would rename to `test_discard_cards_2`

Tests/test_cards.gd
```diff
@@ -0,0 +1,99 @@
...
+
+func test_draw_to_max():
+	_card_container.draw_cards(50)
+	
+	assert_eq(_card_container.cards_in_hand.size(), 10)
+	assert_eq(_card_container.draw_pile.size(), 40)
+	assert_eq(_card_container.discard_pile.size(), 0)
+
+
+func test_discard_specific():
```
Turtyo, line 91: maybe draw 5 and discard the 2 card for example, it would be more relevant than discarding the only card in hand I think.
But we would have to test that we indeed discarded the correct card if we do that, so we would need an identifier. Can we add data to the cards in `before_each` to give them a number ?

Turtyo (commented):
Thank you for adding the tests, are they all passing ?

Managers/RandomGenerator.gd
cheesefrycook: Yeah, we don’t have a UI or anything for this, so if you want to set the seed you can just modify the const

Tests/test_cards.gd
cheesefrycook: We can set the name of the card and check that. I’ll add it

cheesefrycook:
Yes all tests are passing. I’ll make the few changes above and push a bit later

Managers/RandomGenerator.gd
Turtyo: maybe a little comment above to ask people to set the const back to -1 when they are done with testing.

Tests/test_cards.gd
Turtyo: thanks

Tests/test_cards.gd
```diff
@@ -15,12 +15,20 @@ func before_each():
 	
 	# fill deck with 50 default cards
 	_card_container.default_deck.resize(50)
-	_card_container.default_deck.fill(CardBase.new())
 	_card_container.max_hand_size = 10
 	_card_container.starting_hand_size = 5
 	
+	# assign cards names of Card1-Card50
+	for card_index: int in _card_container.default_deck.size():
```
Turtyo, line 22: `in range(0,_card_container.default_deck.size())` ? will go from 0 to 49 so the index is the same as the card name
I don't know if it works without a range anyway

Tests/test_cards.gd
```diff
@@ -89,11 +97,12 @@ func test_draw_to_max():
 
 
 func test_discard_specific():
-	_card_container.draw_cards(1)
-	var card: CardWorld = _card_container.cards_in_hand[0]
+	_card_container.draw_cards(5)
+	var card: CardWorld = _card_container.cards_in_hand[2]
+	assert_eq(card.card_data.card_title, "Card3")
```
Turtyo, line 102: should also assert that `assert_eq(_card_container.cards_in_hand[0].card_title = "Card1"` (or Card0 if we go with the comment I added above). And the same for every card left in the hand

Tests/test_cards.gd
cheesefrycook: I think 1-50 is a bit more clear personally

Tests/test_cards.gd
cheesefrycook: Yeah that can be added, wouldn’t hurt. I won’t be able to push for a little bit so feel free to make the change if you’d like.

Tests/test_cards.gd
Turtyo: Well that's a personal choice, but does it work with `_card_container.default_deck.size()` and not `range(_card_container.default_deck.size())` ?

cheesefrycook: You mean does the current implementation properly loop through every index 0-49 without using range()? Yes it does I believe

Turtyo: I'm curious, because 50 is an int, and `in` needs an iterator iirc

Tests/test_cards.gd
Turtyo: I'll need to check if I can push on your branch

Tests/test_cards.gd
cheesefrycook: Just double checked and yes, this syntax loops through all 0-49

Tests/test_cards.gd
cheesefrycook: I'm back now, just pushed this

Tests/test_cards.gd
Turtyo: oh well that's strange I thought you needed a `range()` for it to work correctly.

Tests/test_cards.gd
```diff
@@ -98,10 +98,16 @@ func test_draw_to_max():
...
-	_card_container.discard_card(card)
+	var discard_card: CardWorld = _card_container.cards_in_hand[2]
+	
+	assert_eq(_card_container.cards_in_hand[0].card_data.card_title, "Card1")
+	assert_eq(_card_container.cards_in_hand[1].card_data.card_title, "Card2")
+	assert_eq(discard_card.card_data.card_title, "Card3")
+	assert_eq(_card_container.cards_in_hand[3].card_data.card_title, "Card4")
+	assert_eq(_card_container.cards_in_hand[4].card_data.card_title, "Card5")
+	
+	_card_container.discard_card(discard_card)
```
Turtyo, line 110: this should be before the `assert_eq` and the lines after should be:
```
        assert_eq(_card_container.cards_in_hand[0].card_data.card_title, "Card1")
	assert_eq(_card_container.cards_in_hand[1].card_data.card_title, "Card2")
	assert_eq(_card_container.cards_in_hand[2].card_data.card_title, "Card4")
	assert_eq(_card_container.cards_in_hand[3].card_data.card_title, "Card5")
```
This test is to make sure cards are discarded properly, which is not what is done here if you test before discarding

cheesefrycook: Oh I see what you mean now. I'll fix this

cheesefrycook: Should be good now

Turtyo (approved):

Turtyo:
I'm guessing all tests are still passing ?

cheesefrycook:
Yes

Paper-2 (approved):

Issue #18, opened 2023-12-13
Add UI popup for draw/discard pile


cheesefrycook:
## Feature Description

Adds a UI popup that shows all the cards in the draw or discard pile on click. It will show all the cards in a grid.

## Motivation

Currently there is no way for a player to view what is in the discard pile or draw pile.

## Concept art / Designs

![image](https://github.com/Saplings-Projects/1M_sub/assets/151570126/098c46a3-8ff6-4ce1-991e-ec8f2c264715)

## Acceptance Criteria

- Clicking the draw pile shows a grid of all the cards in the draw pile
- Clicking the discard pile shows a grid of all the cards in the discard pile
- Clicking the back button when on the UI popup will take you back into the game
- Each UI popup is scrollable
- You cannot interact with the game while the UI popup is on screen

## Proposed Solution

Use a GridContainer Control Node to lay out the cards. Use a ScrollContainer Control Node to provide scrolling functionality. Populate the container by creating an instance of CardWorld and providing CardBase data to `init_card()`. Clicking either the draw pile or discard pile button will create a new CardScrollUI popup scene. Pressing the back button in this scene will destroy the scene.

## Implementation Steps (Optional)

-Add CardScrollUI scene
-Initialize scene when draw/discard button is pressed
-Destroy scene when back button is pressed
-Populate scene with either draw pile or discard pile based on what button was clicked. Use the `CardPileUISetter` for this, and obtain the draw pile or discard pile via `CardManager.card_container.draw_pile`

## Additional Context

Depends on #17

Paper-2:
The order should be an arbitrary one, like card rarity, energy cost, or alphabetical order. I don't think the player will have fun knowing what cards they will get next.

Paper-2:
Also this issue should deal with making an UI for the player inventory, which is all the cards that the player has obtained.

Turtyo:
Yes it can be the same as the inventory, since  the only difference is where you call it from

cheesefrycook:
That’s a good point. Maybe this issue should include adding a “deck view” button which shows your whole deck?

Paper-2:
Yes, That's I meant by an inventory. I should of said " ... player has obtained in the current run." instead of just "... player has obtained."

cheesefrycook:
If you'd like to do that in this issue, you can just use `default_deck` from CardContainer

Turtyo:
[...]

PR #22, opened 2023-12-17
Add UI popup for draw/discard pile

Paper-2:
# Description

Adds a UI popup that shows all the cards in the draw or discard pile on click. It will show all the cards in a grid.

## Related issue(s)

https://github.com/Saplings-Projects/1M_sub/issues/18#issue-2038974793

## List of changes

- Adds a UI that shows the content of the discard/Draw Pile. 

- Adds Deck view


Turtyo:
Yup just like that. You can use `-` before items to list them btw. It's the default format in the template that you should have seen when creating the PR.

Paper-2 (commented):
Me when git

Paper-2 (commented):
I'm missing something for the card to appear. It would be nice if someone told me what that thing is.

UI/card_pile_ui.gd
```diff
@@ -0,0 +1,54 @@
+extends Control
+
+@onready var gridcontainer: GridContainer = $GridContainer
+@onready var parent_node = $".."
+
+#I wanted to use pure_card as card_scene is used in (https://github.com/Paper-2/1M_sub/blob/bc1048d2eddec7aa6205d1bc2be40c519858b40a/Cards/CardContainer.gd#L117-L125)
+@onready var pure_card: PackedScene
```
Turtyo, line 7: what is this var ?

UI/card_pile_ui.gd
```diff
@@ -0,0 +1,54 @@
...
+
+var discard_pile: Array[CardBase] = CardManager.card_container.discard_pile
+var draw_pile: Array[CardBase]    = CardManager.card_container.draw_pile
+var deck: Array[CardBase]         = CardManager.card_container.default_deck
+var cards_to_display: Array[CardBase]
```
Turtyo, line 12: maybe initialize it at `[]` ?

UI/card_pile_ui.gd
```diff
@@ -0,0 +1,54 @@
...
+
+
+# TODO bind esc to return to the game
+func _on_button_pressed():
```
Turtyo, line 16: think about adding typing:
`func _on_button_pressed() -> void:`

UI/card_pile_ui.gd
```diff
@@ -0,0 +1,54 @@
...
+	for card in cards_to_display:
+		var cardworld: CardWorld = CardWorld.new()
+
+		#var cardworld: CardWorld = pure_card.instantiate() This line also gives an error :WideFauna2:
+		print(card, cardworld)
+		
+		#Why doesn't this line work? I Keep getting the error Cannot call method 'add_child' on a null value.
+		#gridcontainer.add_child(cardworld)
+
+		$GridContainer.add_child(cardworld)
```
Turtyo, line 44: do you have a grid container node in the scene to begin with ?

UI/card_pile_ui.gd
```diff
@@ -0,0 +1,54 @@
...
+
+		#the card gets initialize yet its node doesn't show up in the screen
```
Turtyo, line 46: probably due to previous errors

UI/card_pile_ui.gd
Paper-2: The equivalant to card_scene used in cardcontainer

UI/card_pile_ui.gd
Paper-2: Yes, there is a grid container in the scene, otherwise I wouldn't be able to use $Gridcontainer

UI/card_pile_ui.gd
Turtyo: maybe keep the same name so we don't get confused

Turtyo:
How is it working currently ?

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
+extends Control
+
+var card_scene: PackedScene = preload("res://Cards/Card.tscn")
+
+var discard_pile: Array[CardBase] = CardManager.card_container.discard_pile
+var draw_pile: Array[CardBase]    = CardManager.card_container.draw_pile
+var deck: Array[CardBase]         = CardManager.card_container.default_deck
+var cards_to_display: Array[CardBase] = []
+var card_worlds: Array[CardWorld] = []
+var cardworld: CardWorld
```
Turtyo, line 10: `= null` for default value

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
...
+
+func _on_button_pressed() -> void:
+	queue_free()
+
+#Pressing escape will return you to the game
+func _input(event):
```
Turtyo, line 16: `-> void`

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
...
+	if event.is_action("escape"):
+		queue_free()
+
+
+
+func populate(parent_name: String) -> void:
+	
+	if parent_name == "DiscardPile":
```
Turtyo, line 24: might be good to have a switch case instead of an if elif

Turtyo (commented):
Will checkout locally to test the PR before approving anything. I also need to check if the changes I pointed in previous comments have been taken into account (if they are still relevant ofc)

Turtyo:
Tested in local everything seems to be working fine. Will approve once the comments have been addressed.

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
...
+var deck: Array[CardBase]         = CardManager.card_container.default_deck
+var cards_to_display: Array[CardBase] = []
+var card_worlds: Array[CardWorld] = []
+var cardworld: CardWorld = null
+
+func _on_button_pressed() -> void:
+	queue_free()
+
+#Pressing escape will return you to the game
+func _input(event) -> void:
```
Tysterman74, line 16: Can we specify the type here, to help with readability?

UI/ShowCardPile.gd
```diff
@@ -0,0 +1,13 @@
+extends Button
+## Shows an UI for the player to see the content of that specific Card Pile
+
+@onready var parent_node = $".."
+@onready var root = parent_node.get_parent()
+@onready var cardUI: PackedScene = preload("res://#Scenes/CardScrollUI.tscn")
+
+	
+func _pressed() -> void:
+	#print(parent_node.get_name())
```
Tysterman74, line 10: Commented out print line here, is this still needed?

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
...
+
+		cardworld.init_card(card)
+	
+	
+
+
+func control_cardworld(card: CardWorld) -> Control:
+
+	var control: Control = Control.new()
+	var cardui: Control = card.get_children()[2] # Gets the control node CardUI
```
Tysterman74, line 51: Magic Number here, I would recommend to make a constant of the index 2 to specify exactly which node you're getting here. Something like:

```
const CONTROL_NODE_INDEX = 2
```

That way we can remove that comment you have right next to it.

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
...
+var cards_to_display: Array[CardBase] = []
+var card_worlds: Array[CardWorld] = []
+var cardworld: CardWorld = null
+
+func _on_button_pressed() -> void:
+	queue_free()
+
+#Pressing escape will return you to the game
+func _input(event) -> void:
+	if event.is_action("escape"):
```
Tysterman74, line 17: @Turtyo Should we add a constants file to the AutoLoader that specifies all of the different input keys we have? This would help with other screens that potentially may use "escape" input event identifier and can reference the constant itself.

Tysterman74 (changes requested):
Mostly just clean up comments for readability, and a question to Turtyo in regards to the Input Keys.

UI/CardScrollUI.gd
Turtyo: yes you totally can, good catch ! also maybe naming it `event` when we will have an `EventBase` class is maybe not a good idea i think

UI/CardScrollUI.gd
Turtyo: yes good idea ! It's preferable to have this kind of thing in the same place and everything using this referencing to the same place

UI/CardScrollUI.gd
Paper-2: Oh right, Im going to that and make all of the arbitrary variables constant too for future stuff too

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,71 @@
...
+		queue_free()
+
+
+
+func populate(parent_name: String) -> void:
```
Tysterman74, line 22: Let's have a unit test for this function. We're using this function for DiscardPile, DrawPile, and Deck, so having a unit test on the logic there would help. Checking that the text is correct when set, and that we're setting the proper Array as well.

Tysterman74 (changes requested):

UI/CardScrollUI.gd
cheesefrycook: Could alternatively use is_key_pressed with the KEY_ESCAPE constant. I think the only buttons we’ll ever have in this game is mouse buttons and escape, so we don’t need to do anything too complicated for input probably

Turtyo: Having another file where you go search your constant doesn't really complicate it (it's just referencing somewhere else basically). Not making a change because "we might not need it" is hoping that we don't make anything more complicated is generally not good (because we might indeed need it later, so it's better to have now than to have to change it later when there are more things to add). Especially when the change is simple.

We might want to add more controls later, we never know.

Paper-2:
Currently, there's a problem where only the top part of the last row gets shown, I'm going to wait until  PR #25 gets finished to continue working on this.

Tests/test_pile_ui.gd
```diff
@@ -6,6 +6,9 @@ var _card_container_scroll_scene: PackedScene  = load("res://#Scenes/CardScrollU
 var _card_container: CardContainer = null
 var _card_scroll: Control = null
 
+const SCROLLCONTAINER = 2
```
Turtyo, line 9: what are those exactly ?

Paper-2 (commented):
> what are those exactly ? 

the index to get the scrollcontainer, same with the other constant

Turtyo:
You can reply directly on the review instead of making a new comment btw. And i think I missed some commits which is why I don't get the code, will need to check that.

Turtyo:
Closing this because of the commits from the main being rebased in the PR instead of merged (and the state of the local was a bit too complex to reorganize), so we'll just open a new PR with a clean branch and all the relevant commits.

Turtyo:
New PR is #40

PR #25, opened 2023-12-18
Deck animation and layout

cheesefrycook:
# Description

Adds drawing and discarding animations for cards. Lays out cards in the hand in a fan-shaped layout.

## Related issue(s)

#12 

## List of changes

- Adds `CardMovementComponent` which controls movement state switching on cards.
- Adds `CardMovementState` Resource and several child resources to represent each movement state.
- Adds `CardStateProperties` to store data for each state.
- `CardContainer` creates a fan-shaped layout for cards.
- Drawing and discarding is handled differently since cards are drawn/discarded with an animation. Now when cards are drawn/discarded, they are added to a queue to be spawned/destroyed at a later time.
- Changes `Cards` and `CardContainer` Node to Control instead of Node2D so we can easily anchor the container to the bottom of the screen.
- Restricts End Turn button to only allow presses if cards are not being dealt.

## Tests

`test_cards.gd` was updated to work with these changes.
Removed `before_all` check in `test_health.gd` because it would never start now that it's no longer the first test.

## Additional notes

https://github.com/Saplings-Projects/1M_sub/assets/151570126/a47532fc-790a-4dd2-a4f4-bf45b11e135f


Turtyo:
Damn that's clean (at least for the video, so it's good proof that it works). Well check the code later :+1:

Cards/CardMovementComponent.gd
```diff
@@ -28,14 +41,26 @@ func set_movement_state(new_state: Enums.CardMovementState):
 
 
 func _on_state_enter(state: Enums.CardMovementState):
-	pass
+	if _has_state(state):
```
Turtyo, line 62: I'm not too up to date on what this does, quick explanation ? Since I see it's a repeating pattern between some functions

Cards/Movement State/MoveState_Discarding.gd
```diff
@@ -0,0 +1,18 @@
+extends CardMovementState
+class_name MoveState_Discarding
+
+
+# @Override
+func on_state_enter() -> void:
+	# Move to discard pile
+	_state.card.create_tween()\
```
Turtyo, line 16: was this function `create_tween` function declared previously ? if not what does it do ?
based on the name I would guess it's something for curves maybe ?

Cards/CardContainer.gd
```diff
@@ -321,4 +325,28 @@ func _update_card_positions() -> void:
...
+		var viewport_width: float = get_viewport_rect().size.x
+		var card_index_scaled: float = Helpers.convert_from_range(card.global_position.x, 0.0, viewport_width, -1.0, 1.0)
+		
+		# curve the cards in hand by finding the y value on a parabola
+		# y = ax^2 - a where a = max offset and x = card_index scaled to a range of -1 to 1
+		card_y = (pow(card_index_scaled, 2.0) * max_hand_offset_y) - max_hand_offset_y
+		
+		# rotate cards on a cubic curve
+		# y = ax^3 where a = max_rotation and x = value between [-1, 1]
+		var rotation_amount: float = pow(card_index_scaled, 3) * max_rotation
```
Turtyo: in the same way that the cards width update based on number of card, does the rotation update based on number of cards ? it probably does based on the card index being cast to -1,1 ?

Cards/CardContainer.gd
```diff
@@ -321,4 +325,28 @@ func _update_card_positions() -> void:
 				else:
 					card_x += scaled_hover_offset
 		
-		movement_component.state_properties.desired_position.x = card_x
+		# scale card's x value to a range of [-1, 1] from range of [0, viewport width]
+		var viewport_width: float = get_viewport_rect().size.x
+		var card_index_scaled: float = Helpers.convert_from_range(card.global_position.x, 0.0, viewport_width, -1.0, 1.0)
```
Turtyo: this says that the card_index is being scaled to -1,1 but i feel like the viewport width is being scaled to -1,1 (which doesn't feel like a card index, ie the number of the card in the hand)

Cards/Movement State/MoveState_InHand.gd
```diff
@@ -2,6 +2,10 @@ extends CardMovementState
 class_name MoveState_InHand
 
 
+@export var lerp_speed: float = 10.0
```
Turtyo: should this be renamed to `movement_speed` since functions containing the word `lerp` have been renamed too ?

Cards/CardContainer.gd
```diff
@@ -335,24 +335,5 @@ func _update_card_positions() -> void:
...
-				movement_component.state_properties.desired_rotation = rotation_amount
-			Enums.CardMovementState.QUEUED:
-				movement_component.state_properties.desired_position.x = card_x
-				movement_component.state_properties.desired_position.y = card_y
-				movement_component.state_properties.desired_rotation = rotation_amount
-			Enums.CardMovementState.IN_HAND:
-				movement_component.state_properties.desired_position.x = card_x
-				movement_component.state_properties.desired_position.y = card_y
-				movement_component.state_properties.desired_rotation = rotation_amount
+		movement_component.state_properties.desired_position = Vector2(card_x, card_y)
```
Turtyo, line 381: Much simpler, this is nice :+1:

UI/EndTurnButton.gd
```diff
@@ -2,13 +2,11 @@ extends Button
...
 
-func _ready() -> void:
-	PhaseManager.on_phase_changed.connect(_on_phase_changed)
-
-
-func _on_phase_changed(new_phase: Enums.Phase, _old_phase: Enums.Phase) -> void:
+func _process(delta: float) -> void:
 	# disable button during enemy attack phase
-	disabled = new_phase == Enums.Phase.ENEMY_ATTACKING
+	disabled = Enums.Phase.ENEMY_ATTACKING == PhaseManager.current_phase\
```
Turtyo, line 7: should maybe add a little delay also for the end button, to make sure all effects have been applied. I don't remember, do we apply effects then discard or discard then apply effects ? If it's the second, we could image lots of effects playing and the player pressing the end turn button.
Maybe we can add a signal for when all effects have been applied ?

Turtyo (commented):
This review was long so I hovered on the stuff that seemed important at first glance. I will try to read a bit more in depth on some parts tomorrow but being tired doesn't help (which is why i may be asking some questions on stuff that changes later on and I didn't realize it).

The video helps to see that it actually works, so at least we don't have doubts on this ! A very nice job. Did you watch any kind of video or any resources to make this . If yes, I think linking it could be nice to have an overview of what you did.

Cards/CardMovementComponent.gd
```diff
@@ -0,0 +1,41 @@
...
+
+var current_move_state: Enums.CardMovementState = Enums.CardMovementState.NONE
+var desired_position: Vector2 = Vector2.ZERO
+
+
+func _process(delta):
+	_on_state_process(current_move_state)
+
+
+func set_movement_state(new_state: Enums.CardMovementState):
```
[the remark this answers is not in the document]
cheesefrycook: This is an outdated version, should be good in latest

Cards/CardContainer.gd
```diff
@@ -267,7 +274,14 @@ func _update_card_positions() -> void:
...
 		var card: CardWorld = cards_in_hand[card_index]
 		var card_x: float = per_card_width * card_index
+		
+		var movement_component: CardMovementComponent = card.get_card_movement_component()
+		var move_state: Enums.CardMovementState = movement_component.current_move_state
 
 		card_x -= total_hand_width / 2.0
 
-		card.get_card_movement_component().desired_position.x = card_x
+		match move_state:
```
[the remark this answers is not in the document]
cheesefrycook: I believe this is an outdated version, we don't do this match anymore

Cards/Movement State/MoveState_Discarding.gd
cheesefrycook: `create_tween` is a built in godot function. It pretty much creates a request to change a value over time.  [https://docs.godotengine.org/en/stable/classes/class_tween.html](url)

Cards/CardContainer.gd
cheesefrycook: Yeah, so it's using the card's x-position as an input for the y-position and rotation functions. The `card_index_scaled` is an outdated name, I believe I updated this naming to something else in the latest version

Cards/CardContainer.gd
cheesefrycook: Yeah, this is an old naming. I believe I updated this naming in the latest version

Cards/CardMovementComponent.gd
cheesefrycook: You mean `_has_state`? This is just checking if the _state_mapping has a non-null value.

Cards/CardMovementComponent.gd
```diff
@@ -4,10 +4,23 @@ class_name CardMovementComponent
 
 signal on_movement_state_update(new_state: Enums.CardMovementState)
 
-@export var lerp_speed: float = 4.0
```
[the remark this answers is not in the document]
cheesefrycook: Movement speed is handled in each `CardMovementState`, we aren't doing any movement in the `CardMovementComponent` anymore

Cards/Movement State/MoveState_InHand.gd
cheesefrycook: Which functions are you referring to that were renaming? I'm okay with changing the naming if it's more clear

UI/EndTurnButton.gd
cheesefrycook: We discard and then apply effects. Right now the only delay is the 1 second delay in Battler whenever the enemies attack, which should give the effects enough time to cast. In the future we probably want to play each effect on the card one by one and play some sort of animation. I want to write a issue on this soon

Cards/CardMovementComponent.gd
Turtyo: no I meant what was coming after, sorry

UI/EndTurnButton.gd
Turtyo: Good idea, we would know when each effect animation is done via a signal probably, which would solve the problem I mentioned at the same time.

Cards/Movement State/MoveState_InHand.gd
Turtyo: for example `get_lerp_component` was renamed to `get_card_movement_component`

Cards/CardMovementComponent.gd
cheesefrycook: Yeah this is when the state initializes. This part gets called right when we enter a new state from set_movement_state. And init_state is just passing the state resource object stored here to the newly set state. Then when you change the state_properties in here, it will also update in the state itself

Cards/Movement State/MoveState_InHand.gd
cheesefrycook: It’s still technically doing a lerp in the CardMovementCimponent, but I can change it because it’s more clear

cheesefrycook:
As a heads up, I most likely wont be able to push any changes for the next two weeks because I’ll be away for the holidays. So if it’s possible for someone else to push updates to my fork that’d be good so we don’t have to wait

Turtyo:
We should be able to push to your fork since `Maintainers are allowed to edit this pull request. ` is on.
As this is visual and we have a lot of back-end left to do, I might to push to this though (since there is a lot of code, and I would need to really read everything, not just a quick review).
If someone else wants to do it, then good. If not we'll wait the two weeks. Thanks for the heads up anyway !

Tysterman74:
If it's alright with @Turtyo I can take over the changes.

UI/EndTurnButton.gd
Tysterman74: Hmm spitballing here, would having a signal on the CardContainer like `on_card_effect_finished` be useful for checking here?

cheesefrycook: It’s not going to do much right now I think since we already have the one second delay when enemies attack. But we can probably add something similar to that in a future PR along with the card animations I mentioned

Turtyo:
Sure, go ahead

Cards/Movement State/MoveState_InHand.gd
Turtyo: honestly, I never saw the word lerp before starting this project

UI/EndTurnButton.gd
Turtyo: Just keeping this open so i remember to put an issue or a little note for myself somewhere

Turtyo (commented):
Thank you for the changes ! I'm just waiting for the points made in unresolved discussions to be addressed before approving

Tysterman74:
In reference to this thread about adding a delay for effects: https://github.com/Saplings-Projects/1M_sub/pull/25#discussion_r1430778946

I made a change on this PR to add a new signal:
```
signal on_card_action_finished(card: CardWorld)
```

There is a new array in CardContainer.gd that will hold an array of these active cards called `_active_cards`. Within Battler.gd, when we remove the card from being queued in `_try_player_play_card_on_entity`, we then add it to the active cards array, then within CardBase.gd at the end of `on_card_play` we will fire the signal to then remove it from the `_active_cards`. This feels pretty lightweight since it's dependent on whatever shenanigans we do in `CardBase`. Let me know what y'all think!

Turtyo:
We can indeed play a signal once the card effects have been played in the `on_card_play`
Though things might change due to https://github.com/Saplings-Projects/1M_sub/issues/31

cheesefrycook:
I like the active card idea! But yes I believe we will only have one active card at a time, so may make more sense to not do an array

Tysterman74:
Understood, I'll make the change to just a CardWorld object. Thanks for looking!

Turtyo:
I'll just pull the branch later to check if everything is working in local (since this is visual). If it does, I'll approve, since I don't really see anything else to change for the code

Turtyo:
I'm noticing that the enemy turn is being played before the animation of the cards being discarded is finished, which feels a bit strange. If you can think of an easy way to fix that, then go ahead, if not we can think about it later.

cheesefrycook:
That’s a good point. We could also do other stuff when that signal fires, like “Start Enemy Turn” text pop up

Turtyo:
Indeed we could do it that way. i'm not the most knowledgeable about this part of the code, think you can make something for it to work that way @Tysterman74 ?

Cards/Card.tscn
Paper-2, line 1: if possible can you add an area/size to the root node? it would help a lot.

Turtyo: Area/size of what ? the hand ? the cards ?

Paper-2: To the control node of the card scene.

cheesefrycook: I believe this PR makes the root node of the Card scene a TextureButton with a width and height

Paper-2: It is not, the only TextureButton that I see is the node named SelectionButton under the CardUI node.

cheesefrycook: I misremembered then. I think we can give the root node a width and height that matches the width and height of the button. Or we could change the root node to be the button

Paper-2: I think the former should be easier to change.

Tysterman74: Hey! I ended up moving the Transform values from the CardUI Control to the Root Card Control. Within the game, it looks to be the same, but the .tscn file just looks weird now with how it's shifted. 🤔 let me know if this is good for ya Paper!
image

Paper-2: It looks weird because the position is not zero.

Tysterman74: Without any changes it looks like CardUi is already set to Anchor to Center so we should be good here 👍 if it's good to you Paper you can go ahead and Resolve this thread. Thanks for looking!

Tysterman74:
Okay waiting for the card animation is throwing some major hands. I couldn't figure out a solution, but I do have some setup code I want to run it by y'all before I push the change and we can wrap back to it later.

I created a new Phase called PLAYER_FINISHING to basically handle anything we may need to resolve for the player's turn, which in this case is the animation to finish.
image

There is a new signal on CardContainer.gd called `on_finished_discarding_hand` that is emitted at the end of `discard_all_cards()`.
image

The signal is connected to in Battler.gd which will then trigger the phase to ENEMY_ATTACKING. When we're in PLAYER_FINISHING phase we check if we need to connect to the signal as well:
image

The issue with this however is that the Card animation itself is independent of what happens in CardContainer, so if we do any check if anything is queued for discard it's going to be zero since it looks like the animations are firing off asynchronously. TLDR Cheese you did a fine job implementing the animation 😂 

I'll leave a TODO on where we should have logic to handle refactoring to pick up reading when all cards are done animating, but I got a feeling this may be a bigger refactor of CardContainer.

Thoughts?

cheesefrycook:
I wonder if we can fire the signal in ‘_handle_ discard_queue’ similar to how the ‘on_cards_finished_dealing‘ signal fires off in ‘_handle_card_draw_queue’. That will make sure the signal fires when the discarding animations are all done

Turtyo:
I had an idea that might solve the problems. Instead of waiting for a signal for when the animation has ended we do the following:
- We know when the turn ends (when the end of turn button was pushed by the player)
- This is also the same time at which the hand is discarded
- We know the time it takes to discard a given card (but this is a constant we set)
- We emit signal on turn end to start a counting, we go to next phase at the end of counting
- Value of the counter is `time_to_discard_a_card * numbers_of_card_in_hand`

Tysterman74:
Cheese's comment worked like a charm. I spent hours on that one yesterday haha

https://github.com/Saplings-Projects/1M_sub/assets/5923102/406b7cff-d58a-4cc0-a950-87c94bbd6d9f

Tysterman74:
Actually, I think we would need to have a new Phase in-between PLAYER_ATTACKING and ENEMY_ATTACKING. `_handle_discard_queue()` fires off whenever we PLAY a card, and it will eventually fire off the signal after playing a single card. We also can't set Phase to ENEMY_ATTACKING as well, cause then the Battler.gd will then start the Enemy turn, async from the CardContainer logic.

cheesefrycook:
That’s a good point, didn’t think about that. Do you have an ideas for how to fix? I wonder if we should address this in a separate issue so we can go ahead and unblock #18

Tysterman74:
I'd think just having another phase to be able to handle all of this is good. That being said, I think we're getting a bit lost in the sauce discussing this out haha. I think we can make this into a bug, this feels like a good issue to get someone familiar with the codebase. I'll leave it to fearless leader @Turtyo though.

Turtyo:
Yeah I think getting into its own issue is good, as you said good for someone who wants to start on the project (though the animation part might not be the easiest). I might be a bit biased because I know it less, will need to go back to all this code to approve anyway. Unless someone does it before me.
I'll pull the branch to test locally, but it looks nice in video :+1: thank you very much @cheesefrycook and @Tysterman74

Turtyo:
When testing in local, I get a few warnings:
```
W 0:00:00:0527   The parameter "delta" is never used in the function "on_state_process()". If this is intended, prefix it with an underscore: "_delta".
W 0:00:00:0527   The parameter "new_state" is never used in the function "can_transition_from()". If this is intended, prefix it with an underscore: "_new_state".
W 0:00:00:0534   The parameter "card" is never used in the function "remove_active_card()". If this is intended, prefix it with an underscore: "_card".
W 0:00:00:0534   The local variable "move_state" is declared but never used in the block. If this is intended, prefix it with an underscore: "_move_state".
  <GDScript Error>UNUSED_VARIABLE
  <GDScript Source>CardContainer.gd:362
```
And the enemy turn still happens before all cards are discarded. **Are you sure you pushed your last changes** @Tysterman74 ? Because I don't see new pushes between the moment you said you were searching for a solution and now.

Tysterman74 (dismissed):
From discussions above, I have created https://github.com/Saplings-Projects/1M_sub/issues/34 to address the bug of waiting for discard animation to finish before starting enemy turn. Everything else looks good and will approve 😊

Turtyo:
Thank you for the issue, indeed better to treat it in another discussion. I didn't realize this was why you didn't push the changes...
I'm confused as to how you approved the PR even though you are the last pusher @Tysterman74 . I'll hold on from approving for now, I would rather have someone else than you review your own changes.

Turtyo (approved):

Paper-2 (approved):

Issue #26, opened 2023-12-18
Reformat `test_health.gd` so it uses a card as if it was being played


Turtyo:
## Is your amelioration related to a problem? Please describe.

Currently, `test_health.gd` is testing by directly changing the values of health (with `DealDamageData.new()` and `_player_health_component.deal_damage(deal_damage_data)`). This was good for a first test, but this is not how a player will interact with the game. When I was writing card effects, the tests were passing but the cards didn't deal any damage at all (because the implementation was incomplete at the time I tested it by launching the test scene).

## Describe the solution you'd like
Reformat the test so that it uses cards with effects (after #14 is closed), instead of directly dealing the damage via function call.

## Describe alternatives you've considered
Make another test that tests the cards in themselves, that way we can test separately the functions that deal damage from the thing that calls them (eg the cards)

## Additional context
Needs #14 to be closed by PR

cheesefrycook:
Sounds good to me!

Turtyo:
Would we change `test_health` or do we create a new test file ?

cheesefrycook:
Yeah we may as well create a new test file. Like test_effects.gd or something, and we can test a bunch of different cards in there

Turtyo:
Or maybe we rename test_card to test_draw instead, and the new test_card file actually tests the cards

Tysterman74:
"amelioration" is such a fancy word I didn't even know about it until reading this. I'd love to take this one once #14 is done, gives me a chance to dive into the unit tests a bit more!

Turtyo:
It didn't seem that fancy to me, but probably because it's often used in french... I could have said "betterment", "improvement" or such... Or enhancement
I can change the template if this word is unclear / it might give trouble to some people

Issue #21, opened 2023-12-15
 Instant-cast cards


cheesefrycook:
## Feature Description

Cards should have the ability to instantly play their effects, instead of always having to select a target. To better facilitate this, we can change card input from click-only to drag-and-drop.

## Motivation

Currently, all cards require you to select a target. If it's a healing card, you have to click the player. If it's an attack card, you have to click an enemy. However, this doesn't make much sense for cards where there isn't a specific target. Some examples:
- a "draw cards" card - currently you have to select the player to play this card
- cards that damage the entire enemy party - currently you have to select a specific enemy

Because of this, it makes sense to allow those cards to be cast instantly without needing to select a specific target.

A drag-and-drop system makes more sense if we implement instant-cast cards. If we implement instant cast cards with the current input system, then the player could easily click a card in their hand on accident and play it by mistake. It's a lot more difficult to make a mistake with a drag-and-drop system. A drag-and-drop system also feels more intuitive, in my opinion. 

## Acceptance Criteria

- Certain cards can be instantly played
- Cards can be dropped into the play area to play them
- Right click cancels dragging and returns the card to the player's hand

## Proposed Solution

To play a card, you first have to grab it and drop it into the "play area", which is the portion of the screen outside of your hand. Upon doing so, one of two things happens:
- The card instantly applies its effects to the relevant targets.
- The card moves to a central location and the player must now select a target. Upon selecting a target, the card's effects are applied.

This will require modification of the CardContainer. When a card is clicked, instead of becoming the `queued_card` it becomes the `dragging_card`. If the `dragging_card` is a certain distance away from its starting position on the y-axis, it can be **played** by clicking again. A new state will be added to `CardMovementComponent` (see #12) called `FOLLOWING_MOUSE`.

CardEffects will contain a new `CastType` enum:
- `CastType`
  - `SINGLE_TARGET` - affects a single target. If a card with this effect is played by a player, they have to select an enemy to cast the card on.
  - `ALL_TARGETS` - affects all targets. If the card only contains this CastType, then the card is instantly cast. The player will not have to select a target.

## Implementation Steps (Optional)

- Implement new `CastType` enum
- Allow cards to be cast instantly instead of having to select a target
- Implement drag-and-drop style input

## Additional Context

Depends on #14 and #12

Turtyo:
Will also need to change the queue/unqueue cards to take this new step in account. Apart from that, everything is clear for me !

cheesefrycook:
[...]

PR #27, opened 2023-12-18
Class structure for the cards

### LINKED ISSUE #14: Class structure for the cards

Turtyo:
# Description

This PR rewrites the system that was in `CardBase` to have a more flexible system of effects that can be added to a card.

## Related issue(s)

Closes #14 

## List of changes

- Creates `EffectBase` and `EffectData`: `EffectBase` is parent of all effects. EffectData is the actual thing that is useful for the card, as it contains the effect, the caster, the target and the value useful to the effect (damage to deal, card to draw, number of poison status to apply...)
- Moves functions that were in `CardBase` to their respective effect
- Rewrites the way damage is dealt to multiple targets by providing a list of targets (eventually reduced to only one element) instead of relying on boolean checks inside functions. This list of targets should be created at card creation (ie at the start of combat). It should be updated once an enemy dies during the combat. Creation will be done by the card parser (TODO), and the update is to discuss in further issues.

## Tests

No new tests were provided, but #26 should be addressed once this PR closes.

## Additional notes

In Draft status so the code can be reviewed. I need to test in the actual player that everything is working correctly, but I need to edit the cards that are given to the player in that scene to be sure (which I don't have time to do for a bit of time).

I am also not sure of what to do with effects that deal damage indirectly like poison. Currently those status don't seem to directly deal damage, as the only thing we do is add or remove them from the list of current status. Changes will probably need to be made to those status to inflict damage (either by calling the `EffectDamage` or directly calling the `deal_damage` function, which seems more logical as effects are the player interface and it's a bit too much to call them for internal logic).


Cards/Effects/EffectBase.gd
```diff
@@ -0,0 +1,10 @@
+# Base class for card effects
+class_name EffectBase extends Node
```
cheesefrycook, line 2: Should probably extend from Resource

Cards/Effects/EffectData.gd
```diff
@@ -0,0 +1,22 @@
+class_name EffectData extends Node
```
cheesefrycook, line 1: Should extend from Resource?

Cards/Effects/EffectDamage.gd
```diff
@@ -0,0 +1,12 @@
...
+# @Override
+func apply_effect(caster: Entity, target: Entity, value: int) -> void:
+    var target_damage_data: DealDamageData = DealDamageData.new()
+    target_damage_data.damage = value
+    target_damage_data.caster = caster
+    target.get_health_component().deal_damage(target_damage_data)
+
+# TODO : what to do for Status that inflige damage over time ?
+    # Should they call an EffectDamage when they trigger each turn ?
+    # or should they directly call deal_damage on the target ?
```
cheesefrycook, line 12: Could do either, but makes more sense to just call deal_damage I think

Cards/CardBase.gd
```diff
@@ -18,78 +18,37 @@ class_name CardBase
...
-func _deal_damage(caster: Entity, target: Entity) -> void:
-	# damage target
-	if damage_to_apply_to_target != 0.0:
-		_damage_entity(caster, target, damage_to_apply_to_target, affect_all_targets)
-	
-	#damage caster
-	if damage_to_apply_to_caster != 0.0:
-		_damage_entity(caster, caster, damage_to_apply_to_caster, affect_all_casters)
+# Need a parser of some sort to read a plainform card data and translate it to EffectData
+# This will call EffectData.add_effect_data() in a loop for every new effect
```
cheesefrycook, line 37: Maybe instead of passing all the data to the EffectData from here, we instead determine what it should be from properties set in the editor. So EffectData would look more like this:
```GDScript
#EffectData.gd
@export var effect: EffectBase = null
@export var application_type: ApplicationType = Enums.ApplicationType.ENEMY_ONLY
@export var value: int = 0
```
So with this, we just set the EffectData array directly from the godot inspector on each CardBase because of the `@export` tag. Then to determine the proper `list_enemy` and `caster` we use `application_type`.

Cards/Effects/EffectDamage.gd
Turtyo: I agree

Cards/Effects/EffectBase.gd
Turtyo: I didn't really check the difference between Node and Resources. Node is for instances ? I will check that tomorrow and change if needed

cheesefrycook: Nodes are meant to be added to the scene tree, so they have transform data. We don't really need that for this. Also, Resources are automatically garbage collected

Cards/CardBase.gd
Turtyo: When we set from inspector, is that not manual ? how does it work when we are doing this automatically ?
I also don't really like to have an `application_type` honestly.

cheesefrycook: How would setting up EffectData work automatically? I imagine there's always a bit of manual setup for each of the EffectData - otherwise how would you set the `value` or associated `EffectBase`?

For the `application_type`, it's going to be useful when we implement instant-cast cards #21. For an instant-cast card, the `application_type` will be used to determine which party to cast on. Otherwise I'm not sure how we would determine who to cast on since the player isn't selecting a target.

Cards/Effects/EffectBase.gd
Turtyo: Will change in next push

Cards/Effects/EffectDamage.gd
Turtyo: Will probably open a new issue for the Status modifications I think.

Cards/CardBase.gd
Turtyo: The goal of the parser is to translate a card into our infrastructure automatically, so we don't need to manually create each card in the editor. It would work by having a certain structure when writing cards (like the JSON file I mentioned).

I also thought about the `application_type`. The instant-cast card can know who to cast to because of the list of enemies. But it might not know that the list of enemies is the entire enemy team, so that's why the `ENEMIES_ALL` could be useful. But for the logic of applying effects I think it's better to have a list like that.

We could do the following:
- Each card has either a type ALL or SELECT  (the select could be a mono-target, or for example 1 target and the two adjacent to the target, but you would still need to select a main target anyway...). Actually it'll probably be 4, SELECT_ENEMY, SELECT_ALLY, ALL_ENEMY, ALL_ALLY
- Depending on the type: if it's an ALL card, we populate the list with all the party member (depending if it's enemy or ally). If it's a select, the list will get the selected elements (that the player is clicking on / dragging on / whatever the pointing system is).
But ultimately we would keep the list of targets in this system, we just have an enum to know how to fill the list.

Turtyo:
Will make the cards that are played using the new structure and test that it works then it'll be good for merge

Turtyo:
Alright here is the problem:
If I want to be able to modify the list of target entity from the editor, I need to export them. But since entities are Node, EffectData should extend from a Node and not a Resource.
So either I don't export and we don't modify from the inspector, or i export but I need to make EffectData a Node a and not a Resource.

Cards/CardBase.gd
cheesefrycook: Is the parser the thing you mentioned in the issue that auto generates the card from a text description? 

That system for targeting may make more sense. If we do it like I was thinking, we would have an application_type and a cast_type that together form the 4 combinations you were thinking of. Consolidating to one is probably better if we don’t need them separate for any reason. Would that still work for cards that target both the enemy and player? (Ex: heal 1 hp to caster, deal 5 to all enemies)

cheesefrycook:
We probably don’t want to assign the targets in the inspector, because that list can change at runtime. Can we somehow pass the list of enemies to the effect data at runtime when they’re needed? Maybe from Battler when the card is played?

Turtyo:
Or I just export everything except the entity. That works too.

Turtyo:
Note that damage all doesn't work anymore due to the targeting system. We will need to treat that at the same time as instant  cast cards (by implementing the "all / target" system that was discussed in https://github.com/Saplings-Projects/1M_sub/issues/21 )

Turtyo:
Also I noticed a bug on the draw, which can't be fixed in the scope of the current PR (I think) so i'll open an issue for that

Cards/CardBase.gd
Turtyo: After having worked with the inspector to make some cards, it works nicely. The problem of that is that you need to have Godot to make cards. On the other hand, it allows us to not have to write a parser (which could be complicated due to the way Godot writes in resource files, which I'm not exactly sure to understand all the time).

And to answer your question @cheesefrycook the way we can do for cards with multiple effects is to just consider the effect with the most restrictive cast (ie if you have an ALL and a TARGET effect, then the card becomes a TARGET card, you can't auto-cast because it has an effect with a target). And we can easily have multiple effects on the same card now.

Tysterman74 (changes requested):
Overall changes look good. I have a question though: what's the difference between a Status and an Effect?

Turtyo:
The difference between a status and an effect is that the effect is the action (like dealing damage, applying poison)
The status is a condition on the character that provides either buff or debuff (weakness, strength, poison).
If we talk a bit more about poison, the poison effect is the thing that applies the poison, the status poison is the condition that gets inflicted to the character and that ends up dealing damage each turn.

Cards/Effects/EffectDamage.gd
```diff
@@ -0,0 +1,12 @@
...
+# @Override
+func apply_effect(caster: Entity, target: Entity, value: int) -> void:
+	var target_damage_data: DealDamageData = DealDamageData.new()
+	target_damage_data.damage = value
+	target_damage_data.caster = caster
+	target.get_health_component().deal_damage(target_damage_data)
+
+# TODO : what to do for Status that inflige damage over time ?
+	# Should they call an EffectDamage when they trigger each turn ?
+	# or should they directly call deal_damage on the target ?
```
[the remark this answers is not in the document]
Turtyo: The thing about EffectDamage it adds another layer of interaction. It's useful for making cards, but when the call is to deal damage
The status already have a countdown in themselves to know when it triggers. And generally speaking, a status does one thing (it doesn't trigger multiple things, so it doesn't really make sense for it to have effects, because it's not a card)

Tests/test_cards.gd
```diff
@@ -6,7 +6,7 @@ var _battler_scene: PackedScene = load("res://Core/Battler.tscn")
 var _card_container_scene: PackedScene = load("res://Cards/CardContainer.tscn")
 var _battler: Battler = null
 var _card_container: CardContainer = null
-var _cards_in_hand: Array[CardWorld] = []
+# var _cards_in_hand: Array[CardWorld] = []
```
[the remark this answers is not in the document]
Turtyo: Will check, probably not (I commented during the pushes to check if it broke nothing, but sometimes I forget to remove the lines I commented)

Cards/Effects/EffectDamageHealth.gd
```diff
@@ -0,0 +1,11 @@
+class_name EffectDamageHealth extends EffectDamage
+
+# @Override
+@warning_ignore("unused_parameter")
+func apply_effect(caster: Entity, target: Entity, value: int) -> void:
+	var _amount_of_health_lost: int = 0
+	var _caster_health_component: HealthComponent = PlayerManager.player.get_health_component()
```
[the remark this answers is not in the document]
Turtyo: As the commit message said, I don't think such a function would be used as is anyway. It's a modification of a card cheesy made last time (so i tried to reproduce it given the description, which says "deals damage equal to health lost")

of course this is OP and wouldn't go in the game like that.

Cards/CardBase.gd
```diff
@@ -14,82 +14,34 @@ class_name CardBase
...
-func _draw_cards() -> void:
-	CardManager.card_container.draw_cards(amount_of_cards_to_draw)
+func can_play_card(caster: Entity, target: Entity) -> bool:
+	return caster.get_party_component().can_play_on_entity(application_type, target)
 
 
-func _discard_random_cards() -> void:
-	CardManager.card_container.discard_random_card(amount_of_cards_to_discard)
+func on_card_play(entity: Entity) -> void:
+	_apply_all_effects(entity)
```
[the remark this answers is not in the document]
Turtyo: this will most likely always be caster (unless we imagine that mobs can make you discard cards :thinking:)

Cards/CardBase.gd
```diff
@@ -14,82 +14,34 @@ class_name CardBase
...
-		var party: Array[Entity] = target.get_party_component().party
-		assert(party.size() > 0, "Entity has an empty party. Make sure you added party members.")
-		
-		for party_member: Entity in party:
-			party_member.get_health_component().deal_damage(target_damage_data)
-	else:
-		target.get_health_component().deal_damage(target_damage_data)
+func _apply_all_effects(entity: Entity) -> void:
+	for effect_data: EffectData in card_effects_data:
+		effect_data.apply_effect_data(entity)
```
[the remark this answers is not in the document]
Turtyo: i'm not sure. Currently it wouldn't really make sense to put "target" there, because you can target the caster (for heal and draw for example), **BUT** later on the effects that apply to the caster would be quick-cast (meaning you wouldn't need to actually target the caster).

So TL;DR : we can put `target` even if it doesn't make sense entirely for now, because it will make sense with quick-cast

Tests/test_cards.gd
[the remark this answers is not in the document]
Tysterman74: Feel free to resolve this once you've removed it!

Cards/Effects/EffectDamageHealth.gd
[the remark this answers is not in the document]
Tysterman74: Ok! Resolving this comment.

Cards/CardBase.gd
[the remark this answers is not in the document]
Tysterman74: Okay, makes sense. I'll resolve this comment.

Cards/Effects/EffectDamage.gd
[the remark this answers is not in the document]
Tysterman74: Understood! Think it was just my misunderstanding of not knowing Effect vs. Status. If you want to resolve this thread feel free to do so, unless you wanted other people to chime in.

Cards/CardBase.gd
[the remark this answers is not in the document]
Tysterman74: Okay! I only ask for readability. 👍

Tests/test_cards.gd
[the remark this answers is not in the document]
Turtyo: I remembered it was because the variable was unused that I commented it. Will resolve once I made the push

Cards/Effects/EffectDamage.gd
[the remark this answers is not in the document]
Turtyo: Yeah there is some confusion right now because not everyone knows the different classes (which is why I opened #32 because it will be needed)

Cards/CardBase.gd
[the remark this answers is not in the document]
Turtyo: ah nvm I was reading the line that got deleted. It would be the target I think. will push later

Turtyo:
I will let you resolve the conversations if you think the changes are correct @Tysterman74

Cards/Effects/EffectDamage.gd
Turtyo: I realized Status already inflict damage by calling deal_damage, so no need for an issue

Cards/Effects/EffectDamage.gd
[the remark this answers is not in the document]
Turtyo: I realized Status already inflict damage by calling deal_damage, so no need to change anything actually

Tysterman74 (approved):

Cards/CardBase.gd
Tomzkk: Basically repeating @cheesefrycook's question - would we be able to affect both player and an enemy with different effects with current setup ? - since on_card_play applies all effects to just the card target ?
EffectData does have both _caster and _target variables, but per my understanding caster is never actually set and target is the same for all effects on the card ?

Turtyo: `on_card_play` applies effect to the target. But effects that apply to the player don't need a target, since we only have one player. If you have a heal effect, it will heal the player.
Caster is not currently set because we are waiting for tag system for targeting. And yes target is the same for all effects on the card, because you may only target one entity when a card (and effects that apply to all targets don't need a target).

Tomzkk: So any effect that would target caster would have to be made individually as one without a target - so if we already had heal with a target for example we would have to make heal without one that just heals the caster ? Probably not a problem, and don't think there will be too many like that I imagine, just wondering 🤔

Turtyo: No, the effect can have the caster be both caster and target. If you have an effect that heals self and a target, then the system will use the effect with a targeting to self and the other selected target.

Tomzkk: I mean if we already had a regular targetted heal that takes target and heals it, that the monsters would use for example and then we wanted to make it target the caster, wouldn't we need to create a new one for it ? Or am I missing something ?

Turtyo: The caster is an entity. So you can target the target (ie target yourself).
In practice, we will implement tags that tells you if the target is self or not, so you can know in advance if you need to check for a target. That way we can use the same effect for both targeting someone else and self, but just by changing the tag associated with it (and we also have a tag for teams, so with a different tag you can heal an entire team).

Turtyo: In a simpler fashion, currently you can't target both yourself and a target at the same time. That will come later, and that was not the goal here. Currently you can target any entity on the board, but one at a time.

Tomzkk (approved):

Issue #20, opened 2023-12-15
Event structure


Turtyo:
## Feature Description

An `Event` is a description of the content in each `Room`. It tells us what the encounter is (mystery encounter, mob, boss, shop etc...). Depending on the type of the encounter, it gives more information, for example for mobs it gives the mobs that are in the room. For a random encounter, it gives the chosen encounter.

## Motivation

This would be useful to have a place to hold the data of the different encounters. We could put it directly in the room in itself, but i think it's better to have separated, since it will probably be easier to test and debug that way.

## Acceptance Criteria

- The `Event` can be modified at runtime (due to dependence to the light level)
- We can easily differentiate between different possible events
- Hierarchy between events depending on light level is clear
- A room event can't change anymore once the player went above the level of the room containing the event

## Proposed Solution

The `Event` class in itself is the parent of possible events. So we would have `Event_mob`, `Event_shop`, `Event_heal` etc... as extending from the `Event` class.
The event is only chosen when it enters player view (because they used a torch or just that they are next to it). The `Event` holds an array of the level depending on light value.
For example if at light 0 we have a mob, at light 1 we have a random event and at light 2 we have a shop, it looks like this `[Event_mob, Event_random, Event_shop]`.

Note that this doesn't mean we will necessarily have a mob at light level 0, it just explains which event you go to when light increases/decreases. We could have a random event at light 0, thus when we get to light level 1 we get a shop (because it's the next one in the list).

The choice of the event will probably be handled at map level, because it will have a global vision and can distribute events better.

## Additional Context

In link with #6 and #10

cheesefrycook:
So each Event would contain an array of Events? I think it makes more sense to store that array in the Room, since the Room knows about the light level. 

Other than that this looks good to me. I think Event can extend from Resource, then each child Event could do whatever logic it wants. I imagine each Event type will load a unique type of scene with some data passed in from the Event (like Scene_Encounter, Scene_Shop, Scene_Heal, etc)

Turtyo:
No, only the base `Event` class would hold an array. This array is meant as a way to make a link before the different possible rooms depending on the light level. If the light level changes, you ask the `Event` class for the corresponding sub event_class (ie `Event_mob`, `Event_shop` etc...). Each event would only be one event in itself, but if there are changes we have a table that tells us which event we should get next.

Each event will indeed have a unique type of scene (or multiple type of scene if we want different backgrounds depending on where we are in the tree, having different background at the bottom and at the top).

[...]

PR #29, opened 2023-12-19
Issue 20: Events Structure

Tysterman74:
# Description

Creation of the Event Parent and Event Children.

## Related issue(s)

Resolves https://github.com/Saplings-Projects/1M_sub/issues/20

## List of changes

- Created Event.gd which acts as the parent and holds some virtual functions that the children can override. It also holds a scene string of which scene should load to trigger this event
- Created Event_Heal.gd, Event_Mob.gd, Event_Random.gd, Event_Shop.gd which are child classes of Event and can have specific behavior defined for them
- Created EventsClassification.gd which holds an enum of all of the current types of Events we have. This has been added to the AutoLoader in order to make the enum accessible anywhere if needed.

## Tests

TBD

## Additional notes

If you have anything more to add.


Event/Event.gd
```diff
@@ -0,0 +1,11 @@
+extends Resource
+class_name Event
```
Turtyo, line 2: would rename to `EventBase` (because even if we didn't agree explicitly on that, it seems to be the convention for parent classes until now)

Event/Event.gd
```diff
@@ -0,0 +1,11 @@
...
+
+var EVENTS_CLASSIFICATION = [Event_Mob, Event_Random, Event_Shop, Event_Heal]
```
Turtyo: maybe a `const` would make more sense ? since it probably won't be modified at runtime.
Also i think it's lacking typing:
`var EVENTS_CLASSIFICATION: Array[Event] = [Event_Mob, Event_Random, Event_Shop, Event_Heal]`

Event/Event_Heal.gd
```diff
@@ -0,0 +1,8 @@
+extends Event
+class_name Event_Heal
```
Turtyo, line 2: rename to EventHeal. I need to modify the contributing to include the class name conventions...

Event/EventsClassification.gd
```diff
@@ -0,0 +1,16 @@
+extends Node
+
+enum EventsClassification { EVENT_MOB, EVENT_RANDOM, EVENT_SHOP, EVENT_HEAL }
```
Turtyo, line 3: the problem with an enum is that it doesn't hold classes, under the hood it's int.
The advantage of having a list of Event is to be able to directly instantiate the correct scene by checking the Array, without having to make a correspondence between `EVENT_MOB` (in the enum) and `Event_Mob` (the class)
So the function to create an event would become:
```GDScript
func createEvent(index: int) -> Event:
    return EventsClassification[index].new("res://Event/EventTest.tscn")
```

Turtyo (commented):
The different events are placeholders for now (which is good, since we don't really have anything to put there for now). Later, we will probably add respective sprites into those scenes.

Event/Event.gd
Tysterman74: I'll need some help here, I get this compiler error yelling at me when I try to change it to a const

```
Line 4:Cannot have an element of type "Event_Mob" in an array of type "Array[EventBase]".
Line 4:Assigned value for constant "EVENTS_CLASSIFICATION" isn't a constant expression.
```

Event/EventsClassification.gd
Tysterman74: Gotcha! I'll remove this and just stick with the Array then. That createEvent function was mainly used for testing anyways.

Event/Event.gd
Turtyo: ah probably because you can't have a const with an event that has values that can change

Event/EventsClassification.gd
```diff
@@ -1,17 +0,0 @@
-extends Node
```
Turtyo: You could keep the `EventsClassification` if you want but with the simplified structure. Or you can put everything in `EventBase`

Event/EventBase.gd
```diff
@@ -1,13 +1,13 @@
 extends Resource
 class_name EventBase
 
-const EVENTS_CLASSIFICATION: Array[EventBase] = [EventMob.new(""), EventRandom.new(""), EventShop.new(""), EventHeal.new("")]
+var EVENTS_CLASSIFICATION: Array[EventBase] = [EventMob.new(), EventRandom.new(), EventShop.new(), EventHeal.new()]
```
Turtyo, line 4: i'm not sure it's needed to have a .new() here i think, it's not really useful (will will create new or copy later on if we need it anyway

Event/Event.gd
Tysterman74: Ok, so Godot's yelling at me at this new change:

```
var EVENTS_CLASSIFICATION: Array[EventBase] = [EventMob, EventRandom, EventShop, EventHeal]

Line 4:Cannot have an element of type "EventMob" in an array of type "Array[EventBase]".
```

Which is kind of weird, since all of those classes extend from EventBase. However, 
```
var EVENTS_CLASSIFICATION: Array[Resource] = [EventMob, EventRandom, EventShop, EventHeal]
```

Doesn't throw a compiler error. I'll leave it as that for now, unless you have a better recommendation.

Event/EventBase.gd
Tysterman74: Updated!

Event/Event.gd
Turtyo: Yeah it doesn't know they are child of this class if not instantiated, I had the some problem the other day

Event/EventBase.gd
```diff
@@ -1,7 +1,7 @@
 extends Resource
 class_name EventBase
 
-var EVENTS_CLASSIFICATION: Array[EventBase] = [EventMob.new(), EventRandom.new(), EventShop.new(), EventHeal.new()]
+var EVENTS_CLASSIFICATION: Array[Resource] = [EventMob, EventRandom, EventShop, EventHeal]
```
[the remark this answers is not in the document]
Turtyo: Nvm I saw you already talked about it above

Turtyo:
Some things in the issue have not been done, but after reading back I realize it's probably more fit to be treated at map level (like locking events once a certain level of the map is passed through)

Turtyo (approved):

Event/EventBase.gd
```diff
@@ -0,0 +1,18 @@
+extends Resource
+class_name EventBase
+
+var EVENTS_CLASSIFICATION: Array[Resource] = [EventMob, EventRandom, EventShop, EventHeal]
```
Tomzkk, line 4: What is the purpose of events_classification ? Will every event have these 4 (and some will be empty) or are these the types that are possible ? It feels a little bit out of place to me but it might be because I'm not understanding the purpose properly

Tomzkk (approved):
Don't have much more to say as we didn't really set specifics for events yet, as a base looks okay

Event/EventBase.gd
Turtyo: You should read https://github.com/Saplings-Projects/1M_sub/issues/20#issuecomment-1858996318 and https://github.com/Saplings-Projects/1M_sub/issues/20#issuecomment-1861759738

Turtyo: If everything is clear for you after reading those comments, merge the pull request (since you are the last one to approve)
if not, ask for me explanations on points you don't get

Tomzkk: Looking at the 2nd comment I'd say I'm still not 100% clear on it, reading the 1st one made me think the same thing as what Ty commented there - that it's pretty much corresponding to the light level and progression of it. 

That being said I imagine this will probably change altogether once we actually implement the light system in it's entirety ?

Turtyo: Will make a drawing once i'm free today

Turtyo: Tell me if you understand with this @Tomzkk 

light_mechanic_explanation

Tysterman74: https://media.tenor.com/qhcQRwZKtJMAAAAM/it-makes-so-much-sense-jimmy-o-yang.gif

Tomzkk: Looks okay, I think I get it. Sounds like I had the right idea just it's not happening for all the rooms

Turtyo:
This PR was meant to be merged on map-implementation, so it has been cherry-picked onto the `map-implementation` branch.

Issue #10, opened 2023-12-07
Room class structure


Turtyo:
## Feature Description

What should be the structure of the `Room` class ? 

## Motivation

Rooms being the most basic element of the map should have a carefully thought structure, since the entire map is just assembling rooms.

## Acceptance Criteria

Tests should pass.

## Proposed Solution

Rooms have the following elements:
- an integer to know the light level (0 being unlit, 1 is dimly lit, 2 is lit, 3 is brightly lit), could also be an `ENUM`
- a boolean to know if a torch is in the room (but maybe we want to let the player put multiple torches in a room ?)
- an event ( #20 )

## Additional Context

This issue is in link with the map implementation #6

cheesefrycook:
Makes sense to me. I think the Room can extend from Resource similar to CardBase. Then the Map can generate a list of Room when the run starts. Then when we want to display the map to the player, we initialize a bunch of RoomWorld with the data from the Room array. RoomWorld can be a Node2D/Control node that just holds Room data and displays it. Having the separate Room and RoomWorld structure like this will let us load/unload the map display, while keeping the actual data stored in the array in Map.

Turtyo:
Let's call it `RoomDisplay` then.

Turtyo:
I think the display will be made in another PR and not in the initial room implementation

Issue #6, opened 2023-12-06
Map implementation


Turtyo:
## Feature Description

A map composed of different floors, each floor composed of different rooms. After each combat you open the map and choose your next destination on the floor above you.
Each room has an icon depending on the encounter. The player can move to the room directly above them or the ones on the side of the room directly above them.
The entire map shouldn't be visible, and should interact with torch placement, cf #9

## Motivation

Since we are making a dungeon crawler, we have decided on using a map with floors and nodes (which we will probably need to name "rooms" or something else to not get confused with Godot nodes). This issue is meant to discuss the implementation of such a map system.

## Concept art / Designs

![floor_concept](https://github.com/Saplings-Projects/1M_sub/assets/49365680/28c91478-a888-4f67-a6e7-2f54025cc472)

## Acceptance Criteria

- Map should only be used once after each room.
- Movements on the map must make the player go up one floor (no movement on the same floor or to a floor below).
- You can toggle viewing the map (if you have something else you to check first this can be useful)
- Map should have icons on each room depending on the room type
- Each room can have its event changed depending on light level (cf #10 )
- The map should have an indicator of the rooms the player can access from their current position
- The entire game can be composed of several maps (sections), where you access next section after beating the current section's boss
- Map should be interacted with via the mouse (open to discussion)

## Proposed Solution

The most basic element should be the `Room`. Then we have `Floor` that are composed of rooms and finally the `Map` is composed of floors.
When the player starts a new map, an algorithm makes a distribution of the rooms, starting with the most influential events such as mini-boss and shops / heal points, then fills the remaining rooms with mobs / mystery events (or even other types of events depending on if we add more).
The indicator of the rooms that are accessible could be a having the border of the rooms glow.

A `Room` should have variables to do:
- associated event (and thus we need to create an "event" class)
- light level and a function to calculate the light level
- a boolean to know if we show the tile or not. This could also be directly linked to the light level to have non-visible, dimly lit, lit, brightly lit which appear differently

Cf #10 for the discussion on `Room`.

Floors are a collection of rooms. Not sure of the date structure to use here. We need the floor to ensure that movements are made only from one floor to the one above

Map can be an array of floors.

Other possibility instead of having floors and map class is to have the `Map` be a `2D Array` of rooms, which might facilitate the manipulation of said object.

## Implementation Steps

We should first focus on implementing the structure of the map, then:
- how to move between the rooms and indicator
- create the `Event` class [ #20  ]
[...]

PR #30, opened 2023-12-21
Randomized map generator

multi-arm:
# Description

Added base classes for map, floor and room
Added manager script to manage said base classes

## Related issues

Issue 10 (#10), Issue 6 (#6)

## List of changes

Created MapBase, FloorBase and RoomBase classes
Created MapManager script
Created test_map

## Tests

Added one test for generation. More tests will be added in the future

## Additional notes

Trying this again, everything should be fixed now hopefully but I'll make this a draft too just in case.


Map/FloorBase.gd
```diff
@@ -0,0 +1,6 @@
+extends Resource
+class_name Floor
```
Turtyo, line 2: The `Floor` class was decided against as per this comment https://github.com/Saplings-Projects/1M_sub/issues/6#issuecomment-1852966312
The current structure for the map should be a 2D array instead of an Array of floors

Map/MapBase.gd
```diff
@@ -0,0 +1,7 @@
+extends Resource
+class_name Map
+## Class for a map. Provides classes for the map
+##
+## Holds basic functionality for a map
+@export var floors: Array[Floor]
```
Turtyo, line 6: This should be a 2D array of rooms since the floor implementation was decided against.

Map/RoomBase.gd
```diff
@@ -0,0 +1,11 @@
+extends Resource
+class_name Room
+## Class for a room on the map. Provides base classes for all rooms to use regardless of the type
+##
+## Holds basic functionality for a room on the map
+@export var light_level: int = 0
+@export var has_torch: bool = false
+
+func get_room_type():
```
Turtyo, line 19: by room type do you mean the event in the room ?

Map/MapBase.gd
```diff
@@ -0,0 +1,7 @@
+extends Resource
+class_name Map
```
Turtyo, line 2: should be named `MapBase`

Map/RoomBase.gd
```diff
@@ -0,0 +1,11 @@
+extends Resource
+class_name Room
```
Turtyo, line 2: should be named `RoomBase`

Map/MapManager.gd
```diff
@@ -0,0 +1,26 @@
+extends Node2D
+## Class to manage backend for rooms (generation and such)
+
+signal populate_floor ## call this to generate rooms in a floor
```
Turtyo, line 4: is this signal really needed ? the signal to say we finished maybe, but a signal to start ? can't we just call the function ?

Edit : it won't even be needed since we won't use floors

Turtyo (changes requested):
The implementation using `Floor` was decided against in #6
The implementation should use a 2D array of rooms instead. Changes are needed to remove the use of floors in this implementation

Map/RoomBase.gd
```diff
@@ -1,21 +1,17 @@
...
 
 @export var light_level: int = 0
 @export var has_torch: bool = false
-@export var this_room_type:RoomType = RoomType.UNKNOWN
-@export var coordinates:Vector2
-func get_room_type():
-	return this_room_type
-
+@export var this_room_event: RoomEvent
+@export var coordinates: Vector2
```
Turtyo, line 17: i'm not exactly sure a room needs to know where it is on the map. That's more of an information for the map level I think

Map/MapManager.gd
```diff
@@ -1,26 +1,21 @@
...
-			var _generated_room:Room = Room.new()
-			var _possible_room_types = _generated_room.RoomType
-			var rand_type = _possible_room_types.keys()[randi() % _possible_room_types.size()]
-			_grid[x][y] = rand_type
+	for x in range(_map.width): ## loop through elements of the grid and assign a room type
+		for y in range(_map.height):
+			var _generated_room: RoomBase = RoomBase.new()
+			var _rand_type: String = _generated_room.RoomEvent.keys()[randi() % _generated_room.RoomEvent.size()]
+			var _rand_type_index: int = _generated_room.RoomEvent.get(_rand_type)
+			_grid[x][y] = _rand_type
```
Turtyo, line 16: _rand_type of type `String` gets assigned to _grid[x][y]
then we do `_map.rooms = _grid`
so the rooms are string, which is not what we want, we want a map of rooms

Map/MapBase.gd
```diff
@@ -1,7 +1,10 @@
 extends Resource
-class_name Map
-## Class for a map. Provides classes for the map
+class_name MapBase
+## Class for the map. Provides functions for the map
 ##
 ## Holds basic functionality for a map
-@export var floors: Array[Floor]
+
+@export var rooms: Array = [] ## 2D array of rooms on the map. this will be assigned when generating rooms
```
Turtyo, line 7: should be `@export var rooms: Array[RoomBase] = []`

Map/MapBase.gd
```diff
@@ -1,7 +1,10 @@
...
+@export var width:int = 3
```
Turtyo, line 8: this supposes that the width of the map is constant. This is a supposition we can make on height, but as seen in the drawing of the issue #6 the width can change.
The width should be an array instead of a singular value. There is also a condition on the array which is that each floor can only be 2 more wide than the floor before it (to be sure we can always access all the rooms on the next floor).
This wasn't explicitly discussed in the issue, i will add a comment now to talk about this with more details.

Turtyo (commented):
Thank you for the contribution ! I left comments on points that I felt needed changes. We can discuss if you think some of the changes I suggest shouldn't be made

Turtyo:
i'm also realizing the room doesn't have an EventBase element, but you will need to wait for #29 to be closed to be able to reference that properly (which will also give you access to the list of possible events)

Turtyo:
Now that #29 has been closed, you can properly reference the Event class in the rooms

Turtyo:
Did you commit everything again ? Because I don't see the previous commits, they seem to have disappeared.

Tests/test_map.gd
```diff
@@ -0,0 +1,7 @@
+extends GutTest ## Tests for MapManager, more will be added in the future
+
+var test_generator = load("res://Map/MapManager.gd")
+
+func test_map_gen():
+	var result = test_generator.create_map()
+	assert_eq(result, result, "Expected to return generated class")
```
Tysterman74, line 7: A better test here would be `assert_not_null(result, "Expected to return generated class")`. We would want to be checking for equality if we make a test object within this file to compare it against!

Map/RoomBase.gd
```diff
@@ -0,0 +1,10 @@
+extends Resource
+class_name RoomBase
+## Class for a room on the map. Provides base classes for all rooms to use regardless of the type
+##
+## Holds basic functionality for a room on the map
+
+@export var possible_events: Array = [EventHeal, EventMob, EventRandom, EventShop]
+@export var light_level: int = 0
+@export var has_torch: bool = false
+@export var this_room_event: EventBase
```
Tysterman74, line 10: If we're going to be making several instances of RoomBase, I don't think it's necessary to prepend `this` for `this_room_event`. 🤔 It shouldn't have knowledge of other rooms in this class so we should be able to just name it `event` or `room_event`!

Tysterman74 (changes requested):
Overall, solid implementation on creation of the 2D for the map! Just added a few comments for cleanup and readability.

Map/RoomBase.gd
```diff
@@ -0,0 +1,10 @@
+extends Resource
+class_name RoomBase
+## Class for a room on the map. Provides base classes for all rooms to use regardless of the type
+##
+## Holds basic functionality for a room on the map
+
+@export var possible_events: Array = [EventHeal, EventMob, EventRandom, EventShop]
```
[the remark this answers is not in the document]
Turtyo: `EventBase` has one but you can't reference it directly. A way to do that is to create a new `EventBase` and then do a `event.EVENT_CLASSIFICATION`

A better way is to use a global file to store our `EVENT_CLASSIFICATION`, which extends from a Node and to auto-load it at project start with (Project -> Project settings -> Auto-load). If we call this script `GLOBAL_VAR.gd` you can reference his elements with `GLOBAL_VAR.ELEMENT`.
I think we should implement this second option because it might be useful in the future. I should have thought about loading in other files when I first asked for this array...

Map/RoomBase.gd
Turtyo: `room_event` seems like a good name

Map/MapBase.gd
```diff
@@ -0,0 +1,9 @@
+extends Resource
+class_name MapBase
+## Class for the map. Provides functions for the map
+##
+## Holds basic functionality for a map
+
+@export var rooms: Array = [RoomBase] ## 2D array of rooms on the map. this will be assigned when generating rooms
+@export var width: PackedInt32Array = [2,3,2]
```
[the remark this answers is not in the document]
Turtyo: This is kinda "magic number" but it actually makes sense. If you look at the issue, you can see that the map has a variable width. The numbers here represent the width of successive floors. I don't think it would be good to add naming because this might be extended to more than those 3 numbers (there will be a many width as there are floors).

A comment would be good to explain that though.

Map/MapManager.gd
```diff
@@ -0,0 +1,21 @@
...
+		for j: int in _map.width[i]:
+			_grid[i].append(0)
+
+	for x: int in range(_map.width.size()): ## loop through elements of the grid and assign a room type
+		for y: int in range(_map.width[x]):
+			var _generated_room: RoomBase = RoomBase.new()
+			var _rand_type_index: int = randi() % _generated_room.possible_events.size()
+			var _room_event: EventBase = _generated_room.possible_events[_rand_type_index].new()
+			_generated_room.this_room_event = _room_event
+			_grid[x][y] = _generated_room
```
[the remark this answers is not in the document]
Turtyo: It probably works to do both at the same time. Would also rename `x` to `width_index` and `y` to `height_index` maybe ? or `floor_index` but we should probably avoid using the name `floor` for now to not get confused with the initially discussed implementation

Map/RoomBase.gd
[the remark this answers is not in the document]
multi-arm: > **A better way** is to use a global file to store our `EVENT_CLASSIFICATION`, which extends from a Node and to auto-load it at project start with (Project -> Project settings -> Auto-load). If we call this script `GLOBAL_VAR.gd` you can reference his elements with `GLOBAL_VAR.ELEMENT`. I think we should implement this second option because it might be useful in the future. I should have thought about loading in other files when I first asked for this array...
for this, did you want it to be added in this pr, or a separate one?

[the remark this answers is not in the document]
Turtyo: You can do that in the current PR since it's quick to do and you need the list of events (which already exists).

Map/MapBase.gd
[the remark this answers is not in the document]
Tysterman74: A comment for clarity on this would be fine then.

Map/MapBase.gd
```diff
@@ -5,5 +5,4 @@ class_name MapBase
 ## Holds basic functionality for a map
 
 @export var rooms: Array = [RoomBase] ## 2D array of rooms on the map. this will be assigned when generating rooms
-@export var width: PackedInt32Array = [2,3,2]
-
+@export var floors_width: PackedInt32Array = [2,3,2] ## array for variable map width, each element is the width of a floor
```
Turtyo, line 8: good comment
on the topic of map width, you should check my latest comment #6 so you can see the latest ideas I had for map requirement width (i think it should make it so that the map will properly work, I'm not sure for the end though)

Map/MapManager.gd
```diff
@@ -4,18 +4,14 @@ extends Node2D
 static func create_map(): ## Generates and Populates a map with rooms that have random room types. More in depth algorithms will be added in the future
 	var _map: MapBase = MapBase.new()
 	var _grid: Array = [] ## 2d array to return. this will be populated with rooms
-	
-	for i: int in _map.width.size(): ## construct 2d array to return
+	for index_width: int in range(_map.floors_width.size()): ## loop through elements of the grid and assign a room type
```
Turtyo, line 7: this should be the opposite actually, `index_height`. Since the array gives the different width, the array is being read iterating through the heights.

Map/MapManager.gd
```diff
@@ -4,18 +4,14 @@ extends Node2D
...
 		_grid.append([])
-		for j: int in _map.width[i]:
-			_grid[i].append(0)
-
-	for x: int in range(_map.width.size()): ## loop through elements of the grid and assign a room type
-		for y: int in range(_map.width[x]):
+		for index_height: int in range(_map.floors_width[index_width]):
```
Turtyo, line 9: On this it should be: `for index_width: int in range(_map.floors_width[index_height]):`
this is logical since you are on the array that gives different width, so an element of it is a width and not a height

Map/MapManager.gd
```diff
@@ -4,18 +4,14 @@ extends Node2D
...
+			_grid[index_width].append(0)
```
Turtyo, line 10: is this really necessary since you could append the generated rooms later instead of replacing the 0 ? maybe it's actually better for readability, but it's a change of type. Maybe `null` instead ?

Turtyo (commented):
Nice on the global var, is it working properly with it ? if yes, I need the same global var file for my own PR, so I'll copy yours to not have merge conflicts

Map/MapBase.gd
multi-arm: this can be freely changed, and in the future it will be added with more in depth algorithms. i've changed it to reflect it though.

Turtyo:
I was thinking a little something. In the loop we append the rooms. This means that every room in the map is "justified to the left" in the matrix kinda (since we append at the start); it looks like that:
X
X X X
X X X X X
X X X X X X X
X X X X X
X X X
X

I think for traveling purpose logic later (to know which room you can be in), it would be better to be closer to the visual representation of the room, meaning like that:
         X
      X X X
   X X X X X
X X X X X X X
   X X X X X
      X X X
         X

A way to do that in the loop would be to check the max of the width (let's say it's 7), for each line you start by appending `null` exactly `padding = (max - current_width)/2` (so 3 for the first stage). You then append your rooms (one for the first floor). Then you append `padding` `null` times again.

For the second floor, it works again, `max` is 7, `width` is 3, you get `(7-3)/2 = 2` so you pad with 2 null, 3 rooms then 2 null. 

If we do it that way, the matrix will be square, symmetrical left right and symmetrical up down

Tests/test_eventbase.gd
```diff
@@ -7,15 +7,15 @@ func before_each():
...
-	var event1 = eventBase.EVENTS_CLASSIFICATION[1]
+	var event1 = GlobalVar.EVENTS_CLASSIFICATION[1]
 	assert_eq(event1, EventRandom)
 	
-	var event2 = eventBase.EVENTS_CLASSIFICATION[2]
+	var event2 = GlobalVar.EVENTS_CLASSIFICATION[2]
 	assert_eq(event2, EventShop)
 	
-	var event3 = eventBase.EVENTS_CLASSIFICATION[3]
+	var event3 = GlobalVar.EVENTS_CLASSIFICATION[3]
```
Tysterman74, line 19: Thank you for updating the tests 🙏

Map/MapManager.gd
```diff
@@ -6,14 +6,19 @@ static func create_map(): ## Generates and Populates a map with rooms that have
 	var _grid: Array = [] ## 2d array to return. this will be populated with rooms
 	
 	for index_height: int in range(_map.floors_width.size()): ## loop through elements of the grid and assign a room type
+		var _max_floor_size = _map.floors_width[_map.floors_width.size()/2]
```
Turtyo: this can be outside the loop, no need to calculate the max width of the map each time
the max isn't necessarily at the middle of the map either (you could imagine a map that gets big, smaller at the middle, then big again).
For this, i would advise on using the `max` function (which works on array). So probably:
`var _max_floor_size = max(_map.floors_width)`

Map/MapManager.gd
```diff
@@ -6,14 +6,19 @@ static func create_map(): ## Generates and Populates a map with rooms that have
...
 		_grid.append([])
+		var _padding = (_max_floor_size - _map.floors_width[index_height])/2
+		var _floor_width = _map.floors_width[index_height]
```
Turtyo: if you put this before the previous line, you can more easily write previous line as:
`var _padding = (_max_floor_size - _floor_width)/2`

Map/MapManager.gd
```diff
@@ -6,14 +6,19 @@ static func create_map(): ## Generates and Populates a map with rooms that have
 	var _grid: Array = [] ## 2d array to return. this will be populated with rooms
 	
 	for index_height: int in range(_map.floors_width.size()): ## loop through elements of the grid and assign a room type
+		var _max_floor_size = _map.floors_width[_map.floors_width.size()/2]
 		_grid.append([])
+		var _padding = (_max_floor_size - _map.floors_width[index_height])/2
```
Turtyo: would rename to `_padding_size`

Map/MapManager.gd
```diff
@@ -6,14 +6,19 @@ static func create_map(): ## Generates and Populates a map with rooms that have
...
+		var _floor_width = _map.floors_width[index_height]
+		for index_padding:int in _padding*2:
```
Turtyo, line 13: why use `_padding*2` ?
I think a much simpler way to add the padding is to Array constructors directly:
```GDScript
var _padding = Array(_padding_size)
_padding.fill(null)
_grid.append(_padding)
```
Then you add the rooms, and you add the padding again (no need to calculate it a second time, you already have it)

Turtyo (changes requested):
Thank you for implementing the idea I gave earlier @multi-arm !
I think it needs some changes. 

This would probably be a good thing to test in the map tests also (generate a small map and compare it to a reference we have, to be sure the padding an the rooms all are correctly set)

Turtyo:
I'm also seeing you added some typing for functions in the latest commit, could you take a look at these comments i left at the beginning about some missing types ?
https://github.com/Saplings-Projects/1M_sub/pull/30#discussion_r1434508980
https://github.com/Saplings-Projects/1M_sub/pull/30#discussion_r1434515781

Tysterman74:
Sanity checking this, I think we need to sync up `map-implementation` with main. I see the Contributing.md file has been changed and I believe multi synced up changes with main and is still merging into `map-implementation` with the new changes.

Turtyo:
If it's just for the Contributing.md, I don't think it's really needed to sync main in map (it's not a code dependency for example)

Map/MapManager.gd
```diff
@@ -1,25 +1,27 @@
...
-		var _padding: int = (_max_floor_size - _map.floors_width[index_height])/2
-		var _floor_width: int = _map.floors_width[index_height]
-		for index_padding:int in _padding*2:
-			if _padding > 0:
-				_grid[index_height].insert(index_padding, null)
+		
+		var _floor_width : int = _map.floors_width[index_height]
+		var _padding_size : int = (_max_floor_size - _floor_width)/2
+		
+		_grid[index_height].resize(_max_floor_size-_floor_width)
```
Turtyo, line 15: for clarity's sake, I would just put line 9 `_grid.append([])` before this one, so we can understand it's directly linked

Map/MapManager.gd
```diff
@@ -1,25 +1,27 @@
...
+		_grid[index_height].fill(null)
```
Turtyo, line 16: i see the idea (basically the same as appending at the start before putting rooms). But i don't see that being done after the rooms are put in the map.
Coming back to my previous comment on this, i think:
```GDScript
var _padding = Array(_padding_size)
_padding.fill(null)
_grid.append(_padding)
#
# file with rooms
#
_grid.append(_padding)
```
would be a clear approach

Map/MapManager.gd
```diff
@@ -1,20 +1,21 @@
 extends Node2D
 ## Class to manage backend for rooms (generation and such)
  
-static func create_map() -> MapBase: ## Generates and Populates a map with rooms that have random room types. More in depth algorithms will be added in the future
+#layout changes the width of the map's floors
+static func create_map(_layout: Array[int]) -> MapBase: ## Generates and Populates a map with rooms that have random room types. More in depth algorithms will be added in the future
```
Turtyo, line 5: I think `floors_width` was clear, why change to `_layout` which can describe more things ? to not be confused with the width of the map in itself ?
if that's the case, i would like to have a name that is close

Tests/test_map.gd
```diff
@@ -2,6 +2,20 @@ extends GutTest ## Tests for MapManager, more will be added in the future
 
 var test_generator = load("res://Map/MapManager.gd")
 
+
+
 func test_map_gen():
-	var result = test_generator.create_map()
-	assert_not_null(result, "Expected to return generated class")
+	var test_width: Array[int] = [1,3,5,3,1]
```
Turtyo, line 6: good way of testing it. Just to be exhaustive I think it should be good to test that the place where we expect a room is actually not null
Also from a code standpoint, maybe doing the following would be clearer:
```GDScript
var test_width: Array[int] = [1,3,5,3,1]
var test_map: MapBase = test_generator.create_map(test_width)
var expected_null_array: Array[Array] = [[0,0],[0,1],[0,3],[0,4],[1,0] ...]
for couple: Array[int] in expected_null_array:
    assert_null(test_map.rooms[couple[0]][couple[1]])
# you can another loop for the not null
```
need to check syntax since I just wrote it like that, but this allows to not add a new assert line every time you add more things (and the code is also shorter). One down side of that might if we have an error, it'll always be the same line. To prevent not knowing which position gives the error, a print that gives the current room would be good (so that if it fails, we know where we were)

Tests/test_map.gd
```diff
@@ -2,20 +2,12 @@ extends GutTest ## Tests for MapManager, more will be added in the future
...
-	assert_null(test_map.rooms[0][4])
-	assert_null(test_map.rooms[4][0])
-	assert_null(test_map.rooms[4][1])
-	assert_null(test_map.rooms[4][3])
-	assert_null(test_map.rooms[4][4])
-	assert_null(test_map.rooms[1][0])
-	assert_null(test_map.rooms[1][4])
-	assert_null(test_map.rooms[3][0])
-	assert_null(test_map.rooms[3][4])
+	for index_height in test_map.rooms.size():
```
Turtyo, line 8: I would rather have an explicit list than testing like that. This rises the chances of a bug slipping through because this code is incorrect, which will have less chances of happening with an explicit list. If you want something shorter, cf my previous comment on this test function.

Turtyo (changes requested):
Thank you for addressing the changes ! I think we will soon be done with this PR, that's a good implementation for a basis.
Some parts still need changes imo

Map/MapManager.gd
```diff
@@ -0,0 +1,28 @@
...
+		_grid.append([])
+		
+		var _floor_width : int = _layout[index_height]
+		var _padding_size : int = (_max_floor_size - _floor_width)/2
+		
+		_grid[index_height].resize(_max_floor_size-_floor_width)
+		_grid[index_height].fill(null)
+		
+		for index_width: int in range(_layout[index_height]):## loop through elements of the grid and assign a room type
+			var _rand_type_index: int = randi() % GlobalVar.EVENTS_CLASSIFICATION.size()
```
Tysterman74, line 19: Would `randi_range(0, GlobalVar.EVENTS_CLASSIFICATION.size() - 1)` be better here? `randi()` can return `2^32 - 1` so not sure if we want to be doing that big of a calculation for every room we go through.
`

Map/MapManager.gd
```diff
@@ -0,0 +1,28 @@
...
+			var _room_event: EventBase = GlobalVar.EVENTS_CLASSIFICATION[_rand_type_index].new()
+			var _generated_room: RoomBase = RoomBase.new()
+			
+			_generated_room.room_event = _room_event
+			
+			_grid[index_height].insert(index_width+_padding_size,_generated_room as RoomBase)
+	#print(_grid)
```
Tysterman74, line 26: Comment here, please remove if not needed.

Tysterman74 (changes requested):

Map/MapManager.gd
Turtyo: you are right, a range is probably better

Map/MapManager.gd
```diff
@@ -0,0 +1,29 @@
...
+		_padding.fill(null)
+		_grid.append([])
+		_grid[index_height].append_array(_padding)
+		
+		for index_width: int in range(floors_width[index_height]):## loop through elements of the grid and assign a room type
+			var _rand_type_index: int = randi_range(0, GlobalVar.EVENTS_CLASSIFICATION.size() - 1)
+			var _room_event: EventBase = GlobalVar.EVENTS_CLASSIFICATION[_rand_type_index].new()
+			var _generated_room: RoomBase = RoomBase.new()
+			_generated_room.room_event = _room_event
+			_grid[index_height].insert(index_width+_padding_size,_generated_room as RoomBase)
```
Turtyo, line 26: why use insert instead of append btw ?

Map/MapManager.gd
```diff
@@ -2,27 +2,28 @@ extends Node2D
 ## Class to manage backend for rooms (generation and such)
  
 #layout changes the width of the map's floors
-static func create_map(_layout: Array[int]) -> MapBase: ## Generates and Populates a map with rooms that have random room types. More in depth algorithms will be added in the future
+static func create_map(floors_width: Array[int]) -> MapBase: ## Generates and Populates a map with rooms that have random room types. More in depth algorithms will be added in the future
```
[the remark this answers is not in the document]
Tysterman74: Turtyo has a point, we have floors_width already defined in MapBase, so it may be beneficial to update this name so we don't run into any complications once we add this to the scene.

Tysterman74 (approved):
I'll approve since everything looks good from my end, I do have the same comment as Turtyo though.

Tests/test_map.gd
```diff
@@ -8,7 +8,9 @@ func test_map_gen():
 	var expected_null_array: Array[Array] = [[0,0], [0,1], [0,3],[0,4],[1,0],[1,4],[3,0],[3,4],[4,0],[4,1],[4,3],[4,4]]
 	var expected_exists_array: Array[Array] = [[0,2],[1,1],[1,2],[1,3],[2,0],[2,1],[2,2],[2,3],[2,4],[3,1],[3,2],[3,3],[4,2]]
 	for couple: Array[int] in expected_null_array:
-		assert_null(test_map.rooms[couple[0]][couple[1]], "expected null at"+str(couple))
-		
+		var _room = test_map.rooms[couple[0]][couple[1]]
+		assert_null(test_map.rooms[couple[0]][couple[1]], "Expected null at %s but got %s" % [couple, _room])
```
Turtyo, line 12: since you have the `_room` variable defined, you can use that instead of `test_map.rooms[couple[0]][couple[1]]` in the assert

Tests/test_map.gd
```diff
@@ -8,7 +8,9 @@ func test_map_gen():
...
+
 	for couple: Array[int] in expected_exists_array:
-		assert_not_null(test_map.rooms[couple[0]][couple[1]], "expected not null at"+str(couple))
+		var _room = test_map.rooms[couple[0]][couple[1]]
+		assert_not_null(test_map.rooms[couple[0]][couple[1]], "Expected not null at %s but got %s" % [couple, _room])
```
Turtyo, line 16: same logic as above, us `_room` in the `assert_not_null`

Turtyo (commented):
Will pull in local to test

#Scenes/TestingScene.tscn
```diff
@@ -11,13 +11,47 @@
...
 [ext_resource type="Resource" uid="uid://bsrdu33ukb1ym" path="res://Cards/Resource/Card_EnemyAttack.tres" id="11_3k5t2"]
 [ext_resource type="Texture2D" uid="uid://caemucaya30wh" path="res://Cards/draw_pile.png" id="11_pw70x"]
 [ext_resource type="Texture2D" uid="uid://d4muqvs3etnr8" path="res://Cards/discard_pile.png" id="12_kxw48"]
 [ext_resource type="Script" path="res://UI/EndTurnButton.gd" id="14_dpe64"]
+[ext_resource type="Texture2D" uid="uid://hqkt8t1v2f5h" path="res://Cards/deck_pile.png" id="18_b54hn"]
+[ext_resource type="Script" path="res://UI/DeckPileUISetter.gd" id="19_ucc6f"]
+
+[sub_resource type="Resource" id="Resource_82ci8"]
+script = ExtResource("10_qowiy")
+damage_to_apply_to_target = 0.0
```
Turtyo: also I notice that from `damage_to_apply_to_target` to `amount_of_cards_to_discard` came back in the merge
I believe this in an error that slipped through, could you please remove it?

Turtyo: Disregard this comment, it is not relevant since we are on the map and not the main

Turtyo (approved):

Turtyo:
I pulled in local to test, it loads and all the tests are passing. We don't have any visual for now but the code seems good, and this is a solid basis imo. Thank you @multi-arm !
@Tysterman74 we are just gonna need your approval again since need pushes were made. If you could also squash merge once that is done :+1:

Tysterman74 (approved):

Issue #36, opened 2023-12-27
Fixing error in merge where old code in CardBase was readded


JonaLam:
## Problem
When the prodject newly got merged, a bit of old code which had been removed when changing the way card effects were handled got put back in which adds a lot of bloat to the code and especially the resource where it could be annoying to make and edit cards.

## Solution
Remove the old system and make sure the new one works

PR #37, opened 2023-12-28
Fixing error in merge where old code in CardBase was readded 

JonaLam:
# Description

Removed an old card effect system which got readded in a merge

## Related issue(s)

#36 

## List of changes

Removed unecesary variables and functions from CardBase
Small changes in CardBase Recources to make some of them work, primaraly heal

### DIFF
--- a/Cards/CardBase.gd
+++ b/Cards/CardBase.gd
@@ -13,14 +13,6 @@ class_name CardBase
 ## wish to take damage in some contexts.
 ## For example, consider the card: "Deal 10 damage to all enemies, but take 3 damage"
 
-@export var damage_to_apply_to_target: float = 0.0
-@export var damage_to_apply_to_caster: float = 0.0
-@export var status_to_apply_to_target: Array[StatusBase]
-@export var status_to_apply_to_caster: Array[StatusBase]
-@export var affect_all_targets: bool = false
-@export var affect_all_casters: bool = false
-@export var amount_of_cards_to_draw: int = 0
-@export var amount_of_cards_to_discard: int = 0
 @export var application_type: Enums.ApplicationType = Enums.ApplicationType.ENEMY_ONLY
 @export var card_title: String = "NULL"
 @export var card_key_art: ImageTexture = null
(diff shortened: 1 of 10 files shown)


Cards/CardBase.gd
```diff
@@ -50,25 +42,9 @@ func can_play_card(caster: Entity, target: Entity) -> bool:
...
-	# damage target
-	if damage_to_apply_to_target != 0.0:
-		_damage_entity(caster, target, damage_to_apply_to_target, affect_all_targets)
-	
-	#damage caster
-	if damage_to_apply_to_caster != 0.0:
-		_damage_entity(caster, caster, damage_to_apply_to_caster, affect_all_casters)
-
-
 func _damage_entity(caster: Entity, target: Entity, damage_amount: float, damage_all: bool) -> void:
```
Turtyo, line 48: looking at my own PR, this got deleted. i think I put the effect in the `EffectDamage` card. Would like confirmation from someone that reviewed my initial PR (@Tysterman74 or @Tomzkk ) but i'm pretty sure you can also delete this function from CardBase

Turtyo (commented):
I will need to check that in local too.
When you modified the cards, did you do that via your text editor or did you use Godot's inspector ? Because looking at some of those resources, it seems like they might be missing some data to work (might just be me, as i said, i didn't test in local yet)

Cards/CardBase.gd
JonaLam: I can see if anything breaks if I remove it

JonaLam: Ye, it seems like it's fine without so I'll push that

Turtyo: perfect thanks !

Turtyo:
I tested in local. It works fine. You should just remove the Card_EnemyAttack.tres from the player deck to not have a card with NULL in hand. Also maybe damage all should be removed since it doesn't do anything currently

Cards/CardBase.gd
```diff
@@ -50,62 +42,5 @@ func can_play_card(caster: Entity, target: Entity) -> bool:
 
 func on_card_play(caster: Entity, target: Entity) -> void:
```
Tomzkk, line 43: Looking at #27, on_card_play was changed to only have a target and no longer have a caster. Was this change missed or was it at some point changed back to have both ?

Tomzkk (commented):
Tried to check by comparing what was removed from here with #27 and it looks like the removed stuff should be okay. Left a comment, I'm not entirely sure if that was changed at some later point or if that's a missing change. If it's missing there might be more like that ?

Cards/CardBase.gd
Turtyo: Good catch, there is no caster in this function for now. Though that will probably change with the introduction of stats

JonaLam: I did miss it, but if we're gonna use it later it should be fine to just keep it, right?

Turtyo: Mmh maybe we should check where the function is used to make use don't introduce a bug by keeping it.

Turtyo:
Let's only merge this one after #22 is merged to prevent another merge conflict on an old PR

Cards/CardBase.gd
JonaLam: I think it's only called in battler, and it's not like anything happens with the caster so it shouldn't make any bugs until we use the caster

Turtyo (approved):

Tysterman74 (approved):

Turtyo:
#22 was closed to be reopened in a new and cleaner PR, waiting for this new PR to merge #37

Turtyo:
After looking, merging this one now shouldn't give problems with 22 so we'll squash merge now

PR #39, opened 2024-01-01
Insta cast

JonaLam:
# Description

Added card dragging and insta cast
small rework to effect base to allow for target all effects

## Related issue(s)
 #21 

## List of changes

-Added dragging cards when pressed
-Added cast type enum which automaticly sorts cards into either insta cast or target
-instacast cards can be played by pressing anywhere above player's cards
-target cards move to the side when in cast zone so that the player can target who they want
-Added enum to EffectBase which allows for damageing only a single enemy or all enemies, as well as having cards that do both.
-Right clicking unqueues card


cheesefrycook:
Just pulled and tested, seems to work well! Will do my review in a bit

Cards/CardBase.gd
```diff
@@ -43,19 +43,44 @@ func _apply_all_effects(target: Entity) -> void:
...
+			all_target += [PlayerManager.player]
+		Enums.ApplicationType.ENEMY_ONLY:
+			all_target = CardManager.card_container.battler_refrence._enemy_list
+		Enums.ApplicationType.FRIENDLY_ONLY:
+			all_target = [PlayerManager.player]
+	#apply effect to every target
+	for entity : Entity in all_target:
+		for effect : EffectData in all_effects:
+			_apply_effect(entity, effect)
+
```
cheesefrycook, line 82: I think it will probably be better to move this logic into the Battler, then we could remove the `battler_reference` that we have here. However I believe Turtyo has mentioned there will be future changes to how targeting works. So for now this works fine I suppose.

Cards/CardContainer.gd
```diff
@@ -240,6 +244,7 @@ func _bind_card_input(card: CardWorld) -> void:
 	card_click_handler.on_click.connect(_on_card_clicked.bind(card))
 	card_click_handler.on_mouse_hovering.connect(_on_card_hovering.bind(card))
 	card_click_handler.on_unhover.connect(_on_card_unhovered.bind(card))
+	card_click_handler.on_right_click.connect(on_right_click.bind(card))
```
cheesefrycook, line 247: The problem with having the right click in the ClickHandler is that you always have to right click the card to cancel it. This is a bit annoying when you have a SINGLE_TARGET type card, because it's not going to be at your mouse position

cheesefrycook: So maybe we can check for the right click in `_process` using `Input.is_mouse_button_pressed` or something similar. This way it isn't checking for a click on a specific card

Cards/Movement State/MoveState_Queued.gd
```diff
@@ -23,10 +25,18 @@ func on_state_enter() -> void:
...
+	offset_desired_position += OFFSET_FROM_MOUSE
+	
+	var is_in_play_area : bool = CardManager.card_container.is_queued_card_in_play_area()
+	var is_insta_cast : bool = _state.card.card_cast_type != Enums.CardCastType.INSTA_CAST
 	
-	_state.card.position = _state.card.position.lerp(offset_desired_position, delta * MOVE_SPEED)
+	if(is_in_play_area && is_insta_cast):
+		_state.card.position = _state.card.position.lerp(TARGET_CARD_POS, delta * MOVE_SPEED)
+	else :
+		_state.card.position = _state.card.position.lerp(offset_desired_position, delta * MOVE_SPEED)
```
cheesefrycook, line 39: Maybe we should introduce a new `TARGETING` card state. So the idea is that when you place a card in the play area, if it's `SINGLE_TARGET`, then it enters the `TARGETING` state. Right now, the card will move to the `TARGET_CARD_POSITION` immediately when it enters the play area, which is a bit confusing to me.

cheesefrycook: Actually playing with it a bit more, I think it's fine in the current state. I'd be curious to hear other people's thoughts though

Cards/CardBase.gd
JonaLam: I guess, but it feels more intuitve for me that playing cards happens in the card script, but we can see what's better if changes happen

Cards/CardContainer.gd
JonaLam: Ye, that's fair, I'll do that

Cards/Movement State/MoveState_Queued.gd
JonaLam: I think it will feel better if we add some more effects to make it less confusing, I think it's better that it doesn't require an extra click

Turtyo:
Will check this tomorrow. i'm not too sure about "Added enum to EffectBase which allows for damageing only a single enemy or all enemies, as well as having cards that do both." for two reasons:
- This should be put in the general const file (which exists only in the map generation for now, but it's very useful so i might just do a small issue / PR for that only)
- i'll see how that plays with effects, because effects in themselves already have the ability to damage one or all (and you only need two of those effects to do both)

Maybe right click shouldn't unqueue card because it might be counter intuitive. We could unqueue the card if we click in some areas instead (hand, areas without enemies...). i believe the hovering / selecting a card already works in that way also.

Cards/Movement State/MoveState_Queued.gd
cheesefrycook: Yes I think effects will help a lot. I tested in StS as well and it works pretty much exactly like you have it here

Cards/CardBase.gd
```diff
@@ -35,12 +35,41 @@ func _apply_all_effects(target: Entity) -> void:
 	for effect_data: EffectData in card_effects_data:
 		effect_data.apply_effect_data(target)
 
+func _apply_effect(target : Entity, effect: EffectData):
```
Turtyo, line 38: `EffectBase` already has an `apply_effect` function, and this is not doing much, so I don't really see the goal of this function

Cards/CardBase.gd
```diff
@@ -35,12 +35,41 @@ func _apply_all_effects(target: Entity) -> void:
...
+	effect.apply_effect_data(target)
 
 func can_play_card(caster: Entity, target: Entity) -> bool:
 	return caster.get_party_component().can_play_on_entity(application_type, target)
 
 
-func on_card_play(caster: Entity, target: Entity) -> void:
-	_apply_all_effects(target)
+func on_card_play(caster: Entity, targets: Array[Entity]) -> void:
```
Turtyo, line 45: this honestly feels like it could be moved in the `_apply_all_effects` function, because this is basically what it does

Cards/CardBase.gd
```diff
@@ -35,12 +35,41 @@ func _apply_all_effects(target: Entity) -> void:
...
+	#Split up targeted attacks and all attacks
+	var target_effects: Array[EffectData] = []
+	var all_effects: Array[EffectData] = []
+	
+	for effect_data: EffectData in card_effects_data:
+		if(effect_data.target_type == Enums.TargetType.SINGLE_TARGET):
+			target_effects.push_front(effect_data)
```
Turtyo, line 52: why use `push_front` instead of `append` ? this will make effects play in reverse
for example if you have a card that says "Does A, then B, then C", when we read it the list will be `[A]` then we push B to the front it becomes `[B,A]` then `[C,B,A]`

Cards/CardBase.gd
Turtyo: I agree that matching the type should probably be moved to the battler and given to the card_play maybe. And yes, I will open an issue for the targeting i'm thinking of to see if we agree.
The problem of having all the logic by inside `on_card_play` is also that we will need to reformat it every time we want to change something about the logic. If instead we use functions that do things, we only need to change the sub-function without needing to change the card play in itself. `on_card_play` should only deal with having to play the card, and not knowing how to build the list of effects or knowing how to build the list of targets. Those things should be dealt in their respective functions, and `on_card_play` should only call them.

Cards/CardContainer.gd
```diff
@@ -393,3 +412,18 @@ func _update_card_positions() -> void:
...
+	_unfocus_card(card)
+	card.get_card_movement_component().set_movement_state(Enums.CardMovementState.IN_HAND)
+
+func is_queued_card_in_play_area() -> bool:
+	return get_global_mouse_position().y < play_at_height 
+
+func _handle_queued_card():
+	if(queued_card == null):
+		return
+	if(Input.is_mouse_button_pressed(2)):
```
Turtyo, line 427: what is the 2 here ?

Cards/CardContainer.gd
```diff
@@ -393,3 +412,18 @@ func _update_card_positions() -> void:
 		# set position and rotation
 		movement_component.state_properties.desired_position = Vector2(card_x, card_y)
 		movement_component.state_properties.desired_rotation = rotation_amount
+
+func un_queue_card(card : CardWorld):
```
Turtyo, line 416: wasn't all the queue-ing already done in functions that Cheesy made ?

Turtyo (commented):
I think some functions need to be broken down a bit to be more modular, and so that we don't need to modify their logic every time we change a little thing, and instead change the logic of the sub-functions

Turtyo:
@JonaLam will let you solve the merge conflict, i checked it's just the testing scene, shouldn't be hard

Cards/CardBase.gd
JonaLam: That's fair, I'll change it

Cards/CardBase.gd
JonaLam: I don't know why I did it like that either, I'll change it

Cards/CardBase.gd
JonaLam: That's a good point, I'll at least move some of the logic out to other functions

Cards/CardContainer.gd
JonaLam: I don't think so, I couldn't find anything like that, and the on click function didn't call anything like that so I don't know

Cards/CardContainer.gd
JonaLam: It's for right clicking to cancel, I'll comment it at least for readability

Turtyo: using a variable name would be even better

Cards/CardBase.gd
```diff
@@ -31,29 +31,49 @@ func parse_card_data(card_data: Dictionary) -> void:
...
+	var target_effects: Array[EffectData] = get_target_effects()
+	var all_effects: Array[EffectData] = get_all_effects()
+	
+	#Apply single target
+	_apply_all_effects(targets, target_effects)
+	#Get every unit that is to be affected by card
+	var all_target : Array[Entity] = get_all_targets()
+
+	#apply effect to every target
+	_apply_all_effects(all_target, all_effects)
```
Turtyo, line 54: doesn't this assume all the effects have the same targets ? this might not be the case for a card like "does 1 damage to a selected enemy, heal 1 to self"

Cards/CardBase.gd
```diff
@@ -31,29 +31,49 @@ func parse_card_data(card_data: Dictionary) -> void:
...
-	#Apply single target
-	for entity : Entity in targets:
-		for effect : EffectData in target_effects:
-			effect.apply_effect_data(entity)
-	#Get every unit that is to be affected by card
+			target_effects.append(effect_data)
+			
+	return target_effects
+
+func get_all_effects() -> Array[EffectData]:
```
Turtyo, line 67: the name is a bit ambiguous. I would expect it to mean "all the effects of the card" but you mean it as "all the effects that apply to a target ALL"
`get_effects_target_all` would probably be a clearer name

Cards/CardContainer.gd
Turtyo: @cheesefrycook can you confirm ? I thought I saw something like that in your PR ? or was it a discussion we had but chose to deal with later ?

Cards/CardBase.gd
JonaLam: Right now, they do all have the same target, and it's kinda difficult to affect both sides with one card. There should definitaly be some changes to this, but maybe it should happen with the target changes

Cards/CardBase.gd
JonaLam: I can change that

Cards/CardBase.gd
Turtyo: did you check #42 in link with this ?

JonaLam: Yeah, I saw, I was just thinking that fixing this issue would be better to do in #42, but I could do a bit of refactoring to enable the changes more

Cards/CardContainer.gd
cheesefrycook: Unqueuing a card previously you had to click the card, so I handle that behavior in _on_card_clicked. So this function is useful if you want to unqueue without clicking I guess.

Cards/CardBase.gd
Turtyo: thank you

JonaLam: I can't actually think of a good way of fixing this without just refactoring the whole thing, so I think I'll leave it to #42

Turtyo: alright fair enough, this might be a bit out of the scope anyway

Turtyo:
@JonaLam don't forget to solve the merge conflict with the main

cheesefrycook (dismissed):

Cards/CardBase.gd
Turtyo: @JonaLam unresolving this because i see it's back to the name I had a problem with

Cards/CardBase.gd
```diff
@@ -31,16 +31,45 @@ func parse_card_data(card_data: Dictionary) -> void:
...
 func can_play_card(caster: Entity, target: Entity) -> bool:
 	return caster.get_party_component().can_play_on_entity(application_type, target)
 
 
-func on_card_play(caster: Entity, target: Entity) -> void:
-	_apply_all_effects(target)
+func on_card_play(caster: Entity, targets: Array[Entity]) -> void:
+	#Split up targeted attacks and all attacks
+	var target_effects: Array[EffectData] = get_target_effects()
+	var all_effects: Array[EffectData] = get_all_effects()
```
Turtyo, line 46: would rename to `effects_targeting_all`

Cards/CardBase.gd
```diff
@@ -31,16 +31,45 @@ func parse_card_data(card_data: Dictionary) -> void:
...
+	#Get every unit that is to be affected by card
+	var all_target : Array[Entity] = CardManager.card_container.battler_refrence.get_all_targets(application_type)
+
+	#apply effect to every target
+	_apply_all_effects(all_target, all_effects)
+	
 	CardManager.on_card_action_finished.emit(self)
-	# TODO add other functionality that lots of cards may share (eg: restore AP)
+
+func get_target_effects() -> Array[EffectData]:
```
Turtyo, line 58: would rename to `get_effects_targeting_single`

Cards/CardContainer.gd
```diff
@@ -284,26 +289,40 @@ func _on_phase_changed(new_phase: Enums.Phase, _old_phase: Enums.Phase) -> void:
...
 	else:
 		# If we click a card with no card queued, queue it
 		set_queued_card(card)
 
 		card.get_card_movement_component().set_movement_state(Enums.CardMovementState.QUEUED)
 		_focus_card(card)
 
+func _play_card():
+	var queued_card_data: CardBase = queued_card.card_data
+	var target : Array[Entity]
```
Turtyo, line 318: maybe naming it `list_targets` would be less confusing

Cards/Resource/Card_DamageAll.tres
```diff
@@ -1,20 +1,10 @@
...
-value = 2
 
 [resource]
 script = ExtResource("1_j02oq")
-application_type = null
-card_title = "Damage All"
-card_description = "Deal 2 damage to all enemies"
-card_effects_data = Array[ExtResource("2_kr6on")]([SubResource("Resource_s752v")])
+application_type = 1
+card_title = "NULL"
```
Turtyo, line 8: why set the card title to null ?

Cards/Resource/Card_DamageAll.tres
```diff
@@ -1,20 +1,10 @@
...
+card_description = "NULL"
```
Turtyo, line 9: missing description

Cards/Resource/Card_DamageAll.tres
```diff
@@ -1,20 +1,10 @@
...
+card_effects_data = Array[Resource("res://Cards/Effects/EffectData.gd")]([])
```
Turtyo, line 10: This looks like the array is empty, meaning the card doesn't do anything

Cards/Resource/Card_Damage_EVERYTHING.tres
```diff
@@ -0,0 +1,21 @@
...
+[sub_resource type="Resource" id="Resource_4ktkc"]
+script = ExtResource("2_581pa")
+effect = SubResource("Resource_16eay")
+value = 10
+target_type = 1
+
+[resource]
+script = ExtResource("1_lia0m")
+application_type = 2
+card_title = "NULL"
```
Turtyo, line 19: missing `card_title`

Turtyo (changes requested):

#Scenes/TestingScene.tscn
```diff
@@ -136,22 +133,4 @@ text = "0"
...
-offset_right = -22.0
-offset_bottom = 75.0
-texture_normal = ExtResource("18_b54hn")
-texture_pressed = ExtResource("18_b54hn")
-texture_hover = ExtResource("18_b54hn")
-texture_disabled = ExtResource("18_b54hn")
-texture_focused = ExtResource("18_b54hn")
-stretch_mode = 0
-script = ExtResource("19_ucc6f")
-metadata/_edit_use_anchors_ = true
```
Turtyo, line 155: why was this all removed btw ?

Cards/CardBase.gd
Turtyo: After reading again, it does more than just `_apply_all_effects, so it's fine

Cards/CardBase.gd
Turtyo: Could you give us a little summary of what you moved out ? I don't remember how it was looking at that point

Cards/CardContainer.gd
```diff
@@ -425,7 +425,9 @@ func _handle_queued_card():
 	if(queued_card == null):
 		return
 		
+	var rightMouseButtonIndex : int = 2
```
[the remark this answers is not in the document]
JonaLam: Do we have a file like that?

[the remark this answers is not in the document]
Turtyo: I don't think we do. My PR has a `GLOBAL_VAR.gd` file but that's a bit different.
Basically you go to project settings, autoload and you create a new node file from there then populate it with the variables you need. when you want to call it you do `your_file_class_name.variable`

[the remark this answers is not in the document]
JonaLam: Do you think it would be better to just put it on the input map instead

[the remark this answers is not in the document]
Turtyo: the input map ? i don't remember a file named like that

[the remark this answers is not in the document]
JonaLam: it's a in the prodject settings and I think it would do about the same as the sulution you were proposing, just built into godot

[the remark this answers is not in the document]
Turtyo: I see it allows reconfiguration also. This is a good idea !

Turtyo (changes requested):
`test_health.gd` must be updated to use `Array[Entity]` instead of just `Entity` when using `on_card_play`
The default deck should also use a heal card to test that the insta-cast works on self also (I think it does given Damage EVERYTHING but I prefer to be sure)

Cards/CardBase.gd
JonaLam: I moved out the filtering of single and all target attacks into get_effects_targeting_single and get_effects_target_all respectively, as well as how the function got the array of every target into get_all_targets in the battle manager

cheesefrycook (dismissed):

Turtyo (changes requested):
Those 3 new commits don't address the initial issue that I reported ie all the `on_card_play` calls in `test_health.gd` must be updated to be given an `Array[Entity]` instead of just an `Entity`, given that the insta-cast works with a list of entities.

Do you not run the tests in local and check the stack trace ? This would allow you to find the errors easily. **Please run the GUT tests in local in the future.** 

I need to check why they are passing with the Github action, but this should be more of a safety net than the main way to run the tests anyway.

Cards/CardBase.gd
Turtyo: alright seems nice enough
changes will be made probably due to the targeting PR, but this looks okay
@cheesefrycook feel free to resolve if you feel like it's ok

cheesefrycook (approved):

Cards/CardContainer.gd
```diff
@@ -393,3 +412,20 @@ func _update_card_positions() -> void:
...
+
+func is_queued_card_in_play_area() -> bool:
+	return get_global_mouse_position().y < play_at_height 
+
+func _handle_queued_card():
+	if(queued_card == null):
+		return
+		
+	#check if right button is pressed
+	if(Input.is_action_just_pressed("cancel_queued_card_in_battle")):
```
Tysterman74, line 429: We're calling this in one spot, which is fine for now, but if any other file is going to utilize this action, we would want to create a global file to hold this in, so we can update it in one spot.

Tysterman74 (changes requested):

Cards/CardContainer.gd
Turtyo: I think it's a bit different because of the input map which already holds this

JonaLam: I'm a bit confused, what is it that you want changed here?

Turtyo: the fact you used a string "cancel_queued_card_in_battle" instead of using a variable that holds the string
but in fact, the string is linked to a variable (kinda) with the input map, meaning we already have a central place to hold changes to multiple instances if needed
the only problem we could have is if we decide we don't want to use the name "cancel_queued_card_in_battle", but technically that would be equivalent to changing the name of the variable in itself if we used one (so same problem).

At least if I understood correctly the input map

JonaLam: So you think it's fine as is?

Turtyo (approved):

Cards/CardContainer.gd
Tysterman74: I think this is fine for now, but if we see this being used in another file in another PR we should move it out to a global file.

Tysterman74 (approved):

PR #40, opened 2024-01-02
CardPileUicl

Paper-2:
# Description

Adds a UI popup that shows all the cards in the draw or discard pile on click. It will show all the cards in a grid.

## Related issue(s)

#18

## List of changes

* Adds a UI that shows the content of the discard/Draw Pile.

* Adds Deck view

## Tests

test_pile_ui test if the number of cardbases is equal to number of children the gridcontainer


Turtyo:
Thank you :+1:

#Scenes/TestingScene.tscn
```diff
@@ -11,13 +11,47 @@
...
+[ext_resource type="Script" path="res://UI/DeckPileUISetter.gd" id="19_ucc6f"]
+
+[sub_resource type="Resource" id="Resource_82ci8"]
+script = ExtResource("10_qowiy")
+damage_to_apply_to_target = 0.0
+damage_to_apply_to_caster = 0.0
+status_to_apply_to_target = Array[Resource("res://Status/StatusBase.gd")]([])
+status_to_apply_to_caster = Array[Resource("res://Status/StatusBase.gd")]([])
+affect_all_targets = false
+affect_all_casters = false
```
Tysterman74, line 30: These fields have been removed in #37 I believe. Should we wait for that PR to be merged in first? @Turtyo

Tests/test_pile_ui.gd
```diff
@@ -0,0 +1,44 @@
...
+var _card_container_scene: PackedScene = load("res://Cards/CardContainer.tscn")
+var _card_container_scroll_scene: PackedScene  = load("res://#Scenes/CardScrollUI.tscn")
+var _card_container: CardContainer = null
+var _card_scroll: Control = null
+
+const SCROLLCONTAINER = 2
+const GRIDCONTAINER = 0
+
+
+func _test_setter():
```
Tysterman74, line 13: Instead of calling this `_test_setter()` on the `test_populate()` function, you can override `before_all()` which is an override in GutTest that will do initial setup of stuff before running each test! This'll basically just remove `_test_setter()` from `test_populate()` and automatically provide the setup for any other tests we may make in this file.

https://gut.readthedocs.io/en/latest/Creating-Tests.html

Tests/test_pile_ui.gd
```diff
@@ -0,0 +1,44 @@
+extends GutTest
+#test for ui pop up
+
+var _card_container_scene: PackedScene = load("res://Cards/CardContainer.tscn")
+var _card_container_scroll_scene: PackedScene  = load("res://#Scenes/CardScrollUI.tscn")
+var _card_container: CardContainer = null
+var _card_scroll: Control = null
+
+const SCROLLCONTAINER = 2
+const GRIDCONTAINER = 0
```
Tysterman74, line 10: Maybe a nitpick here, but I do like to separate words with `_` so in this case `SCROLL_CONTAINER` and `GRID_CONTAINER`

Tysterman74 (changes requested):

#Scenes/TestingScene.tscn
Turtyo: ah that's what you were talking about ! It's fine, we can merge this in and then #37 that will prevent having to merge the main into this PR again
If it's just something like that and not Paper using explicitly something deleted then it's fine

Tests/test_pile_ui.gd
Turtyo: yes indeed ! I didn't explicitly add it in the `CONTRIBUTING.md` in the formatting part, but this is the format of the [GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#trailing-comma)

UI/CardPileUISetter.gd
```diff
@@ -22,3 +22,9 @@ func _on_card_counts_updated() -> void:
 
 func _get_count() -> int:
 	return CardManager.card_container.get_draw_pile_size()
+
+func _pressed() -> void:
+
+	var uiPile = cardUI.instantiate()
```
Turtyo, line 28: type of uiPile ?

UI/CardPileUISetter.gd
```diff
@@ -22,3 +22,9 @@ func _on_card_counts_updated() -> void:
...
+	uiPile.populate(get_name())
+	$"..".add_child(uiPile)
```
Turtyo, line 30: maybe a `var parent = $".."` then `parent.add_child` would be a bit less cryptic. Not sure that syntax is possible though

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,54 @@
...
+			cards_to_display = discard_pile
+			$Label.text = "Showing the discard pile"
+		"DrawPile":
+			cards_to_display = draw_pile
+			$Label.text = "Showing the draw pile"
+		"DeckPile":
+			cards_to_display = deck_pile
+			$Label.text = "Showing the deck pile"
+	
+	cards_to_display.sort()
```
Turtyo, line 40: what are they sorted by ? how does the sort function works for an array of `Cardbase` ?
I don't think the sort function is implemented by default for us, and i didn't see an implementation either. My guess is that it currently sort based on object index, which for the player might not have any logic.

Some things we can do for now:
- add a sort function (call it `card_sorting`)
- use this on the array of cards
- good thing is that now we can change the internal logic of our sorting function without having to touch this code again (we could do alphabetical order, obtained order, by type etc...)

UI/CardScrollUI.gd
```diff
@@ -0,0 +1,54 @@
...
+
+const CARD_SCALE: Vector2 = Vector2(.7, .7)
+const CARDUI_INDEX: int = 2
+const CARDUI_POS: Vector2 = Vector2(4,4)
+const SIZE_OFFSET: Vector2 = Vector2(10,10)
+
+func _on_button_pressed() -> void:
+	queue_free()
+
+func _input(_inputevent: InputEvent) -> void:
```
Turtyo, line 21: so that removes all the cards from the display when called right ?

Tests/test_pile_ui.gd
```diff
@@ -0,0 +1,43 @@
+extends GutTest
+#test for ui pop up
+
+var _card_container_scene: PackedScene = load("res://Cards/CardContainer.tscn")
+var _card_container_scroll_scene: PackedScene  = load("res://#Scenes/CardScrollUI.tscn")
+var _card_container: CardContainer = null
+var _card_scroll: Control = null
+
+const SCROLL_CONTAINER = 2
+const GRID_CONTAINER = 0
```
Turtyo, line 10: not sure to understand what those two are as i didn't see them in other places

Tests/test_pile_ui.gd
```diff
@@ -0,0 +1,43 @@
...
+
+
+func before_all():
```
Turtyo, line 13: should add an `after_all` to free the queue. Maybe even doing a `before_each` and `after_each` would be better in case we add more tests that would need a clean environment between each of them

Tests/test_pile_ui.gd
```diff
@@ -0,0 +1,43 @@
...
+	
+
+func test_populate():
+	
+	_card_scroll.populate("DeckPile")
+
+	var cards_in_pile: int = len(_card_scroll.deck_pile)
+	var grid: GridContainer = _card_scroll.get_child(SCROLL_CONTAINER).get_child(GRID_CONTAINER)
+
+	assert_eq(cards_in_pile, grid.get_child_count())
```
Turtyo, line 40: yeah since it's visual, i guess there is not much that can be tested. Maybe could also assert that every card in cards_in_pile is in the grid but i'm not sure how (because those are two different objects right ?)
Maybe best way to test that is to look at it, though it would be better if it was automated

UI/CardScrollUI.gd
Paper-2: Yes, it deletes the root node CardScrollUI to be more specific.

Tests/test_pile_ui.gd
Paper-2: Those are the indexes to get their node. I can the INDEX as a suffix.

UI/CardScrollUI.gd
Paper-2: The cards cards can only be sorted by title for now, as there's nothing storing their type or a history of card obtained. I will make a function named card_sort_by_title and use  the sort_custom fuction to applied it.

Tests/test_pile_ui.gd
Paper-2: It's not just that is visual. Godot doesn't allow the children of a container to have their own position, instead they return the position of the parent. If that wasn't the case then I could make the test check if the cards are collading with each other. Right now, the test is just checking if the amount of children equals the lenght of deck_pile.

#Scenes/TestingScene.tscn
Tysterman74: Roger that, I will resolve.

Tysterman74 (dismissed):

UI/CardScrollUI.gd
Turtyo: maybe a little comment would be good

Tests/test_pile_ui.gd
Turtyo: What about checking that every card in the grid is a card in the list of cards ?

Turtyo (approved):
Approved but don't forget to address the last few comments that are left open

Turtyo (approved):

Turtyo:
When running the GUT tests in local, I get hit with an error. In `test_pile_ui.gd`:
```GDScript
func before_each():
	_card_container = _card_container_scene.instantiate()
	_card_scroll = _card_container_scroll_scene.instantiate() # <--- error here
```
it calls:
```GDSript
var discard_pile: Array[CardBase] = CardManager.card_container.discard_pile
```
in `CardScrollUI.gd` "Invalid get index 'discard_pile' (on base: previously freed)

This might be due to creating then freeing then creating again as the error suggests

Turtyo (commented):
Tests are passing now, but I have a total of 458 new orphans in the test (and 463 in total), which I assume means they are not being freed correctly ?

Turtyo:
After testing in local, there are several problems:
- Only one card is given at a time
- No cards are shown if i click on a pile
- The return button does not work (only the escape key)
- I can't un-select a card that i selected in my hand

Which is why it's important to look at the game and not just the actual code i guess, i'll think about it before approving next time :sweat:

Tests/test_pile_ui.gd
```diff
@@ -0,0 +1,50 @@
...
+	for card_index: int in _card_container.default_deck.size():
+		_card_container.default_deck[card_index] = CardBase.new()
+		_card_container.default_deck[card_index].card_title = "Card" + str(card_index + 1)
+	
+	get_tree().root.add_child(_card_container)
+	
+
+func after_each():
+	
+
```
Tysterman74, line 32: Some extra whitespace here.

Tysterman74 (approved):

Turtyo (approved):
Alright now everything seems to be working fine, no orphaned nodes in the test or when playing the game, everything is displaying.
Only problem is that we are using free instead if `queue_free` in the test so the behaviour is a bit different between the test and the actual game

Tysterman74 (approved):

Issue #35, opened 2023-12-27
Killing enemies / player and combat end


Turtyo:
## Feature Description

Make it so that reducing enemies to 0HP removes them from combat. Killing all enemies ends the combat.
Killing the player ends the combat.

## Motivation

Well... We need to make progress at one point and being stuck in the first combat because enemies want to chat while at 0HP doesn't seem like a good thing.

## Acceptance Criteria

[Define the criteria that need to be met for this feature to be considered complete.]

- Killing an enemy removes it from the list of enemies
- Killing all enemies ends the combat
- Killing the player ends the combat

## Proposed Solution

We need to check after every attack (and maybe even after every effect, but that part depends on how we want to do things, this is more a game design thing) if an enemy is at 0HP or less. Same for the player.
If he is, remove him from the combat and the list of enemies. If the list of enemies is empty, combat ends. We don't have anything to do after for now, but we can imagine having a **state machine to go between the different possibilities**. For now the state machine can just be a loop of combats, and we will add more states as the development goes on. This is at least good to test that the gameplay loop works correctly (no bug between combats like badly initialized enemies, keeping stats from previous combat, etc...)

## Additional Context

The enemies wanted to chat a bit longer... 
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/81eb99c7-c2ba-4dce-b715-524987aab02f)
Almost wanted to qualify this one as a bug but technically we just never talked about this part. I added some context as a TODO list at the end of `Battler.gd` so someone that takes this can refer to that for later.

cheesefrycook:
Maybe we can check for the health depleting to zero in `Battler` since it has the list of enemies. From there we could send a signal to a global AutoLoad `SceneController` that handles whatever to do next. For now the `SceneController` could just reload `TestingScene` to confirm it works, but later on we could have the state machine as mentioned. Something like:
- Player death: show player death screen. Return to main menu when you click Next
- Enemies dead: show rewards screen. When rewards are collected, transition to map screen

Might also be good to add a new phase to `PhaseManager` called `GAME_OVER` that disables any combat interactions.

Turtyo:
Yes i was thinking something like that. I need to push my changes to battler now that i'm back. So that we are working too much on battler at the same time with @Tomzkk

Tomzkk:
I took a break from working on this waiting on that PR you mentioned before. @Turtyo 

[...]

PR #43, opened 2024-01-09
Use an object to store modifiers associated with a type of card / item effect

### LINKED ISSUE #4: Make stats changes related to character instead of calculating them at attack time
**Is your feature request related to a problem? Please describe.**
Currently (as of [this PR](https://github.com/Saplings-Projects/1M_sub/pull/1), the modified stats of a character are calculated inside the attack function by returning a modified version of the stats. This doesn't feel right for several reasons:
- Stats are calculated inside HealthComponent.gd which is an unusual place to find it.
- Stats are calculated by a function which is here to deal damage to characters, this is not a good place for it.
- Stats are calculated by returning a copy of the stats to not modify them. This is the only variable that is not modified and a copy of it is made instead.

**Describe the solution you'd like**
It would be good to have the stats change be calculated in StatsComponent.gd instead of HealthComponent.gd
This implementation would have buffs with an `on_application` and a `on_removal`, which would trigger the modification of stats once a buff is applied or expires.

**Additional context**
Discussed in [this comment](https://github.com/Saplings-Projects/1M_sub/pull/1#discussion_r1412389425)

### LINKED ISSUE #7: Use an object to store modifiers associated with a type of card / item effect

Turtyo:
# Description

This PR reworks the stat system to use a central dictionary for each entity, to hold their stats.

## Related issue(s)

Closes #7 
Closes #4 (I think ?)

## List of changes

- Added the `StatModifiers` class, this is the most basic element for stats
- Added the `StatDictBase` class, it is a dictionary of `StatModifiers`, with all possible modifiers (those will be moved to a global file later)
- Added an offense and defense dictionary to `EntityStats`
- Updated the functions to take stats into account
- Updated the Battler
- Updated Effects to take stats into account

## Tests

- Test `Strength`, `Weakness`, `Vulnerability`, `Buff_Poison_Duration` and mix of those
- Test removing status
- Test reset of temporary stats


Turtyo:
And tests are not passing because I didn't update the status properly yet

Cards/Effects/EffectBase.gd
```diff
@@ -13,4 +13,20 @@ func build_target_list(target: Entity) -> Array[Entity]:
 
 # To be called by child classes through apply_effect. But since there might be multiple effects that need it, it's available in the parent class
 func _apply_status(status: StatusBase, caster: Entity, target: Entity) -> void:
-    target.get_status_component().add_status(status, caster)
\ No newline at end of file
+    target.get_status_component().add_status(status, caster)
+
+func calculate_value(name: String, caster: Entity, target: Entity, value: int) -> int:
```
Tysterman74, line 18: using a String for "draw" and "damage" feels a bit too loose. Maybe adding an enum of like `CardType` maybe better for readability?

Cards/Effects/EffectDraw.gd
```diff
@@ -3,4 +3,5 @@ class_name EffectDraw extends EffectBase
 # @Override
 @warning_ignore("unused_parameter")
```
Tysterman74, line 4: We can remove this suppression, it looks like you're using all parameters.

Cards/Effects/EffectDamage.gd
```diff
@@ -3,10 +3,12 @@ class_name EffectDamage extends EffectBase
 # @Override
 func apply_effect(caster: Entity, target: Entity, value: int) -> void:
 	var target_damage_data: DealDamageData = DealDamageData.new()
-	target_damage_data.damage = value
-	target_damage_data.caster = caster
+	# calculate modified damage given caster and target stats
+
+	target_damage_data.damage = calculate_value("damage",caster, target, value)
+	target_damage_data.caster = caster # is this useful now that we calculate the stats before anyway ?
```
Tysterman74, line 9: I don't think we need `DealDamageData` any do we? We have access to the health via the Entity object, and since we make different Effects that does unique things EffectDamage should have the logic to subtract the health.

Cards/Effects/EffectBase.gd
Turtyo: I totally agree, i was thinking the same when i was writing it.

The reason for this is that in my dictionary of possible modifiers, it's (String, Data), so I need to use strings (and an enum is int). So i'm not exactly sure, I guess I could maybe change the possible modifiers to be from an enum instead, this'll probably be the best.

Cards/Effects/EffectDamage.gd
Turtyo: Yeah I was thinking the same. Just wanted to make sure. `DealDamageData` was useful when some logic was left to the `deal_damage` function, but now it's treated as any other effect.

Turtyo: Maybe this should be done at another time ? because there are a lot of places that have `DealDamageData` (like the tests for example).
On the other hand, the more we wait, the more things are built on type of it and the more changes will need to be made.

Turtyo:
Just realized that the `on_apply` function might need some changes actually

Turtyo:
I still have some more tests to add (and probably some debugging to do as a result), because i didn't test the remove and reset functions properly yet
Also note that this is only GUT tests, I haven't actually "played" the game to see if it works. The reason for that is that I thought adding a little issue to add effects that apply status might be good if someone wants to take a look at the current state of Effects/Status/Stats but don't want something too big to do.

Turtyo:
Also using an enum is nice when you write what type of stat you want to change in the code (because it's just `something_something.damage` for example). But for debugging it's actually a pain, because when you look at the stack it will show you the value associated to the name instead of the name, so i need to remember that 0 is damage, 1 is poison, 2 is draw etc...)
idk if there is a way to have the name shown when debugging, but that would be helpful

Cards/Effects/EffectDamage.gd
Tysterman74: I'll let you make the final say if you want to refactor this out now or not! If we don't let's make an issue to clean this up.

Tysterman74 (changes requested):

Cards/Effects/EffectDamage.gd
Turtyo: I think it's out of the scope of the PR, because I need to modify more than stats and status
I might take said issue myself, but I will make an issue for it i think

Cards/Effects/EffectBase.gd
```diff
@@ -3,14 +3,30 @@ class_name EffectBase extends Resource
...
-    return [target]
+	return [target]
 
 # To be called by child classes through apply_effect. But since there might be multiple effects that need it, it's available in the parent class
 func _apply_status(status: StatusBase, caster: Entity, target: Entity) -> void:
-    target.get_status_component().add_status(status, caster)
\ No newline at end of file
+	target.get_status_component().add_status(status, caster)
+
+static func calculate_value_modified_by_stats(enum_name: int, caster: Entity, target: Entity, value: int) -> int:
```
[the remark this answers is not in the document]
Turtyo: can you have an enum as type ?

Entity/Components/Stats/EntityStats.gd
```diff
@@ -0,0 +1,61 @@
...
+
+	return ceil(modified_value)
+
+func _calculate_modified_value_defense(enum_name: int, base_value: int) -> int:
+	var modified_value: int = base_value
+	var modifier: Dictionary = defense_modifier_dict.stat_dict[enum_name].modifiers
+	modified_value -= modifier["temporary_add"]
+	modified_value -= modifier["permanent_add"]
+	modified_value /= modifier["temporary_multiply"]
+	modified_value /= modifier["permanent_multiply"]
```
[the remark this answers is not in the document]
Turtyo: you are write, I should even update it in the `StatModifiersBase` then

Entity/Components/Stats/StatModifiers.gd
```diff
@@ -0,0 +1,32 @@
...
+
+func reset_temp_to_default() -> void:
+	modifiers["temporary_add"] = 0
+	modifiers["temporary_multiply"] = 1
+
+func change_modifier(new_modification: StatModifiers) -> void:
+	modifiers["permanent_add"] += new_modification.modifiers["permanent_add"]
+	modifiers["permanent_multiply"] *= new_modification.modifiers["permanent_multiply"]
+	modifiers["temporary_add"] += new_modification.modifiers["temporary_add"]
+	modifiers["temporary_multiply"] *= new_modification.modifiers["temporary_multiply"]
```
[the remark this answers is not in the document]
Turtyo: yes I was thinking that
we don't have a global var file for now because it's in the PR for map generation, but actually I should create one here following the same syntax

Turtyo:
Apart from the things you pointed out in your review @Tysterman74 I have a few other problems with the state of the code as is, and I would like your opinion on this:
- The functions name are confusing, and a lot of them do similar things. You have `change_stat`, `change_modifier_of_given_name`, `change_modifier`, `_modify_stats_with_status, `calculate_value_modified_by_stats`, `calculate_modified_value`; Even I lose myself because the functions are in different files and I don't always remember where each function is, because it's not always clear what they do. I think I should change the names and also moves them to other files (for example there is a function to calculate stat modifications in `EffectBase`, that will probably move)
- Generally speaking, I find the structure that I built a bit confusing, not sure if that's only due to the functions, but it's not really easy to understand what everything does

Entity/Components/StatusComponent.gd
```diff
@@ -3,34 +3,49 @@ class_name StatusComponent
 ## Applies status to the entity and holds a list of current status.
 
 
-var current_status: Array[StatusBase]
+var current_on_turn_start_status: Array[StatusBase] = []
+var current_on_apply_status: Array[StatusBase] = []
```
cheesefrycook, line 7: What's the purpose of splitting these arrays? Could we just assume that any status could do both of these and call the functions for all of them?

Entity/Components/Stats/StatModifiers.gd
```diff
@@ -0,0 +1,32 @@
+class_name StatModifiers extends Resource
+
+var modifiers: Dictionary = {}
+
```
cheesefrycook, line 4: So just to confirm my understanding of the design here:
- temporary = lasts until combat end
- permanent = buffs your character permanently
So I'm guessing the idea behind permanent buffs is you get them from stuff like artifacts that last for the entire run?

Status/Debuffs/Debuff_Vulnerability.gd
```diff
@@ -3,6 +3,14 @@ class_name Debuff_Vulnerability
...
+	is_stat_modification = true
+	is_on_apply = true
+	modifier_name = StatDictBase.POSSIBLE_MODIFIER_NAMES.damage
+
+# @Override
+func init_status(in_caster: Entity, in_target: Entity) -> void:
+	status_caster = in_caster
+	status_target = in_target
+	targeted_modifier_dict = in_target.get_stat_component().get_stats().defense_modifier_dict
+
```
cheesefrycook, line 16: Maybe instead of having the `is_stat_modification` and `is_on_apply` flags, we can override `on_apply` in any of these stat modifying statuses and call `_modify_stats_with_status` from the parent

cheesefrycook: I suppose you would have to override `on_remove` as well, to make sure the stat change is removed

Entity/Components/Stats/EntityStats.gd
```diff
@@ -0,0 +1,61 @@
...
+## We can increase these stats with buffs, items, or equipment. We shouldn't modify the stats
+## directly, but instead create a modified version of the stats whenever we need them. This is so
+## we don't have to worry about reverting the state of the stats whenever the buff/item/equipment
+## is removed. See get_modified_stats on BuffBase for an example.
+
+## The previous statement will probably change, since we want to modify stats at runtime and we revert the changes when effects are removed.
+
+var modification_count: int = 0
+var offense_modifier_dict: StatDictBase = null
+var defense_modifier_dict: StatDictBase = null
```
cheesefrycook, line 15: So `offense_modifier_dict` is what we modify for buffs/debuffs that increase/decrease offensive power, while `defense_modifier_dict` is what we modify for buffs/debuffs that increase/decrease defensive power?

How does a buff like `draw 2 more cards` fit into one of these two dictionaries?

Entity/Components/Stats/EntityStats.gd
```diff
@@ -0,0 +1,61 @@
...
+	modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
+	modification_count += 1
+
+func _calculate_modified_value_offense(enum_name: int, base_value: int) -> int:
+	var modified_value: int = base_value
+	var modifier: Dictionary = offense_modifier_dict.stat_dict[enum_name].modifiers
+	modified_value += modifier["temporary_add"]
+	modified_value += modifier["permanent_add"]
+	modified_value *= modifier["temporary_multiply"]
+	modified_value *= modifier["permanent_multiply"]
```
cheesefrycook, line 40: Does order of operations come into play here?

Entity/Components/StatusComponent.gd
Turtyo: The idea behind splitting is that at the start of the turn, you can do one loop on the turn_start array, apply the effect and reduce by one the duration, and one more loop on the apply_status to just decrease the duration
This allows to not have to check if each status is on_apply. But this was before I even made the `on_apply` boolean for each status. In the current structure, we could totally have a single status_array and check for each status if they have an effect at_turn_start (might need to add a boolean for that though).

The current structure has one flaw, it doesn't allow effects to both have an effect on apply and at turn start, but I didn't find an effect that did such thing (i'm not sure if poison counts ? because technically it's the effect that adds poison stacks, not the status itself that adds poison stacks on apply, and thus it would just be a turn_start status ?)

Entity/Components/Stats/StatModifiers.gd
Turtyo: Exactly. Not to say you can't lose the buff (for example if you lose the object for whatever reason), but as you can see in the reset function, I only reset the temporary modifiers.

Status/Debuffs/Debuff_Vulnerability.gd
Turtyo: I'm not sure to see right now, i'll check the code in more details later

Entity/Components/Stats/EntityStats.gd
Turtyo: It will be stored in the offense dictionary.
The reason is that the value modified by stats (whatever the effect and the stats) is calculated like that:
`base_value + caster_attack_stat - target_defense_stat`
So let's say we buff ourselves with 2 more draw. We add 2 to our draw stat in our `offense_modifier_dict`. We then cast draw 1 card on ourselves. The caster is us, and the target is us
We add the `caster_attack_stat` (which is 2), and we deduce the `target_defense_stat` (which is 0 because we didn't touch our `defense_modifier_dict`).
So the total is:
`1 + 2 - 0` which is indeed 3 cards as expected

Entity/Components/Stats/EntityStats.gd
Turtyo: Let's call modified value `m`, then in the order of operation, the modifier stats are `a`, `b`, `c` and `d`
The operation we are doing is `(m + a + b)*c*d = m*c*d + (a+b)*c*d`
If we do the multiplication first, we get: `m*c*d + a + b` which is different. Doing the multiplication after has a stronger effect. The question is, do we want this ?
If I have a buff that says "draw 2 more cards and double the number of card drawn", for 1 card should I draw:
`(1+2)*2 = 6 cards`
or
`1*2 + 2 = 5 cards` ?
What if we have a card that says the above, and another that says " double the number of card drawn and draw 2 more cards" ? That's a different order of operations. Should we always write the cards in the same order (ie + before x or x before +) ?

Entity/Components/Stats/StatModifiers.gd
cheesefrycook: I see. In that case, there may need to be some changes whenever we implement #35, because the player will get destroyed at the end of combat and therefore the StatComponent will also be destroyed, meaning all the stat objects will be garbage collected. Maybe we can change it to where stats are decoupled from the entity, and there's a global stat holder that keeps stats for the duration of a single run. Maybe in `PlayerManager`

Entity/Components/Stats/EntityStats.gd
cheesefrycook: Gotcha. Would it make more sense to update the naming to `caster_modifier_dict` and `target_modifier_dict`?

Entity/Components/Stats/EntityStats.gd
cheesefrycook: I think either is fine as long as it's consistent. I think we should not have many effects that use multiplication though, because it will make the math confusing for the player. That's more of a design thing though.

Entity/Components/StatusComponent.gd
cheesefrycook: I think it'd be good to combine them into a single array like you mentioned.

I don't think there's a status that does both currently. `turn_start` would be stuff that gets applied each turn, like poison. `on_apply` would be stuff that gets applied once, like a strength buff that lasts for the entire encounter.

I'm not sure I see the reasoning for using a bool. Couldn't we always call `turn_start` or `on_apply` for every status, but the function is left empty if that status doesn't do anything for that specific event?

Entity/Components/Stats/StatModifiers.gd
Turtyo: Ah indeed, I did not realize that the player entity was destroyed at combat end. We would need to hold stats separate from the entity then. The `PlayerManager` indeed seems like the good place to do so. Isn't it the same problem for health though ?

Entity/Components/Stats/EntityStats.gd
Turtyo: Mmh I don't think it would; it might even be confusing. If you cast a strength buff on yourself, do you modify the caster or the target dict ? It would make sense to modify the target dict since you are targeting a buff onto someone, but turns out you need to modify the caster dict.
Same thing if an enemy applies weakness to you, they are targeting the player, but the effect goes on the caster dict; it's more confusing that way i think

Entity/Components/Stats/EntityStats.gd
Turtyo: Yes I also thought it wouldn't be good to have to many effects with multiplications. iirc, StS does times before plus (if you have a weakness effect on an enemy, which reduces damage by 25%, and a relic that reduces one damage for each blow taken, the reduction in percentage by the weakness is applied first)

Entity/Components/StatusComponent.gd
Turtyo: The logic behind adding a boolean for `on_turn_start` was to not assume that every non `on_apply` status was necessarily a `on_turn_start` status.

Entity/Components/Stats/StatModifiers.gd
cheesefrycook: That's a good point. We'll probably have to store our health in `PlayerManager` too

Entity/Components/StatusComponent.gd
cheesefrycook: Right, but is that really necessary? We could just assume that any status can do any of those functions if it wants, and then implement it in that specific status.

The reason I'm worried is because we may have many more conditions for applying status. For example:
- on_turn_start
- on_application
- on_take_damage
- on_card_drawn
- on_heal

With the current implementation, wouldn't we have to have a boolean for each of these different conditions?

Cards/Effects/EffectBase.gd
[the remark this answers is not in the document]
Tysterman74: I believe so. I tried this just now on your branch: 

image

image

The compiler wasn't screaming at me when running the tests, although I'm not sure if this would translate when running the tests.

Entity/Components/Stats/StatModifiers.gd
Turtyo: This should probably be done in a separate issue/PR before #35

Entity/Components/Stats/EntityStats.gd
Turtyo: Currently discussing this with the design team

Entity/Components/StatusComponent.gd
Turtyo: Not necessarily, if they handle in the same way. The reason we need one for `on_apply` and one for `on_turn_start` is that they should be dealt with differently.
`on_turn_start` should have their duration decreased and their effect applied at the start of each turn;
whereas `on_apply` should have their duration decreased at the start of each turn, but they are not applied on each turn, just at the beginning.
`on_take_damage` would probably be an `on_apply` effect
same for `on_card_draw` and `on_heal` actually. Those wouldn't even be inside the status actually. Because let's imagine you want to apply a status when you are healed, this is probably the effect of a card, so it's the role of the card to do it.
If it's more something like getting more heal when you receive a heal, then that's perfectly handled by the stat system already, and same thing for `on_card_drawn` and `on_take_damage`

Cards/Effects/EffectBase.gd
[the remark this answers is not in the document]
Turtyo: alright, will try that then

Cards/Effects/EffectBase.gd
Turtyo: I think the comments of this review can be solved

Cards/Effects/EffectDamage.gd
Tysterman74: I'll go ahead and resolve!

Tysterman74:
> Apart from the things you pointed out in your review @Tysterman74 I have a few other problems with the state of the code as is, and I would like your opinion on this:
> 
>     * The functions name are confusing, and a lot of them do similar things. You have `change_stat`, `change_modifier_of_given_name`, `change_modifier`, `_modify_stats_with_status`, `calculate_value_modified_by_stats`, `calculate_modified_value`; Even I lose myself because the functions are in different files and I don't always remember where each function is, because it's not always clear what they do. I think I should change the names and also moves them to other files (for example there is a function to calculate stat modifications in `EffectBase`, that will probably move)
> 
>     * Generally speaking, I find the structure that I built a bit confusing, not sure if that's only due to the functions, but it's not really easy to understand what everything does

Okay looking into it a bit more, in my opinion i think it feels a bit confusing with just how every function is working across each other. In terms of bullet point 2, if we clean up the functions of bullet point 1 it may make a bit more sense. I'm kind of spitballing and looking at the functions you listed in order and my thoughts on them!

For `change_stat()` in `EntityStats.gd`, a code smell arises from it in terms of the parameters that are being passed in. We pass a `StatDictBase` in, but we have an `offense_modifier_dict` and `defense_modifier_dict` as class variables. Should we not be updating those class variables instead?

For `change_modifier_of_given_name()`, this whole class of StatDictBase seems pretty basic. We're managing a dictionary of the possible modifiers the user may have. I feel this can be cleaned up by just moving this Dictionary of modifiers into the EntityStats. The way I see it, EntityStats.gd is the class that's holding all of the information of a stats of an Entity. Moreover, this StatDictBase is being used in EntityStats for  `offense_modifier_dict` and `defense_modifier_dict`. Would those variables make sense if we're modifying stuff for damage, poison, draw, and card_reward_number?

For `change_modifier()` there's a code smell here that we're modifying too many values at once. In reality of our gameloop, we'd be modifying one value per update, so in most cases some of these values may not be updated.

For `_modify_stats_with_status()`, this function seems to be fine. I think just renaming as `_modify_stats()`, one of the definitions of Clean Code by Agile Bob is to let the lines of code tell the users what it's doing, which the body of the function does just fine. The only thing I would mention is `var target_stats: EntityStats = status_target.get_stat_component().get_stats()` and the issue of chaining calls together to get these stats. I think a better refactor would instead put the `change_stat()` or whatever we refactor it to within the Entity itself. 

For `calculate_value_modified_by_stats()`, `calculate_modified_value()`, to me the naming seems confusing because you're not actually doing the calculations in either of those files until `_calculate_modified_value_offense()` and `_calculate_modified_value_defense()`. I think just renaming those functions to be in-line with what the body of the function is doing should make it a bit clearer.

Entity/Components/StatusComponent.gd
cheesefrycook: I'm not sure how those would fit into those categories. Maybe we are thinking about things differently. Here's some examples of what I'm thinking those events would do:
- "When healed, draw 2 cards" - calls `CardManager.card_container.draw_cards(2)` in the `on_heal` function of the status
- "When a card is drawn, deal 1 damage to all enemies" - calls `deal_damage` on all enemies from the `on_card_drawn` function of the status
- "Take damage at the start of your turn" (aka poison) - calls `deal_damage` on the owner of the status in `on_turn_start`
- "When taking damage, deal 1 damage back to the attacker" - calls `deal_damage` on the attacker in `on_take_damage` of the status
- "Gain 2 strength" - calls `_modify_stats_with_status` in `on_apply`

Unless we want each status to handle durations differently, then every status could just decrease its duration in `on_turn_start`.

Was that what you had in mind? Or were you thinking about things differently?

Cards/Effects/EffectDamage.gd
Turtyo: I'm just keeping this open until i make the issue

Turtyo:
For `change_stat()` in `EntityStats.gd`, we are giving it a `StatDictBase` because both `offense_modifier_dict` and `defense_modifier_dict` are `StatDictBase`. We are indeed modifying either of those class variable with the call to this function, but I don't remember a syntax to specify that the parameter of the function is a variable of a given class (I can't just say we give it an `EntityStats`, because the variables are not `EntityStats` themselves).

For `StatDictBase`, I would even move `enum POSSIBLE_MODIFIER_NAMES` to a global file. If I understand correctly the question about the variables making sense, I think you should read this conversation I had with Cheesy https://github.com/Saplings-Projects/1M_sub/pull/43#discussion_r1451580723
There should probably be a comment or a part in the doc that adds more information about which dictionary to put stats into depending on what you want to modify.

For `change_modifier()` (what do you mean by code smell btw ? I thought the first time was a typo but you've used it twice already), indeed we modify all the values. The reason for this is that I don't want to make a function for each possible modification. The default modification is a modification that does nothing (all the adds are 0, all the multiply are 1). Different stats may modify different add or multiply, so they will all call the same function that does it for them.
Breaking code down into functions is good, but the more code there is, the more there is to debug, so i don't think breaking down those 4 lines into 4 functions would be good.

For `_modify_stats_with_status()`, I think renaming to `_modify_stats()` may lead to confusions with `change_stat()`. The thing with putting it in the entity itself is that as per this discussion https://github.com/Saplings-Projects/1M_sub/pull/43#discussion_r1451576606 , is that the Entity is destroyed at the end of the combat (thus loosing all the stats and HP components). We will probably move all of that to the Player manager, so the calls will change in the future.

I agree for `calculate_value_modified_by_stats()` and `calculate_modified_value()`, since I'm not exactly doing any calculation. Maybe something like `return_modified_value` would be clearer ?

Tysterman74:
> For `change_stat()` in `EntityStats.gd`, we are giving it a `StatDictBase` because both `offense_modifier_dict` and `defense_modifier_dict` are `StatDictBase`. We are indeed modifying either of those class variable with the call to this function, but I don't remember a syntax to specify that the parameter of the function is a variable of a given class (I can't just say we give it an `EntityStats`, because the variables are not `EntityStats` themselves).

Yeah it's the fact we're passing in the parameter of a StatDictBase that confuses me. Why couldn't we just modify `offense_modifier_dict` or `defense_modifier_dict` directly instead? Something like `change_stat(modifier_dict_type: ModifierDictType, modifier: PossibleModifiers, new_modification: StatModifiers)`, `ModifierDictType` being an enum of either `OFFENSE` or `DEFENSE`. That way we can change up the function to be: 
```
func change_stat(   modifier_dict_type: StatDictBase, 
					enum_name: int,
					new_modification: StatModifiers
					) -> void:
    if (modifier_dict_type == OFFENSE):
        // Modify offense_modifier_dict
    else
        // Modify defense_modifier_dict
```

> For `StatDictBase`, I would even move `enum POSSIBLE_MODIFIER_NAMES` to a global file. If I understand correctly the question about the variables making sense, I think you should read this conversation I had with Cheesy [#43 (comment)](https://github.com/Saplings-Projects/1M_sub/pull/43#discussion_r1451580723)
> There should probably be a comment or a part in the doc that adds more information about which dictionary to put stats into depending on what you want to modify.

I saw the change to moving it to the global file and read up on that conversation for that comment. We're good on this. 👍🏽

> For `change_modifier()` (what do you mean by code smell btw ? I thought the first time was a typo but you've used it twice already), indeed we modify all the values. 

Code Smell is for things that when looking at it may lead to further problems if not properly managed and fixed to have a scalable structure. It's subjective from person to person, and just builds from intuition! A good example is [the keys for the modifiers](https://github.com/Saplings-Projects/1M_sub/pull/43/files#r1451118632). We query with using a string literal, but if we need to update this key name for any reason we would need to update several files, instead of moving it to a global file and just updating it in one place.

>The reason for this is that I don't want to make a function for each possible modification. The default modification is a modification that does nothing (all the adds are 0, all the multiply are 1). Different stats may modify different add or multiply, so they will all call the same function that does it for them.
> Breaking code down into functions is good, but the more code there is, the more there is to debug, so i don't think breaking down those 4 lines into 4 functions would be good.

Looking into it a bit more, we do pass all StatModifiers all at once so this is fine 👍🏽 


> For `_modify_stats_with_status()`, I think renaming to `_modify_stats()` may lead to confusions with `change_stat()`. The thing with putting it in the entity itself is that as per this discussion [#43 (comment)](https://github.com/Saplings-Projects/1M_sub/pull/43#discussion_r1451576606) , is that the Entity is destroyed at the end of the combat (thus loosing all the stats and HP components). We will probably move all of that to the Player manager, so the calls will change in the future.

Understandable! I was just giving an example of what to name it as. Reading that thread cheese has a good point so if the call moves or changes that's fine. At the end of the day, we want to just make sure the name of the function matches the block of what it's doing. We are good on this comment. 👍🏽

> I agree for `calculate_value_modified_by_stats()` and `calculate_modified_value()`, since I'm not exactly doing any calculation. Maybe something like `return_modified_value` would be clearer ?

Personally I like `get_modified_value()` more, `return` feels weird to use since that's a keyword used within the language. I would actually even cut `calculate_modified_value()` and just call `_calculate_modified_value_offense()` and `_calculate_modified_value_defense()` directly. Makes it a bit more clearer with the `if caster != null` and `if target != null` blocks.

Turtyo:
So you want to change `change_stat()` from:
```GDScript
func change_stat(   modifier_dict: StatDictBase, 
					enum_name: int,
					new_modification: StatModifiers
					) -> void:
	modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
	modification_count += 1
```
to:
```GDScript
func change_stat(   modifier_dict_type: ModifierDictType, 
					enum_name: int,
					new_modification: StatModifiers
					) -> void:
	if (modifier_dict_type == OFFENSE):
		offense_modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
	else:
		defense_modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
	modification_count += 1
```
that seems like more complicated :thinking: and that doesn't really simplify the calls either:
If you have `stats = entity.get_stat_component().get_stats()`, instead of calling `stats.change_stat(stats.whatever_dict, ...)` you would call `stats.change(ModifierDictType.whatever_type, ...)`

I agree that using string literals is not great at all, i plan to change that anyway (in the global file maybe ?)

Yes for `get_modified_value()`, I agree using `return` sounds strange. For `calculate_modified_value()`, I meant it as a way to call call the two others instead (since those are meant as internal functions). I could rename it to make it clearer, like `select_modified_value_calculation`. I see what you mean with the `if caster != null` and `if target != null` in `EffectBase. I plan to move that function to `EntityStats` because it feels more related to stats than effects directly. So I might just directly call the internal functions from this one, and the need for the selection function will thus be no more.

Tysterman74:
> So you want to change `change_stat()` from:
> 
> ```gdscript
> func change_stat(   modifier_dict: StatDictBase, 
> 				enum_name: int,
> 				new_modification: StatModifiers
> 				) -> void:
> 	modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
> 	modification_count += 1
> ```
> 
> to:
> 
> ```gdscript
> func change_stat(   modifier_dict_type: ModifierDictType, 
> 				enum_name: int,
> 				new_modification: StatModifiers
> 				) -> void:
> 	if (modifier_dict_type == OFFENSE):
> 		offense_modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
> 	else:
> 		defense_modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
> 	modification_count += 1
> ```
> 
> that seems like it's more complicated 🤔 and that doesn't really simplify the calls either: If you have `stats = entity.get_stat_component().get_stats()`, instead of calling `stats.change_stat(stats.whatever_dict, ...)` you would call `stats.change(ModifierDictType.whatever_type, ...)`

It may be more complicated to add another enum in, but at the end of the day the main purpose of my comment was to move away from having to pass in a Dictionary as one of the parameters. We already have `offense_modifier_dict` and `defense_modifier_dict` in this class, we should be updating those values instead. In my opinion, it feels confusing if we had `modifier_dict: StatDictBase` because it would cause us to track down exactly which StatDictBase we are updating. However, the addition of another enum was a suggestion on my end! If you find another solution then all the more power to you! Maybe split the functions up to: 

```
func change_offense_stat(enum_name: int, new_modification: StatModifiers):
    offense_modifier_dict.change_modifier_of_given_name(enum_name, new_modification)

func change_defense_state(enum_name: int, new_modification: StatModifiers):
    defense_modifier_dict.change_modifier_of_given_name(enum_name, new_modification)
```

Then refactor it upstream. My main takeaway is just ensuring we're using the `offense_modifier_dict` and `defense_modifier_dict` properly for these functions instead of passing some StatDictBase dictionary we have no knowledge of.

> I agree that using string literals is not great at all, i plan to change that anyway (in the global file maybe ?)

Global file is a good call, ensuring just one place to update it and it propagates to multiple classes ensures scalability 

> Yes for `get_modified_value()`, I agree using `return` sounds strange. For `calculate_modified_value()`, I meant it as a way to call the `_calculate_modified_value_offense()` and `_calculate_modified_value_defense()` instead (since those are meant as internal functions). I could rename it to make it clearer, like `select_modified_value_calculation`. I see what you mean with the `if caster != null` and `if target != null` in `EffectBase`. I plan to move that function to `EntityStats` because it feels more related to stats than effects directly. So I might just directly call the internal functions from this one, and the need for the selection function will thus be no more.

How I see it, the "internal" portion is the actual logic of calculating the modified value. And that logic is already encapsulated within `_calculate_modified_value_offense()` and `_calculate_modified_value_defense()` which is part of the EntityStats.gd. I think it's fine to call these functions directly since it will help with readability.

Turtyo:
I see what you mean about the unknown dictionary being passed, indeed it could lead to errors where we modify the dictionary of another entity...

Entity/Components/StatusComponent.gd
Turtyo: Why would we bind signals to the `on_apply` ? And yes, I think binding in the child Status would make more sense indeed

The `modifier_name` is set in the `_init()` and not the in the `init_status()` (I think name should probably be changed for clarity btw). I think I prefer having the boolean for now, as I don't really understand the solution you are giving (maybe the lack of sleep is not helping)

Cards/CardBase.gd
```diff
@@ -31,16 +31,17 @@ func parse_card_data(card_data: Dictionary) -> void:
...
 
 
 func can_play_card(caster: Entity, target: Entity) -> bool:
 	return caster.get_party_component().can_play_on_entity(application_type, target)
 
 
 func on_card_play(caster: Entity, target: Entity) -> void:
-	_apply_all_effects(target)
+	_apply_all_effects(caster, target)
+	# TODO add caster as parameter to use stats
```
Tysterman74, line 45: Is this TODO still needed? Looks like Caster is being passed in.

Turtyo: not needed anymore, I forgot to remove the TODO

Entity/Components/StatusComponent.gd
cheesefrycook: I guess we could bind the signals in `_init` or `init_status` too, as long as we just bind them once at the start.

I see, now that I look at it `_init` and `init_status` are pretty much the same thing right? Both get called once when the status is first initialized. Maybe we could remove `_init` all together for clarity?

As for my solution I was talking about, here's what I mean. We would change StatusBase to this:
```GDScript
# StatusBase.gd
func on_apply() -> void:
   pass
```
We would also remove the `is_stat_modification` and `is_on_apply` bools. Then in children of StatusBase we override on_apply and do something like:
```GDScript
# Child of StatusBase.gd
func init_status(in_caster: Entity, in_target: Entity) -> void:
   status_caster = in_caster
   status_target = in_target
   targeted_modifier_dict = in_target.get_stat_component().get_stats().offense_modifier_dict
   modifier_name = StatDictBase.POSSIBLE_MODIFIER_NAMES.damage

func on_apply() -> void:
   _modify_stats_with_status()
```

But like I said, the bool version works as well, just up to preference

Turtyo: The difference between `_init` and `init_status` is that:
- `_init` gets called automatically when you make a new instance of the class (and `init_status` doesn't)
- `_init` is being used for things that we know will always be true (such as the effect of a given status)
- `init_status` is being used to set caster and target (which is dependent on the situation). I would actually rename `init_status` to `set_caster_target` or `set_status_runtime_parameters`

Turtyo:
Will just add buff that changes the amount of Poison applied by a card to test another possible modifier, because for now I've just tested the `DAMAGE` modifier (with `Strength`, `Weakness`, `Vulnerability`).
After that, I think I will have addressed the changes asked (except maybe from @cheesefrycook because you seem to have questions left ?) and I will open this for merge

Turtyo:
Will also add some more tests to see if multiplication works fine (since I've only done addition for now)

Status/Debuffs/Debuff_Vulnerability.gd
Turtyo: I think this is related to the above conversation with the `on_apply` and other types of activation for cards if I understood correctly ?

Tysterman74 (approved):

Status/Debuffs/Debuff_Vulnerability.gd
cheesefrycook: Yes it's the same

project.godot
```diff
@@ -21,6 +21,7 @@ PlayerManager="*res://Managers/PlayerManager.gd"
 CardManager="*res://Managers/CardManager.gd"
 PhaseManager="*res://Managers/PhaseManager.gd"
 RandomGenerator="*res://Managers/RandomGenerator.gd"
+GlobalVar="*res://Global/GLOBAL_VAR.gd"
```
cheesefrycook, line 24: I don't think you need to add this to the autoload, we don't do it for the Enums.gd. You will have to add `class_name GLOBAL_VAR` to the `GLOBAL_VAR.gd` though

Turtyo: Yeah I was thinking about the `Enums.gd` the other day and was wondering if we should move keep that separate or not ?
I was thinking of putting the files from the enum into the global. That was it's clear that this for global variables.

cheesefrycook: I kind of prefer the Enums file because you know it's only enums and the file name is shorter. But either is fine. I still think you can remove GLOBAL_VAR from the Autoload though. I'll go ahead and approve though since I know you're trying to finish this up

cheesefrycook (dismissed):

Turtyo:
I wanted to wait for #39 to be done in order to solve the merge conflicts here (since the other PR has been open for a while). But it is not ready yet, so I think we will merge this one and solve in the other PR.

Turtyo:
Update: we are waiting for #39 
**DO NOT MERGE THIS YET**

Turtyo:
Idk what the github action is smoking, but the tests are not passing at all. Not that they are not passing, they don't even run properly in local.

Turtyo:
Alright now the tests are passing in local for real

Entity/Components/Stats/EntityStats.gd
```diff
@@ -83,3 +83,10 @@ static func get_value_modified_by_stats(modifier_name: GlobalVar.POSSIBLE_MODIFI
 		modified_value = target_stats._calculate_modified_value_defense(modifier_name, modified_value) 
 
 	return modified_value
+
+
+# This function is only used to setup the dictionary in tests
+# In play, the tree logic will already properly setup everything
+func ready_entity_stats() -> void:
+	_offense_modifier_dict.ready_stat_dict()
+	_defense_modifier_dict.ready_stat_dict()
```
cheesefrycook, line 92: Took me a minute to wrap my head around the reasoning for having separate ready functions for test cases. If my understanding is correct, those are needed because _init is called before `@export` variables are updated? So when we are initializing things through the scene we have to do a call_deferred to make sure the export variables are initialized. But when we are doing our test cases we don't care about the export variables, so we call some custom ready functions to make sure things are set up properly.

An alternative I thought of was to add a `on_stat_modifier_initialized` signal to `StatModifiers`. Then put that signal in the `ready` of `StatModifiers` like so:
```GDScript
func ready(	_permanent_add: int = permanent_add, 
			_permanent_multiply: float = permanent_multiply, 
			_temporary_add: int = temporary_add, 
			_temporary_multiply: float = temporary_multiply) -> void:
	modifier_base_dict = {   
					_MODIFIER_KEYS.PERMANENT_ADD:        _permanent_add,  # int
					_MODIFIER_KEYS.PERMANENT_MULTIPLY:   _permanent_multiply,  # float
					_MODIFIER_KEYS.TEMPORARY_ADD:        _temporary_add,  # int
					_MODIFIER_KEYS.TEMPORARY_MULTIPLY:   _temporary_multiply   # float
				}
	on_stat_modifier_initialized.emit()
```
Then only start modifying stats in test cases after we receive the signal. Like so:
```GDScript
# apply strength to player and damage enemy
func test_strength_status():
	var strength_status: Buff_Strength = Buff_Strength.new()
	# add 1 damage offense on the player
	var stat_modifer: StatModifiers = StatModifiers.new()
	stat_modifer.temporary_add = 1

	await stat_modifer.on_stat_modifier_initialized

	strength_status.status_modifier_base_value = stat_modifer
	_player.get_status_component().add_status(strength_status, _player)
```
We will probably need a similar signal when testing Buffs/Debuffs and cards with status, because those also use the stat modifications. Any thoughts?

Turtyo: The reason for the call deferred is because as you guessed, the export variables are not taken into account, because i have to create a new object in the editor to be able to modify the `@export` variables, but by doing so i'm already calling `_init` which leads to the modification of the `@export` variables not being taken into account.
With the `call_deferred`, when playing normally and using the scene tree, the `_init` call is done after the export of variables which allows for editing through the inspector (which is what we want).

I don't think the solution you are giving would directly work:
```GDScript
func test_strength_status():
	var strength_status: Buff_Strength = Buff_Strength.new()
	# the buff gets created, no problem here
	# add 1 damage offense on the player
	var stat_modifer: StatModifiers = StatModifiers.new()
	# the stat_modifier is created, but because of the call_deferred ready is not called (since we are not using the tree in the tests
	stat_modifer.temporary_add = 1
	# you modify the export variable, yes you can do that, no problem here

	await stat_modifer.on_stat_modifier_initialized
	# the ready function was never called (I'm not sure the _init would even if you wait, so you will just be stuck waiting for the signal

	strength_status.status_modifier_base_value = stat_modifer
	_player.get_status_component().add_status(strength_status, _player)
```

I'll quickly put that in my current code to see if it actually works or not

Turtyo: Alright nevermind, that seems to work with a signal. i'm not sure what is the real advantage over calling the `ready` function of `StatModifiers` directly though. Because yes I'm awaiting the signal for this specific case, but when i want to ready the entire offense/defense dictionaries, if I simply `await` for the signal i would continue as soon as I receive one such signal, ie as soon as at least one stat_modifier is ready. And I can't think of a good way to tell it to wait for **all** stat_modifier to be ready. Calling the ready function seems more reliable.

You also say "We will probably need a similar signal when testing Buffs/Debuffs and cards with status, because those also use the stat modifications" but that's already what we do.
We have status like strength, weakness, vulnerability. We even have a status that buffs the time of the poison being applied `Buff_Poison_Duration`

cheesefrycook: I was thinking the signal might be better since it removes the need to have unique code that will only be used for test cases. I worry a bit about that, because its functioning slightly different from the actual game. But the case you mentioned with the StatDictBase is definitely tricky. You'd probably have to add a new signal in StatDictBase that fires when all the StatModifiers are initialized.

Turtyo: If you think about it, the signals also will be used only by the tests in our case; it's also code relevant only to tests because the actual game doesn't await the signals.

If we want to use signals, I think a solution is so add a counter set to 0 at the `StatDictBase` level as well as a signal `on_stat_dict_base_ready`. The counter gets one more each time there is a signal received from `StatModifiers`, when it reaches the size of `Global_var.MODIFIER_KEYS` then it emits
Now the question I have is, what happens with several dictionaries / enemy. The counter is for each dictionary, so there won't be a problem of shared value that will only trigger the signal once, but if we await this signal how do we know that each offense / defense dictionary of each entity is properly setup ?

It would mean I also have to add a signal at the `EntityStats` level with a counter going to 2 (offense and defense) and then emit to say this entity stats are ready. For the entities it's probably fine since we setup sequentially. 
But that's 3 signals and 2 counters to add for each entity basically. For something that would only be useful to tests. I think from a debug standpoint, the functions I currently have are probably easier to debug since it's just recursive calls and for loop.

cheesefrycook: Yeah that's fair. Let's keep your solution then

cheesefrycook (dismissed):

Entity/Components/StatusComponent.gd
```diff
@@ -3,38 +3,64 @@ class_name StatusComponent
 ## Applies status to the entity and holds a list of current status.
 
 
-var current_status: Array[StatusBase]
+var current_status: Array[StatusBase] = []
 
+# func _select_storing_array_for_status(status: StatusBase) -> Array[StatusBase]:
```
Tysterman74, line 8: Commented out function here. Do we still need this?

Tysterman74 (approved):

Cards/Effects/EffectBase.gd
```diff
@@ -8,9 +8,6 @@ func apply_effect(caster: Entity, target: Entity, value: int) -> void:
 # the goal of this function is to provide a base targeting
 # here it's not really useful for what is does, but the goal is to return differently
 # in children class. For example an AoE where you return the entity and the entities on its side
+#TODO change this
```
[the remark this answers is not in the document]
Turtyo: Yes, this is the goal of the targeting PR actually

Entity/Components/StatusComponent.gd
Turtyo: not needed anymore, this was when I made two array for the status
I'm removing it

Tysterman74 (approved):

cheesefrycook (approved):

Issue #38, opened 2023-12-31
Card casting animations


cheesefrycook:
## Feature Description

Each card should have the ability to play a unique animation when cast. This can include anything from animated sprites, particles, screen effects, or sounds.

## Motivation

- Cards have no way to play animations when cast.
- Cards play effects immediately instead of in a sequence. Playing each effect in sequence will be more clear to the player.

## Acceptance Criteria

- Cards should have the ability to play a unique animation for each effect.
- Cards play each associated animation in sequence.

## Proposed Solution

Casting animations can be `PackedScenes`. This is flexible, and allows us to do whatever we want for the animation - like particles, screen effects, animated sprites, or sound.

Each card should have the ability to set its own casting animations. Animations should be separate from card effects, because you may want a different animation for the same effect. For example, consider two cards
- Slap - deal 5 damage
- Headbutt attack - deal 10 damage

Both of these cards have the same underlying effect (a damage effect), but we probably want the casting animations to be different. Because of this, attack animations are assigned in `EffectData`.

Each effect on a card has an associated animation. The animations will play for each effect one by one.

Each animation has an associated delay. This is because we want the effect application to align with a certain point in the animation. For a slap animation, for example, this would be the point when the slap connects with the target.

Set up the cast_animation in the `EffectData.gd`:
```GDscript
# EffectData.gd
@export var cast_animation: PackedScene = null
```
Create a `CastAnimation.gd` that includes a delay:
```GDscript
# CastAnimation.gd
extends Node2D
class_name CastAnimation


signal on_animation_cast_complete

@export var cast_delay: float = 1.0


func play_animation():
   # do whatever animation you want in external nodes or in a child of this script
   
   # Wait for delay to expire. You could override this if you want to do delay in a more sophisticated way.
   await get_tree().create_timer(cast_delay).timeout

   on_animation_cast_complete.emit()
```
Then modify `apply_all_effects` to apply each effect when it receives the delayed signal:
```GDscript
# CardBase.gd
func apply_all_effects():
  var animation_scene = card_effects_data[0].cast_animation
  
  # spawn the animation scene
  var animation: CastAnimation = get_tree().add_child(animation_scene)
  # play animation on CastAnimation.gd node
  animation.play_animation()
  
  # wait for the animation to give the signal for when to play the effect
  await animation.on_animation_cast_complete
  
  card_effects_data[0].apply_effect_data()
  card_effects_data.remove_at(0)

  # apply the next effect in the sequence if there is one
[...]

PR #47, opened 2024-01-19
Issue-31: Move removal of queued card to AFTER we do on_card_play

### LINKED ISSUE #31: Draw card can draw itself
## Describe the bug
If the number of cards you have in your draw and discard pile combined is inferior to the number of card that the draw will take, then the draw card that you used will draw itself.

## To Reproduce
Steps to reproduce the behavior (an example with only 4 cards in hand)
1. Play a card (not the draw card)
2. You now have 1 in discard pile and 0 in draw
3. Play the draw 2 card
4. Since the draw is empty, the discard get put in the draw, and you get the card
5. Then you will get the draw card that you just played
6. Note that this also works if you directly play the draw card

## Expected behavior
You shouldn't be able to draw the card that makes you draw

## Screenshots
In chronological order (will try to take a video next time)

gscreenshot_2023-12-21-033445
gscreenshot_2023-12-21-033452
gscreenshot_2023-12-21-033455
gscreenshot_2023-12-21-033502
gscreenshot_2023-12-21-033505
gscreenshot_2023-12-21-033510

## Additional context
Might have to do with the order in which actions are taken (effect before discard or after).

Tysterman74:
# Description

Updated the card action play flow to separate out the removal from the card hand from adding it to the discard pile in order to help facilitate Draw Card actions potentially drawing itself.

## Related issue(s)

Closes #31 

## List of changes

- Updated `_active_card` variable to CardWorld to help with discarding logic
- Update naming of play_card flow for more clarity
- Update play_card flow to the following:
    - Prepare the queued_card to active, by removing it from the cards_in_hands and setting focused_card to null
    - Set the queued_card to active_card then remove the queued_card
    - Play the active_card
- Update PLAYER_FINISHED logic in `_on_phased_changed()` to call `on_finished_discarding_hand` signal
- Renamed `remove_active_card()` to `finish_active_card_action()` to call `discard_active_card()`
- Added `_discard_active_card()` to execute card animation to discard pile and add it to the discard pile

## Tests

Added a test to `test_cards.gd` to test the card action flow.

## Additional notes

I'm on the fence on this change, cause moving the remove_queued_card() call DOES indeed ensure that discard happens AFTER the card is played, but we're also removing it from the queue. I'm thinking if I want to separate out the removal from queue and the discarding of the card, just to not rock the loop TOO much.

### DIFF
--- a/Cards/CardContainer.gd
+++ b/Cards/CardContainer.gd
@@ -41,7 +41,7 @@ var draw_pile: Array[CardBase] = []
 var discard_pile: Array[CardBase] = []
 var queued_card: CardWorld = null
 
-var _active_card: CardBase = null
+var _active_card: CardWorld = null
 var _focused_card: CardWorld = null
 var _cards_queued_for_add: Array[CardBase] = []
 var _draw_timer: SceneTreeTimer = null
@@ -66,11 +66,10 @@ func set_queued_card(card: CardWorld) -> void:
 	queued_card = card
 
 
-func remove_queued_card() -> void:
+# Remove card from focused and queued and remove it from the hand to prepare it to be played
+func queued_for_active() -> void:
 	_focused_card = null
-	_cards_queued_for_discard.append(queued_card)
 	_remove_queued_card_from_hand()
-	set_queued_card(null)
 
 
 func _remove_queued_card_from_hand() -> void:
@@ -82,7 +81,7 @@ func is_card_queued() -> bool:
 	return queued_card != null
 
 
-func set_active_card(card: CardBase) -> void:
+func set_active_card(card: CardWorld) -> void:
 	_active_card = card
 
 
@@ -92,18 +91,14 @@ func finish_active_card_action(card: CardBase) -> void:
 
 
 func _discard_active_card() -> void:
-	var _card_to_discard = _find_active_card_in_queued_discard_list()
-
-	if _card_to_discard != null:
-		discard_pile.append(_card_to_discard.card_data)
+	if (_active_card != null):
+		# Add active_card to discard queued to trigger the discard animation
+		_cards_queued_for_discard.append(_active_card)
 		_handle_discard_queue()
+		
+		# Add to the discard pile to update the counter
+		discard_pile.append(_active_card.card_data)
 		on_card_counts_updated.emit()
-
-func _find_active_card_in_queued_discard_list() -> CardWorld:
-	for card_index in _cards_queued_for_discard.size():
-		if (_cards_queued_for_discard[card_index].card_data == _active_card):
-			return _cards_queued_for_discard[card_index]
-	return null
 
 
 func are_cards_active() -> bool:
@@ -309,7 +304,10 @@ func _on_phase_changed(new_phase: Enums.Phase, _old_phase: Enums.Phase) -> void:
 	if new_phase == Enums.Phase.PLAYER_ATTACKING:
 		deal_to_starting_hand_size()
 	if new_phase == Enums.Phase.PLAYER_FINISHING:
-		discard_all_cards()
+		if (cards_in_hand.size() == 0):
+			on_finished_discarding_hand.emit()
+		else:
+			discard_all_cards()
 
 
 func _on_card_clicked(card: CardWorld) -> void:
@@ -338,14 +336,11 @@ func _on_card_clicked(card: CardWorld) -> void:
 		card.get_card_movement_component().set_movement_state(Enums.CardMovementState.QUEUED)
 		_focus_card(card)
 
-func play_card(list_target : Array[Entity] ):
-	var queued_card_data: CardBase = queued_card.card_data
-	
-	# remove queued card, then play the card
-	# This is so the queued card doesn't have any influence over our hand count
-	CardManager.card_container.remove_queued_card()
-	CardManager.card_container.set_active_card(queued_card_data)
-	queued_card_data.on_card_play(PlayerManager.player, list_target)
+func play_card(list_target : Array[Entity]):
+	queued_for_active()
+	set_active_card(queued_card)
+	set_queued_card(null)
+	_active_card.card_data.on_card_play(PlayerManager.player, list_target)
 	
 
 func _on_card_hovering(card: CardWorld) -> void:
--- a/Cards/CardContainer.gd
+++ b/Cards/CardContainer.gd
@@ -61,6 +61,7 @@ func set_queued_card(card: CardWorld) -> void:
 
 
 func remove_queued_card() -> void:
+	_focused_card = null
 	_cards_queued_for_discard.append(queued_card)
 	_remove_queued_card_from_hand()
 	set_queued_card(null)
@@ -85,15 +86,18 @@ func finish_active_card_action(card: CardBase) -> void:
 
 
 func _discard_active_card() -> void:
-	var _card_to_discard = null
-	for card_index in _cards_queued_for_discard.size():
-		if (_cards_queued_for_discard[card_index].card_data == _active_card):
-			_card_to_discard = _cards_queued_for_discard[card_index]
+	var _card_to_discard = _find_active_card_in_queued_discard_list()
 
 	if _card_to_discard != null:
 		discard_pile.append(_card_to_discard.card_data)
 		_handle_discard_queue()
 		on_card_counts_updated.emit()
+
+func _find_active_card_in_queued_discard_list() -> CardWorld:
+	for card_index in _cards_queued_for_discard.size():
+		if (_cards_queued_for_discard[card_index].card_data == _active_card):
+			return _cards_queued_for_discard[card_index]
+	return null
 
 
 func are_cards_active() -> bool:
--- a/Cards/CardContainer.gd
+++ b/Cards/CardContainer.gd
@@ -93,8 +93,7 @@ func finish_active_card_action(card: CardBase) -> void:
 func _discard_active_card() -> void:
 	if (_active_card != null):
 		# Add active_card to discard queued to trigger the discard animation
-		_cards_queued_for_discard.append(_active_card)
-		_handle_discard_queue()
+		_add_to_discard_queue(_active_card)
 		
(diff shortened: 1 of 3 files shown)


cheesefrycook:
I think this change is fine, makes sense to me that the queued card is removed after actions happen. In fact, we may even want a delay after the on_card_play once #38 is implemented, because the card casting animations will be playing at that point

cheesefrycook (dismissed):

Core/Battler.gd
```diff
@@ -98,6 +98,6 @@ func _try_player_play_card_on_entity(entity: Entity) -> void:
 		if can_play:
 			# remove queued card, then play the card
```
Turtyo, line 99: need to update this comment

Core/Battler.gd
```diff
@@ -98,6 +98,6 @@ func _try_player_play_card_on_entity(entity: Entity) -> void:
...
 			# This is so the queued card doesn't have any influence over our hand count
```
Turtyo, line 100: How does it affect the hand count then ? If we have 9 cards in hand, play draw 2 cards with a max of 10 cards in hand, how many cards are actually drawn ?
I would expect to draw 2 cards because to me the order should look like this:
- card is played, it's not in my hand
- my hand has 8 cards
- I draw 2 to a max of 10
- I discard the card I played (thus not drawing it)

Turtyo (commented):
I feel like this may lead to some problems with the hand count

Core/Battler.gd
Tysterman74: Ah you may be onto something here, if I play the Draw 2 at 9 cards in my hand, I only get one card.
https://github.com/Saplings-Projects/1M_sub/assets/5923102/0e2555a8-9a5a-461a-b2cf-03a65c98c49c

I'll look into this, but I think I'm going to have to split the queue vs. the discard out of the same function.

For other scenarios it does depend on number of cards in draw/discard.

For this scenario, we have no cards in discard or the draw so hence we do not draw anything.
https://github.com/Saplings-Projects/1M_sub/assets/5923102/d6e2d49e-5895-4e6e-98f7-6a3e986705e0

For this scenario, we play some cards that get discarded and playing the Draw 2 will shuffle those cards back into the draw pile and we draw again.
https://github.com/Saplings-Projects/1M_sub/assets/5923102/5fc6d96f-f142-40cd-ac9c-63190354018e

For this scenario, we have some cards already in the draw pile and play some cards in the discard. Draw 2 will just pull from the Draw pile.
https://github.com/Saplings-Projects/1M_sub/assets/5923102/2bb21aa2-6ced-4fc1-afec-166f14fe9d1a

Tysterman74:
Okay so it turns out we can't just move the line to after we play the card. 

Calling `remove_queued_card()` BEFORE the card is played allows it to be removed from the `card_in_hands` so that way if we have 9 cards and we play the Draw 2 then we are able to draw to 10. However, because of that Draw 2 will be put into the discard before the Draw 2 is played and so there is a potential that it will be drawn.

Calling `remove_queued_card()` AFTER the card is played allows the Draw 2 to not be part of the discard pile, BUT since during the actual drawing of the card it's still considered part of the hand so if we're at 9 cards, we only draw one. 

At this point, do we really want to add in a special case for the Draw 2 to not draw itself? From a gameplay perspective, wouldn't that be a waste of a turn if a player continously plays the Draw 2 every time?

Turtyo:
This may cause some other problems, like a card that has the effect of drawing the top most card of the discard pile would draw itself for example (which might not be the effect we want).
This is why I suggested adding an active card, a state between in the hand and discard

cheesefrycook:
That's a good catch, didn't think of that. I think Turtyo's idea is good, we can have a state between the hand and discard where it's not discarded yet but it isn't considered part of the hand. Maybe we can repurpose the `active_card` with that in mind? So when we call `set_active_card` we can remove that card from the `cards_in_hand`, but don't put it into the discard pile until its effects have been played. That will probably require some changes to the way discarding works

Cards/CardContainer.gd
```diff
@@ -74,10 +79,23 @@ func set_active_card(card: CardBase) -> void:
...
+func finish_active_card_action(card: CardBase) -> void:
+	_discard_active_card()
 	_active_card = null
 
 
+func _discard_active_card() -> void:
+	var _card_to_discard = null
+	for card_index in _cards_queued_for_discard.size():
+		if (_cards_queued_for_discard[card_index].card_data == _active_card):
+			_card_to_discard = _cards_queued_for_discard[card_index]
```
cheesefrycook, line 91: Maybe we could change `_active_card` to be CardWorld instead of CardBase? Might make this part cleaner

Cards/CardContainer.gd
```diff
@@ -61,11 +61,16 @@ func set_queued_card(card: CardWorld) -> void:
 
 
 func remove_queued_card() -> void:
-	_focused_card = null
-	discard_card(queued_card)
+	_cards_queued_for_discard.append(queued_card)
+	_remove_queued_card_from_hand()
 	set_queued_card(null)
```
cheesefrycook, line 66: I think we need to add a `_focused_card = null` here, because I noticed a bug in `_update_card_positions` where the hand does a bit of strange movement due to the `_focused_card` still being set whenever we move the active card to the discard pile

Cards/CardContainer.gd
```diff
@@ -74,10 +79,23 @@ func set_active_card(card: CardBase) -> void:
...
 
-func remove_active_card(card: CardBase) -> void:
+func finish_active_card_action(card: CardBase) -> void:
+	_discard_active_card()
 	_active_card = null
 
 
+func _discard_active_card() -> void:
+	var _card_to_discard = null
+	for card_index in _cards_queued_for_discard.size():
```
Turtyo, line 89: I would change this for loop into a custom find function so you can return the result instead of continuing to loop.

Cards/CardContainer.gd
```diff
@@ -61,11 +61,16 @@ func set_queued_card(card: CardWorld) -> void:
 
 
 func remove_queued_card() -> void:
```
Turtyo, line 63: this name might actually be a bit confusing
`remove_queued_card` I would expect that to be when you decide to use another card, Like I select card A, it gets queued, but then I select card B to play it instead, and I remove A from the queue
Maybe `set_awaiting_play` would be more appropriate ? Because from what I understand the card does
`hand -> queued -> queued for discard -> active -> discard`
I also think `queued for discard -> active` is a bit confusing in itself, why this choice ?

Turtyo (commented):
I think the order might be a bit confusing because active cards are played from the `queued_for_discard` which is not something i would expect.

Cards/CardContainer.gd
cheesefrycook: CardWorld has a reference to its `card_data`

Turtyo: Yes right, I saw that after

Cards/CardContainer.gd
Tysterman74: I'll wait for #39 to get merged in, since there are some changes Jona made to these lines indirectly. I will handle the merge conflicts and address this then!

But you're right, the naming should be better for this phase of the card flow. I'll bring it back to the drawing board.

Turtyo: naming things is always hard anyway

Cards/CardContainer.gd
```diff
@@ -61,11 +61,16 @@ func set_queued_card(card: CardWorld) -> void:
...
-	_focused_card = null
-	discard_card(queued_card)
+	_cards_queued_for_discard.append(queued_card)
+	_remove_queued_card_from_hand()
 	set_queued_card(null)
 
 
+func _remove_queued_card_from_hand() -> void:
+	var _index_to_remove = cards_in_hand.find(queued_card)
```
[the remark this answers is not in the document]
Tysterman74: Did you mean if we changed both `cards_in_hand` and `queued_card` to `CaseBase`?

Cards/CardContainer.gd
Tysterman74: You got a good point about that `queued for discard -> active` I was thinking of making it `ready` instead?

The flow would then be:
`hand -> queued -> ready -> active -> discard`

And in the ready state we remove it from the `cards_in_hand` list

Tysterman74: In terms of the naming, I was thinking of moving from `remove_queued_card` to `set_queued_card_to_ready_card`. Thoughts?

Cards/CardContainer.gd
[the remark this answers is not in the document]
Turtyo: no I meant to ask if the find() function was working as expected for the CardBase class

Cards/CardContainer.gd
Turtyo: or maybe `queued_for_active` ? what was the plan with having a `queued_for_discard` in the first place ? a timer to make sure everything is properly discarded ?

Cards/CardContainer.gd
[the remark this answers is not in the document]
Tysterman74: I believe this should work as long as the card we're searching for and the array are of the same type.

Cards/CardContainer.gd
```diff
@@ -66,24 +66,41 @@ func set_queued_card(card: CardWorld) -> void:
...
+func finish_active_card_action(card: CardBase) -> void:
+	_discard_active_card()
 	_active_card = null
 
 
+func _discard_active_card() -> void:
+	if (_active_card != null):
+		# Add active_card to discard queued to trigger the discard animation
+		_cards_queued_for_discard.append(_active_card)
+		_handle_discard_queue()
```
cheesefrycook, line 97: Could maybe just call `_add_to_discard_queue(_active_card)` instead of these two statements

Tysterman74: OH that's helpful. Thank you for the rec

cheesefrycook (approved):

Turtyo (approved):

Issue #51, opened 2024-01-25
Github action is passing when local tests are not


Turtyo:
## Describe the bug
The Github action that helps us test using GUT is passing, but when pulling and running in local the tests are not running properly. This is not a problem with tests not passing, but tests not running properly and being hit with errors at runtime (for example calling a method on a null value).

## To Reproduce

- Pull the pull request #43 commit https://github.com/Saplings-Projects/1M_sub/pull/43/commits/b5499468d60e19ff9d48cf5008fa927b9957f516
- As we can see, the Github action is validated
- Run the tests in local
- You get "failed to load due to parse error", the test can't even run

Note that it also happens for https://github.com/Saplings-Projects/1M_sub/pull/43/commits/290dec72eb3cd0862e7f0fdf956eafe31baa92f7 but it's not a parse error this time, we get a runtime error ("Invalid get index 'battler_refrence' (on base: 'previously_freed')."

## Expected behavior
The Github action should fail if some tests can't be run

## Screenshots
I found this looking at the logs for https://github.com/Saplings-Projects/1M_sub/pull/43/commits/b5499468d60e19ff9d48cf5008fa927b9957f516 :
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/920ca415-2744-4b4e-bf7a-ea16b79226cd)

The reason is obviously wrong, as both files extend from GutTest, they just aren't parsed properly so the tests are not run at all.

For https://github.com/Saplings-Projects/1M_sub/pull/43/commits/290dec72eb3cd0862e7f0fdf956eafe31baa92f7 we get another log:
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/31627d79-7a80-454f-b74c-974d9e52fad5)
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/bebbf415-0984-43ce-ba02-8d78ede90ce5)

This time, the tests exit when it should have hit an error. Actually the actual reason for the error is that I assign enemy to the root when it already exists inside the battler (so it can't assign).

## Possible fix

A first step should probably be to automatically fail in the following cases:
- any warning
- any orphan node

I am not sure this would cover all possibilities, as I believe we could other kind of runtime errors than orphans problems; further testing is probably needed to see if there are other runtime errors that are possible and that would not met the previous two criteria

Turtyo:
If anyone could pull one of the two mentioned commit and check that the tests indeed don't run properly it would be nice
[...]

PR #52, opened 2024-01-27
Issue-51: Update GUT Tests to assert if there are no orphans

Tysterman74:
# Description

Updating our test suite and action to be able to throw an error and fail the CI when there are orphans and if there is a compiler error.

## Related issue(s)

#51 

## List of changes

- Update CI.yml to use my forked action of the action workflow we're using
    - Added `-d` to my forked action to run the Godot command in debug, which throws an error if there are compiler errors 
- Updated all test suites that are instantiating objects to check for `assert_no_new_orphans("Orphans still exist, please free up test resources.")`

## Tests

Updated `test_cards.gd`, `test_health.gd`, and `test_pile_ui.gd` to check for no new orphans.

## Additional notes

Switched over to use my custom forked action: https://github.com/Tysterman74/godot-tester for running the unit tests. May need a sanity check for someone else to try running the CI with their PR to check it works for other users.


Turtyo:
@Tysterman74 you forgot to actually write the context of the PR

Tysterman74:
Ah yes I'll fill it in once I have the changes all good!

Turtyo (changes requested):
I see you tested it indeed failed because of formatting errors. Could you also test that the "now new orphan" indeed fails ? Doing a push with a test that leaves orphans is probably the best way to do so.
If you could also update `test_event_base` with the same line as other tests file, it would be good

Tysterman74:
> I see you tested it indeed failed because of formatting errors. Could you also test that the "now new orphan" indeed fails ? Doing a push with a test that leaves orphans is probably the best way to do so. If you could also update `test_event_base` with the same line as other tests file, it would be good

This commit here tested out the no new orphans assert! https://github.com/Saplings-Projects/1M_sub/actions/runs/7676114896/job/20923230430

I can work on updating the test_event_base.

Turtyo:
Ah indeed I missed this commit!
Yes please, for `test_event_base`. Also since we merged #43 in `test_stats` also

Turtyo:
If it's still passing even if you leave an orphan, there is something not working properly
just to be sure, which `queue_free` did you comment ? I know there are some redundant `queue_free` on enemies, because we do it on enemies and on battler, but the battler has a reference to the enemies so freeing the battler also frees the enemies

Tysterman74:
Here's the commit that I tested the orphan assert: https://github.com/Saplings-Projects/1M_sub/pull/52/commits/fedc73f9348121555d658184083f7152501b331a CI marks it as a pass.

Turtyo:
Indeed, quite strange
The idea behind detecting orphans was to see if there was an error in execution (since I noticed that every time it fails, there are orphans left). It was not to detect "actual" orphans left because forgotten in the `after_each`
Still strange this doesn't leave orphans, but that was not the idea anyway. I think it might be fine to merge as is, if we detect further problems (unless you have an idea to solve the problem now ?), and we monitor this to be sure for a bit

Tysterman74:
I think it's fine to merge. Maybe when running the stats test it manages to clean up after itself before the after_each() is my theory.

Turtyo (approved):

cheesefrycook (approved):

Issue #46, opened 2024-01-18
Map display


Turtyo:
## Feature Description

A visual representation of the map.

## Motivation

We have the data structure of the map thanks to #30 but we don't have a display for it. The goal here would be to have a simple visual representation of the map. It doesn't need to be fancy, just some squares and letters for rooms events will do as placeholders, but it is to have the base structure of the map display laid out before making it look good.

## Concept art / Designs

![map_concept](https://github.com/Saplings-Projects/1M_sub/assets/49365680/1bc3ff99-e49c-412e-9862-1abd3a70717b)

## Acceptance Criteria

- The map is faithful to the data structure
- No hard coded display, it needs to adapt depending on map that is given
- Player can move on the map by clicking (though this might be done later since we don't really have anything to tell where the player is on the map right now)

## Proposed Solution

No specific solution, make use of Godot's tool !

## Additional Context

This might need the PR that will come from #9 if we want to implement a you see/ you don't see the rooms on the display.

Tysterman74:
Did we ever merge `map-implementation` back down to `main`? I think we may want to make that magic happen first before we start any progress on this Issue.

Tysterman74:
In terms of the UI, were we thinking something similar to StS where we have the icon in the top right 
![image](https://github.com/Saplings-Projects/1M_sub/assets/5923102/93439aea-775d-4758-abe1-185ee06e910f)

Where when we click it, it will open up the map full screen?
![image](https://github.com/Saplings-Projects/1M_sub/assets/5923102/874126e4-ed7b-4125-b2b0-528ffade382c)

Turtyo:
Seems good to me.
And no we haven't merged the `map-implementation` into the `main` yet, for the simple reason that the main doesn't need the map for now. If we merge now and change things later, that's more merge conflicts to solve potentially.

Tysterman74:
From our discussions we had @Turtyo are we okay with leaving the light mechanics out of this issue for now and just focus on getting a display for the map? Then we can save the overlaying of the light in #9?

Turtyo:
Yes, for now we just focus on showing a basic map. The light mechanic can wait for #9 to be implemented. The display may be done at the same time as #9 or separately, but for now we should focus on having a basis before adding more stuff to it.

PR #58, opened 2024-02-01
Issue-46: Map Display

Tysterman74:
# Description

Button for Map Icon on main screen and display of map and rooms that has been generated in #30 

## Related issue(s)

Related to Issue #46 

## List of changes

- Creation of MapUI Scene. This will be used for displaying the Map. Has a return button to go back to the main screen
    - Added logic to generate the rooms dynamically and scroll the map if need be. 
- Addition of Map Icon to TestingScene to open up the Map UI
- Update Map Manager to hold the array of floors, and to create the 2D map when Map Manager is instantiated
    - Add Check if map has been instantiated
- Update RoomBase to return the abbreviation of the type of room it is (used just for testing)
- Added MapManager to autoloader in order to access the map_array within MapUI scene

## Tests

Kind of stumped as to what to test here, some recommendations are welcome!

## Additional notes

Vids of the map in different sizes:

https://github.com/Saplings-Projects/1M_sub/assets/5923102/1e1ee418-9107-4fea-991e-12a8e79d2efd

https://github.com/Saplings-Projects/1M_sub/assets/5923102/a29c321f-8cde-41a8-9241-61f7b0521d92


Turtyo:
This should close the map display no ? Since the light display is for another PR iirc

Map/RoomUI.gd
```diff
@@ -0,0 +1,10 @@
+extends Control
+
+var room: RoomBase
+var label: Label
```
cheesefrycook, line 4: Alternatively can do `@export var label: Label` and then assign in the inspector. Then you could remove the `get_child(0).get_child(0)` call in `_ready`. This may be a bit better because we won't have to change the code if the hierarchy changes

Turtyo:
Things I have to say from pulling the PR alone (i'll review the code just after):
- is it possible to make it so that the distance being scrolled is the exact distance between two floors on the display ? At the start, the lower border of the map fits perfectly between floor, but if the map is long it starts to go over rooms. It's just to try to make it look better
Lower border is between the rooms
image
Lower border is not between the rooms
image

- Would it be possible to make the map larger ? The map can't handle more than 11 rooms of width
image

- Center the map inside the frame, the map is a bit skewed towards the right side, especially visible if you have a lot of rooms on a floor
image

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+func _on_return_button_press():
+	queue_free()
+
+
+func _ready():
+	var current_map = MapManager.current_map
+	
+	# Create New Room Object to append to the room container
+	var new_room = room_ui.instantiate()
+	var new_room_size = new_room.get_child(0).get_size()
```
cheesefrycook, line 21: Could do a function in `RoomUI` script called `get_room_size()` that does this. Might make this section a bit clearer because it would remove the `get_child(0)` call

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+	
+	# Get the objets of the Map UI Scene
+	var color_rect = get_child(0)
+	var scroll_container = get_child(1)
+	var room_container = scroll_container.get_child(0)
```
cheesefrycook, line 26: Could use `Helpers.get_child_node_of_type()` instead. Might be a bit more clear

UI/MapUI.gd
cheesefrycook: Or alternatively use `Helpers.get_child_node_of_type()` like I mentioned below

Map/MapManager.gd
```diff
@@ -26,3 +29,9 @@ static func create_map(map_floors_width: Array[int]) -> MapBase: ## Generates an
 		_grid[index_height].append_array(_padding)
 	_map.rooms = _grid
 	return _map as MapBase
+	
+func _ready():
+	current_map = create_map(map_width_array)
```
Turtyo, line 34: `¢reate_map` is static, should be called through `MapManager.create_map`

Map/RoomBase.gd
```diff
@@ -10,3 +10,15 @@ class_name RoomBase
 
 func _to_string() -> String:
 	return "RoomBase"
+
+func get_room_abbreviation() -> String:
```
Turtyo, line 14: this might be handled better by a dictionary I think

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
+extends Control
+
+var map_scene: PackedScene = preload("res://#Scenes/CardScrollUI.tscn")
+var room_ui: PackedScene = load("res://Map/RoomUI.tscn")
+var _padding_offset = 20
+
+func _input(_inputevent: InputEvent) -> void:
+	if Input.is_key_pressed(KEY_ESCAPE):
```
Turtyo, line 8: can you use the `InputMap` as it was done for other key related events please ?

cheesefrycook:
Overall looks good! I think once we have the player's position represented on the map then we will want the scroll container to center on the player. Also, the scrolling feels a bit jarring at the moment. Turtyo mentioned making the tiers of rooms not be halfway cut off which might help. I'm also wondering if we can make the scrolling smooth instead instantly snapping.

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+
+func _ready():
+	var current_map = MapManager.current_map
+	
+	# Create New Room Object to append to the room container
+	var new_room = room_ui.instantiate()
+	var new_room_size = new_room.get_child(0).get_size()
+	
+	# Get the objets of the Map UI Scene
+	var color_rect = get_child(0)
```
Turtyo, line 24: lots of `get_child` with magic numbers, might be good to explain

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+	if Input.is_key_pressed(KEY_ESCAPE):
+		queue_free()
+
+
+func _on_return_button_press():
+	queue_free()
+
+
+func _ready():
+	var current_map = MapManager.current_map
```
Turtyo, line 17: missing typing

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+	
+	# Create New Room Object to append to the room container
+	var new_room = room_ui.instantiate()
```
Turtyo, line 20: missing typing

UI/MapUI.gd
Turtyo, line 21: missing typing

UI/MapUI.gd
Turtyo, line 24: missing typing

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+	var new_room_size = new_room.get_child(0).get_size()
+	
+	# Get the objets of the Map UI Scene
+	var color_rect = get_child(0)
+	var scroll_container = get_child(1)
```
Turtyo, line 25: missing typing
actually I'm gonna stop there, basically no var has typing

UI/MapUI.gd
```diff
@@ -0,0 +1,64 @@
...
+	
+	# Starting positions for the next room to be generated
+	# X = the Halfway position of room container - half the size of the longest floor to account for centering the all of the rooms. 
+	#	  Add a little padding to push it away from the right edge.
+	# Y = Height of the Room Container that was calculated in room_container_height - (Screen Height - scroll container's bottom y position)
+	var start_position_for_next_room_x = room_container_halfway_position_x - longest_floor_size_width / 2 + _padding_offset
+	var start_position_for_next_room_y = room_container.get_custom_minimum_size().y - (color_rect.get_size().y - scroll_container_bottom_y_position)
+	var position_for_next_room = Vector2(start_position_for_next_room_x, start_position_for_next_room_y)
+	
+	for width_array in current_map.rooms:
```
Turtyo, line 50: name is confusing, i'm expecting this to be an array of width but it's an array of rooms organized by floor

Turtyo (changes requested):
I don't know how it was added back, but I also notice that `Card_DamageHealth.gd` has been added back, doesn't seem to come from here, probably a merge in a previous PR but I didn't see where for now, didn't really search
Would be good to delete it, but it's a bit tricky
The best is probably that i do the push, but i will need to modify the deck for that; if you are ok with that Ty

UI/MapUI.gd
```diff
@@ -13,46 +13,52 @@ func _on_return_button_press():
...
 	# Y = Height of the Room Container that was calculated in room_container_height - (Screen Height - scroll container's bottom y position)
-	var start_position_for_next_room_x = room_container_halfway_position_x - longest_floor_size_width / 2 + _padding_offset
-	var start_position_for_next_room_y = room_container.get_custom_minimum_size().y - (color_rect.get_size().y - scroll_container_bottom_y_position)
-	var position_for_next_room = Vector2(start_position_for_next_room_x, start_position_for_next_room_y)
+	var start_position_for_next_room_x: float = room_container_halfway_position_x - longest_floor_size_width / 2 + _padding_offset
+	var start_position_for_next_room_y: float = room_container.get_custom_minimum_size().y - (color_rect.get_size().y - scroll_container_bottom_y_position)
+	var position_for_next_room: Vector2 = Vector2(start_position_for_next_room_x, start_position_for_next_room_y)
 	
-	for width_array in current_map.rooms:
+	for floor: Array in current_map.rooms:
```
Turtyo, line 56: it's an `Array[RoomBase]` to be precise

#Scenes/MapUI.tscn
```diff
@@ -20,18 +21,21 @@ grow_horizontal = 2
...
-follow_focus = true
+script = ExtResource("2_2gata")
+speed = 2.0
+allow_overdragging = false
 
-[node name="RoomContainer" type="ColorRect" parent="ScrollContainer"]
+[node name="RoomContainer" type="ColorRect" parent="SmoothScrollContainer"]
 clip_contents = true
-custom_minimum_size = Vector2(799, 600)
+custom_minimum_size = Vector2(900, 600)
```
Turtyo, line 36: can the minimum size be deduced depending on the size of the map needed ? based on the number of rooms maybe ? That would allow an adaptable size

Turtyo (commented):
Thank you for the changes
I resolved comments for those that are good, I left open those that are not addressed yet

#Scenes/MapUI.tscn
Tysterman74: Ya i should be able to. I do so with the height, can do it for the width.

UI/MapUI.gd
```diff
@@ -13,46 +13,52 @@ func _on_return_button_press():
...
+	var new_room_size: Vector2 = new_room.get_child(_ROOM_TEXTURE_INDEX).get_size()
 	
 	# Get the objets of the Map UI Scene
-	var color_rect = get_child(0)
-	var scroll_container = get_child(1)
-	var room_container = scroll_container.get_child(0)
+	var color_rect: ColorRect = get_child(_COLOR_RECT_INDEX)
+	var scroll_container: SmoothScrollContainer = get_child(_SCROLL_CONTAINER_INDEX)
+	var vertical_scroll_bar: VScrollBar = scroll_container.get_v_scroll_bar()
+	var room_container: ColorRect = scroll_container.get_child(_ROOM_CONTAINER_INDEX)
```
cheesefrycook, line 32: This is fine, but just to provide one more possible alternative to `get_child()` you could do
```GDScript
@export var _scroll_container: SmoothScrollContainer = null
@export var _room_container: ColorRect = null
@export var _color_rect: ColorRect = null
```
at the top of this script. Then the Godot inspector will let you select children of the hierarchy.

Tysterman74: I didn't get to refactoring out those `get_child()` shenanigans, but I think I'll go down the @export route, just easier to read 😂

Map/RoomBase.gd
Tysterman74: I ended up moving this logic into the Children classes of EventBase.

Turtyo: yeah that works nicely too

UI/MapUI.gd
```diff
@@ -5,7 +5,7 @@ var room_ui: PackedScene = load("res://Map/RoomUI.tscn")
 var _padding_offset = 20
 
 func _input(_inputevent: InputEvent) -> void:
-	if Input.is_key_pressed(KEY_ESCAPE):
+	if (_inputevent is InputEventKey and _inputevent.pressed and _inputevent.keycode == KEY_ESCAPE):
```
Turtyo, line 12: can you use the [InputMap](https://docs.godotengine.org/en/stable/classes/class_inputmap.html) instead of having a `KEY_ESCAPE` ?

UI/MapUI.gd
```diff
@@ -14,22 +18,14 @@ func _on_return_button_press():
...
 	# Create New Room Object to append to the room container
 	var new_room: Control = room_ui.instantiate()
-	var new_room_size: Vector2 = new_room.get_child(_ROOM_TEXTURE_INDEX).get_size()
-	
-	# Get the objets of the Map UI Scene
-	var color_rect: ColorRect = get_child(_COLOR_RECT_INDEX)
-	var scroll_container: SmoothScrollContainer = get_child(_SCROLL_CONTAINER_INDEX)
-	var vertical_scroll_bar: VScrollBar = scroll_container.get_v_scroll_bar()
-	var room_container: ColorRect = scroll_container.get_child(_ROOM_CONTAINER_INDEX)
+	var new_room_texture_rect = Helpers.get_first_child_node_of_type(new_room, TextureRect)
```
Turtyo, line 25: missing typing

Turtyo (changes requested):
I think the code is clearer now, I'll pull later to see how it looks now
This is making me think, should we have a minimum map size to make it look consistent ? If we have the  final boss with only 1 room, it could look strange on the map. You calculate the width and choose the max between the min width and the calculated width

Tysterman74:
image

Here's how a [1, 1] Map looks like. 
If we want to keep a minimum map size, I can add in logic to keep a minimum of array [1, 3, 5, 7, 5, 3, 1] and whatever that size is. That may be prove interesting to deal with since the map size is based on the 2D array we get back. The position may need to be updated.

Turtyo:
Yeah I think a minimum size would be good, as we can see small maps look a bit goofy

Tysterman74:
Here's a few examples of the minimum size map display: 

image

image

image

Let me know if there are any changes I can make.

UI/MapUI.gd
```diff
@@ -25,19 +28,31 @@ func _ready():
...
 	var new_room_size: Vector2 = new_room_texture_rect.get_size()
 	
-	# Get the width of the floor that has the most rooms, by getting the size of what a room is w/ some offset, then 
+	# Get the width of the floor that has the most rooms, by getting the size of what a room is w/ some offset
 	var room_container_width: float = ((new_room_size.x + _padding_offset) * MapManager.map_width_array.max()) + _padding_offset
 	
+	# Check if the room_container_width doesn't meet a certain threshold. If it doesn't set it to a new width, 
+	# but save whatever the current width was for positioning later
+	var new_room_container_width: float = room_container_width
+	if (room_container_width < _MINIMUM_ROOM_WIDTH):
```
Turtyo, line 37: why not just `room_container_width = min(room_container_width, _MINIMUM_ROOM_WIDTH)` instead of creating a new variable ?

UI/MapUI.gd
```diff
@@ -25,19 +28,31 @@ func _ready():
...
+	
 	# Calculate the height of the container where the rooms will reside in. This will be dynamic based on the map array that we have.
 	# The array we have in MapManager, each element will increase the height of the map display, 
 	# multiply by the size of a room w/ some offset to dynamically set the size of the container of which we will be scrolling.
 	var room_container_height: float = MapManager.map_width_array.size() * (new_room_size.y + _padding_offset)
 	
+	# Check if the room_container_height doesn't meet a certain threshold. If it doesn't meet the threshold theb set it to a new height,
+	# but save whatever the current height was for positioning later
+	var new_room_container_height: float = room_container_height
+	if (room_container_height < _MINIMUM_ROOM_HEIGHT):
```
Turtyo, line 48: same as previous comment, `room_container_height = min(room_container_height, _MINIMUM_ROOM_height)`

UI/MapUI.gd
```diff
@@ -70,3 +85,14 @@ func _ready():
 		# When we finish the array for a floor we need to go to the next floor up,
 		# so increase the Y position to the new area the room will be displayed on.
 		position_for_next_room.y -= new_room_size.y + _padding_offset
+	
+	# Calculate the position of where we should put the rooms. 
+	# This is to account for in case the size of the rooms is smaller than the container we put it in
```
Turtyo, line 90: isn't the size of the room always smaller than the container we put it in though ?

UI/MapUI.gd
```diff
@@ -70,3 +85,14 @@ func _ready():
...
+	var new_room_position_x = room_container.get_custom_minimum_size().x / 2 - room_container_width / 2
```
Turtyo: i'm not sure to understand what this does, could you give an explanation ?

UI/MapUI.gd
Tysterman74: I kept the room_container_width and room_container_height mainly to help with the calculations on Line 93 and Line 97 so it can save me a calculation haha. But I'll refactor to do the calculation on those lines as well just to help with readability.

UI/MapUI.gd
Tysterman74: Same answer as above.

UI/MapUI.gd
Tysterman74: Not necessarily! The width that we set for the container accounts for the padding as well, so if it exceeds the minimum width we're setting the container to the width that we calculate on Line 32.

For this screenshot, the size of the rooms is less than the size of the container since we have to set that `_MINIMUM_ROOM_WIDTH`
image

For this screenshot, this size of the room exceeds the `MINIMUM_ROOM_WIDTH` hence we just keep the size of the rooms and just apply it to the container:
image

Turtyo: ah I see now !

UI/MapUI.gd
```diff
@@ -86,13 +75,30 @@ func _ready():
...
+# multiplying that by the max number in the map_width_array to get the width of the largest floor then add offset 
+# to account for the other end of the floor
+func _get_combined_room_width(texture_rect: TextureRect) -> float:
+	return ((texture_rect.get_size().x + _padding_offset) * MapManager.map_width_array.max()) + _padding_offset
+
+# Calculate the height of the container where the rooms will reside in. This will be dynamic based on the map array that we have.
+# The array we have in MapManager, each element will increase the height of the map display, 
+# multiply by the size of a room w/ some offset to dynamically set the size of the container of which we will be scrolling.	
+func _get_combined_room_height(texture_rect: TextureRect) -> float:
+	return MapManager.map_width_array.size() * (texture_rect.get_size().y + _padding_offset)
```
Turtyo, line 104: why does width add another padding offset on top of the usual calculated width, but height doesn't ?

UI/MapUI.gd
```diff
@@ -28,31 +28,20 @@ func _ready():
...
-	# The array we have in MapManager, each element will increase the height of the map display, 
-	# multiply by the size of a room w/ some offset to dynamically set the size of the container of which we will be scrolling.
-	var room_container_height: float = MapManager.map_width_array.size() * (new_room_size.y + _padding_offset)
-	
-	# Check if the room_container_height doesn't meet a certain threshold. If it doesn't meet the threshold theb set it to a new height,
-	# but save whatever the current height was for positioning later
-	var new_room_container_height: float = room_container_height
-	if (room_container_height < _MINIMUM_ROOM_HEIGHT):
-		new_room_container_height = _MINIMUM_ROOM_HEIGHT
+	var room_container_height = _get_combined_room_height(new_room_texture_rect)
```
Turtyo, line 36: missing `float` typing

UI/MapUI.gd
```diff
@@ -0,0 +1,104 @@
...
+	
+	# Starting positions for the next room to be generated
+	# X = the Halfway position of room container - half the size of the longest floor to account for centering all of the rooms. 
+	#	  Add a little padding to push it away from the right edge.
+	# Y = Height of the Room Container that was calculated in new_room_container_height - (Screen Height - scroll container's bottom y position)
+	var start_position_for_next_room_x: float = room_container.position.x + _padding_offset
+	var start_position_for_next_room_y: float = room_container.get_custom_minimum_size().y - (color_rect.get_size().y - scroll_container_bottom_y_position)
+	var position_for_next_room: Vector2 = Vector2(start_position_for_next_room_x, start_position_for_next_room_y)
+	
+	for floor: Array[RoomBase] in current_map.rooms:
```
Turtyo, line 62: godot throwing warning because `floor` is the name of a function, might be good to change this

Map/MapManager.gd
```diff
@@ -26,3 +29,9 @@ static func create_map(map_floors_width: Array[int]) -> MapBase: ## Generates an
 		_grid[index_height].append_array(_padding)
 	_map.rooms = _grid
 	return _map as MapBase
+	
+func _ready():
+	current_map = MapManager.create_map(map_width_array)
```
Turtyo, line 34: "The function "create_map()" is a static function but was called from an instance. Instead, it should be directly called from the type: "Node2D.create_map()"."
Well I thought using `MapManager` would work but apparently godot would rather have `Node2D`

Turtyo (commented):
Also adding some more comments based on the warnings I get when running the scene in godot

Map/MapManager.gd
Tysterman74: I think Godot may be a bit confused with calling a static function on itself. Using Node2D throws this error:
image

UI/MapUI.gd
Tysterman74: Oh good catch! I didn't notice the padding at the top not showing.

Map/MapManager.gd
Turtyo: yeah it's strange, it might indeed be confused because we call a static function from the class itself, not sure how to solve this so it doesn't throw a warning...

Turtyo (dismissed):

Turtyo:
I'm wondering, do we have a way to change the scroll speed ? with an export maybe ? i'm finding it a bit slow, but that might not be the case for everyone
I'm also thinking this might be a parameter for players later, so it might be good to have it exposed, so we can change it easily

Map/MapManager.gd
cheesefrycook: MapManager is autoloaded correct? So could we make create_map not static and then do MapManager.create_map?

cheesefrycook:
Looks like scroll speed is an option on the smooth scroll container

Turtyo:
yes I saw it was an option, but i'm wondering where we are supposed to change it

Map/MapManager.gd
Tysterman74: Ya, was there a reason we had this static? It does work fine without static.

Tysterman74:
The SmoothScrollContainer already has it as an export on the inspector itself! 

image

To your query about changing the speed, we can adjust the speed within code with `SmoothScrollContainer.speed`, this speed only adjusts scroll via the mouse scroll. The click and drag will be based on how fast the user flings the mouse when click and scrolling.

Once we get the save system squared away we can have a setting saved on that save file for adjusting player scroll speed. Just a potential thought.

Map/MapManager.gd
Turtyo: can you `MapManager.create_map` if it's not static ? if yes then let's make it not static, but that means changing function calls to `¢reate_map` in relevant places

Turtyo:
Seems good to me :+1:

Map/MapManager.gd
cheesefrycook: Yeah AutoLoad lets you use that syntax

Turtyo: well perfect then, let's call it directly

Turtyo (approved):

cheesefrycook (approved):

Issue #53, opened 2024-01-28
Cards get stuck in hovered state when spam clicking


cheesefrycook:
## Describe the bug
When spam clicking on cards in hand, cards can get stuck in hovered state.

## To Reproduce
1. Spam click on cards in hand
2. Notice that some cards are stuck in the hovered state

## Expected behavior
Cards should only be in the hovered state when hovered. If not hovered, they should be in the hand as normal.

## Screenshots

https://github.com/Saplings-Projects/1M_sub/assets/151570126/0ca7a71c-0887-490b-af4b-b51130e2c779

cheesefrycook:
My suspicion is that the `create_tween` in the `MoveState_Hovered` is not getting canceled when a new card is queued

Turtyo:
I was able to reproduce that bug on my end. If you already have an idea of the problem, could you check that ?

PR #59, opened 2024-02-02
Issue-53: Fix cards getting stuck in hovered state

cheesefrycook:
# Description

This fixes issue #53 by adding a `lock_hover_time` in `ClickHandler`.

## Related issue(s)

#53 

## List of changes

Adds a `lock_hover_time` to the ClickHandler. Setting a non-zero value will set a timer when a card is unhovered, blocking `on_mouse_hovering` signals.

## Additional notes

This is a very strange issue. What I discovered is that Godot will sometimes fire `InputEventMouseMotion` after the `mouse_exit` event. So when a card is unhovered, it entered the `IN_HAND` state, but then the `_on_gui_input_event` of `ClickHandler` would sometimes still trigger the `InputEventMouseMotion` again putting the card back into the `HOVERED` state. The timer locks the `on_mouse_hovering` event from firing for a very short period after the card is unhovered.

### DIFF
--- a/Input/ClickHandler.gd
+++ b/Input/ClickHandler.gd
@@ -15,15 +15,15 @@ signal on_mouse_hovering
 
 # If greater than 0, then on_mouse_hovering will not trigger for the set time after an unhover event
 # NOTE: this is needed because InputEventMouseMotion will sometimes fire after mouse_exited events
-@export var lock_hover_time: float = 0.0
+@export var time_to_lock_hover: float = 0.0
 
 var _is_interactable: bool = true
 var _lock_hover_timer: Timer = null
 var _hover_enabled = true
 
 
-func _ready():
-	if lock_hover_time > 0.0:
+func _ready() -> void:
+	if time_to_lock_hover > 0.0:
 		_create_lock_hover_timer()
 
 
@@ -67,7 +67,7 @@ func _on_mouse_exited() -> void:
 	on_unhover.emit()
 
 
-func _create_lock_hover_timer():
+func _create_lock_hover_timer() -> void:
 	_lock_hover_timer = Timer.new()
 	add_child(_lock_hover_timer)
 	
@@ -75,11 +75,11 @@ func _create_lock_hover_timer():
 	_lock_hover_timer.timeout.connect(func(): _hover_enabled = true)
 
 
-func _set_lock_hover_timer():
+func _set_lock_hover_timer() -> void:
 	if _lock_hover_timer == null:
 		return
 	
 	# Disable hovering, set timer. When timer expires, hovering will re-enable.
 	_hover_enabled = false
 	_lock_hover_timer.stop()
-	_lock_hover_timer.start(lock_hover_time)
+	_lock_hover_timer.start(time_to_lock_hover)
--- a/Input/ClickHandler.gd
+++ b/Input/ClickHandler.gd
@@ -13,7 +13,13 @@ signal on_unhover
 # emits any time the mouse moves in the area
 signal on_mouse_hovering
 
+# If greater than 0, then on_mouse_hovering will not trigger for the set time after an unhover event
+# NOTE: this is needed because InputEventMouseMotion will sometimes fire after mouse_exited events
+@export var lock_hover_time: float = 0.0
+
 var _is_interactable: bool = true
+var _lock_hover_timer: Timer = null
+var _hover_enabled = true
 
 
 func set_interactable(interactable: bool) -> void:
@@ -36,7 +42,8 @@ func _on_gui_input_event(event: InputEvent) -> void:
 		else:
 			on_right_click.emit()
 	if event is InputEventMouseMotion:
-		on_mouse_hovering.emit()
+		if _hover_enabled:
+			on_mouse_hovering.emit()
 
 
 func _on_mouse_entered() -> void:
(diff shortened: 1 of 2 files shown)


Input/ClickHandler.gd
```diff
@@ -49,5 +56,22 @@ func _on_mouse_entered() -> void:
...
+		return
+	
+	if _lock_hover_timer != null:
+		_lock_hover_timer.stop()
+	
+	# Disable hovering, set timer. When timer expires, re-enable hovering
+	_hover_enabled = false
+	_lock_hover_timer = Timer.new()
+	add_child(_lock_hover_timer)
+	_lock_hover_timer.timeout.connect(func(): _hover_enabled = true)
```
Tysterman74, line 76: Want to sanity check the logic, are we okay with calling the connect here since this will be called everytime `_on_mouse_exited()` is triggered? Not sure where else would be a better place to call the connect function.

cheesefrycook: That's a good point, it may be connecting multiple times? Perhaps it's a good idea to call `disconnect` before connecting again

cheesefrycook: Or maybe we create the timer in _ready instead

Tysterman74 (dismissed):

Input/ClickHandler.gd
```diff
@@ -13,7 +13,18 @@ signal on_unhover
 # emits any time the mouse moves in the area
 signal on_mouse_hovering
 
+# If greater than 0, then on_mouse_hovering will not trigger for the set time after an unhover event
+# NOTE: this is needed because InputEventMouseMotion will sometimes fire after mouse_exited events
+@export var lock_hover_time: float = 0.0
+
 var _is_interactable: bool = true
+var _lock_hover_timer: Timer = null
```
Turtyo, line 21: this is almost the same name as the float variable, this might be confusing

Input/ClickHandler.gd
```diff
@@ -49,5 +61,25 @@ func _on_mouse_entered() -> void:
...
 	if not _is_interactable:
 		return
-		
+	
+	_set_lock_hover_timer()
+	
 	on_unhover.emit()
+
+
+func _create_lock_hover_timer():
```
Turtyo, line 70: missing `void` return type

Input/ClickHandler.gd
```diff
@@ -49,5 +61,25 @@ func _on_mouse_entered() -> void:
...
+	_lock_hover_timer = Timer.new()
+	add_child(_lock_hover_timer)
+	
+	# When timer expires, enable hovering
+	_lock_hover_timer.timeout.connect(func(): _hover_enabled = true)
+
+
+func _set_lock_hover_timer():
```
Turtyo, line 78: missing `void` return type

Input/ClickHandler.gd
```diff
@@ -49,5 +61,25 @@ func _on_mouse_entered() -> void:
...
+	if _lock_hover_timer == null:
+		return
+	
+	# Disable hovering, set timer. When timer expires, hovering will re-enable.
+	_hover_enabled = false
+	_lock_hover_timer.stop()
```
Turtyo, line 84: is this used as a way to reset the timer ?

Input/ClickHandler.gd
```diff
@@ -49,5 +61,25 @@ func _on_mouse_entered() -> void:
...
+	
 	on_unhover.emit()
+
+
+func _create_lock_hover_timer():
+	_lock_hover_timer = Timer.new()
+	add_child(_lock_hover_timer)
+	
+	# When timer expires, enable hovering
+	_lock_hover_timer.timeout.connect(func(): _hover_enabled = true)
```
Turtyo, line 75: this lambda function is created withing the scope of `_create_lock_hover_time`
even if you connect it, does it stay after you exit the function ? maybe this should be declared as a function in itself ?

Turtyo (changes requested):
This is kinda treating the effects of a disease instead of its cause it feels like
But if we don't know the reason for Godot's behavior, this might be the best option :+1:

Input/ClickHandler.gd
cheesefrycook: Yes the lambda stays from my testing. If we connect in the function itself it's going to connect each time that card is unhovered.

Input/ClickHandler.gd
cheesefrycook: Yes, if we entered and the timer was already running, then this will reset it

Input/ClickHandler.gd
Turtyo: can we do something to connect the function only once maybe ? using `is_connected`
my problem is that I fail to understand how the function stays even though it's declared within the scope of another function. I feel like it should just disappear once the function has finished running
Maybe the fact that it is connected to something keeps it ?

Turtyo (approved):

Input/ClickHandler.gd
cheesefrycook: I don't really know how lambda's work under the hood tbh, so I can't say I understand how it stays connected either. If you're worried about it, I can just move that lambda out into a separate function, like
```GDScript
_lock_hover_timer.timeout.connect(_on_lock_hover_timeout)

func _on_lock_hover_timeout():
   _hover_enabled = true
```

Turtyo: If it works with the lambda function properly, it could be nice to understand why :thinking:

Tysterman74 (approved):

Input/ClickHandler.gd
Turtyo: The answer if some people were to look at this discussion later
image

PR #60, opened 2024-02-03
Issue-35: Death handling and combat end

Tomzkk:
# Description

Handle enemy deaths and remove them from combat, as well as detect the end of combat and emit a signal
Currently also reloads the TestingScene on end of battle

## Related issue(s)

https://github.com/Saplings-Projects/1M_sub/issues/35

## List of changes

- Update Battler.gd to detect and handle enemy deaths as well as detect and emit signal on end of combat
- Refactored enemy attack loop in Battler.gd to two separate loops
    - First generate enemy actions into a separate list, then execute them in order
    - Created class EnemyAction which holds information on who is trying to do what action on what target
- Added a scene controller which can reload a scene - used to restart combat when signal is detected
- Created a signal in PhaseManager which is used to indicate end of battle (emitted from Battler and connected to SceneController)
- Changed setter in PartyComponent to set the array rather then extend it

## Tests

Currently none

## Additional notes

- I made SceneController just so that we can see combat actually end and restart and play with it. I imagine actual management of scenes etc will probably be working differently. Currently scene is completely reloaded whether player wins or loses for simplicity


Core/Battler.gd
```diff
@@ -70,16 +73,20 @@ func _on_enemy_start_turn() -> void:
 	# apply status
 	for enemy: Entity in _enemy_list:
 		enemy.get_status_component().apply_turn_start_status()
```
Turtyo, line 75: i think `_handle_deaths() should be called after each enemy's turn start status has been applied.
Imagine the following situation:
Two enemies have 1 HP, each has a dot of poison. One enemy heals other enemies when it dies, by 3 HP, let's say enemy A
- If enemy A dies then enemy B, enemy A heals B then B gets his poison dot applies and survives
- If enemy B dies then enemy A, both die
- In the current situation, since we add all the enemies to the list of enemies to remove, both enemies die in each case because they are removed before actions take effect

I don't know if this is preferable, maybe we want to keep it as is so the order of the enemies doesn't influence situations

Core/Battler.gd
```diff
@@ -118,8 +126,32 @@ func get_all_targets(application_type : Enums.ApplicationType) -> Array[Entity]:
...
+	if _enemy_list.is_empty():
+		PhaseManager.on_combat_end.emit(Enums.CombatResult.VICTORY)
+
+
+func _handle_deaths() -> void:
+	_handle_enemy_deaths()
+	_check_and_handle_battle_end()
+
+
 # TODO reset temporary stats at the end of the combat using EntityStats.reset_modifier_dict_temp_to_default()
```
Turtyo, line 157: this TODO should be done also

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
...
+
+	call_deferred("_deferred_goto_scene", path)
+
+
+func _deferred_goto_scene(path):
+	# It is now safe to remove the current scene
+	current_scene.free()
+
+	# Load the new scene.
+	var s = ResourceLoader.load(path)
```
Turtyo, line 30: name is not explicit, missing typing

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
+extends Node
+
+
+var current_scene = null
+
+func _ready():
+	var root = get_tree().root
```
Turtyo, line 7: missing typing

Core/Battler.gd
```diff
@@ -118,8 +126,32 @@ func get_all_targets(application_type : Enums.ApplicationType) -> Array[Entity]:
...
 			
 	return all_target
+
   
-# TODO condition check for killing enemies and removing them from the combat
-# TODO condition check for killing player and ending the combat
-# TODO condition check for killing all enemies and ending the combat
+func _handle_enemy_deaths() -> void:
+	var enemies_to_remove : Array[Entity] = []
+	for enemy in _enemy_list:
```
Turtyo, line 133: missing typing for enemy (same for all `for` loops)

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
...
+	current_scene = root.get_child(root.get_child_count() - 1)
+	PhaseManager.on_combat_end.connect(_combat_end_change_scene)
+
+
+func goto_scene(path):
```
Turtyo, line 12: missing `void` return type

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
...
+	# a bad idea, because it may still be executing code.
+	# This will result in a crash or unexpected behavior.
+
+	# The solution is to defer the load to a later time, when
+	# we can be sure that no code from the current scene is running:
+
+	call_deferred("_deferred_goto_scene", path)
+
+
+func _deferred_goto_scene(path):
```
Turtyo, line 25: missing `void` return type

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
...
+	current_scene = s.instantiate()
+
+	# Add it to the active scene, as child of root.
+	get_tree().root.add_child(current_scene)
+
+	# Optionally, to make it compatible with the SceneTree.change_scene_to_file() API.
+	get_tree().current_scene = current_scene
+
+
+func _combat_end_change_scene(combat_result: Enums.CombatResult):
```
Turtyo, line 42: missing `void` return type

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
...
+	PhaseManager.call_deferred("set_phase", Enums.Phase.GAME_STARTING)
+	if combat_result == Enums.CombatResult.DEFEAT:
+		print('Defeat')
+		goto_scene("res://#Scenes/TestingScene.tscn")
```
Turtyo, line 46: instead of putting the name directly here, I think using a dictionary with `{DEFEAT_SCENE: defeat_scene_path, WIN_SCENE: win_scene_path` will allow us to change the dictionary in the future without having to change this function or other functions that do the same

Global/EnemyAction.gd
```diff
@@ -0,0 +1,29 @@
+extends RefCounted
+
+class_name EnemyAction
+
+var caster: Entity
+var action: CardBase
+var target: Array[Entity]
+	
+func _init(caster: Entity, action: CardBase, target: Array[Entity]):
```
Turtyo, line 9: might be better to use different names from the class names to not be confused

Global/EnemyAction.gd
```diff
@@ -0,0 +1,29 @@
+extends RefCounted
+
+class_name EnemyAction
+
+var caster: Entity
+var action: CardBase
+var target: Array[Entity]
```
Turtyo, line 7: this should not be named target (singular) when it's an array potentially containing multiple targets
`list_targets` might be better

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
+extends Node
+
+
+var current_scene = null
```
Turtyo, line 4: missing typing

Turtyo (changes requested):
The general structure is clear (thank to the earlier explanation in the issue). There are some problems with the code which I commented. Thank you for the PR, I think it will work nicely in the end :+1: 
Tests have to be provided before this can be merged

Core/Battler.gd
```diff
@@ -24,6 +23,7 @@ func _ready() -> void:
 		
 	PhaseManager.on_phase_changed.connect(_on_phase_changed)
 	CardManager.on_card_container_initialized.connect(_on_card_container_initialized)
+	CardManager.on_card_action_finished.connect(_handle_deaths.unbind(1))
```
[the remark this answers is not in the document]
Tomzkk: I am reusing the _"on_card_action_finished"_ signal which is defined in CardManager and used for other purposes. This also comes with an argument for a card that's being played. I just need to know that a card has been played and do not care for which card at this moment, so I opted for unbinding it rather then adding an argument to the __handle_deaths_ function that would never be used. 

I could also make my own signal without an argument and use it for this if you think it would be better to keep the signals separate for some reason. I didn't want to create a signal which would basically perfectly shadow the _on_card_action_finished_ signal we already had as I didn't see a reason to. Let me know if you can think of any concerns regarding that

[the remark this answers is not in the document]
Turtyo: No no it's fine as is, I just wanted to make sure to understand

Core/Battler.gd
Tomzkk: A fair point, although if I wanted to call it after each enemy's turn I would need to do same thing as I did for their attacks as it is within loop through _enemy_list.

Actually, in most games I know, statuses like poison etc are only resolved when it comes to specific enemies' turn, however that would require us to rework the enemy turn to have them first be hit with status effects, then attack and only then hand the turn over to the next enemy. 
Currently, statuses as poison etc are specifically resolved as "on turn start" events, which implies they should be handled right at the start of the turn start
 
If we wanted to rework it into what i mentioned above, I am not sure how much work that would require - maybe it would want to be a separate issue
If we want to keep them as is and be resolved at the start of the turn, then I'm not sure which order would be preferred 🤔

Core/Battler.gd
Tomzkk: Currently, I emit a signal and SceneController completely reloads the scene - this means entities get reloaded from 0 and their stats are gone. As such the todo doesn't really apply as the entity which holds the stats gets recreated per my understanding - which is why I left the todo in place so that it gets implemented once proper handling of combat progression is implemented

Core/SceneController.gd
cheesefrycook: I agree - in the future we could have other scenes in the dictionary too like the Shop and Rest Area

Global/EnemyAction.gd
```diff
@@ -0,0 +1,29 @@
+extends RefCounted
```
cheesefrycook, line 1: How does RefCounted differ from Resource?

Core/SceneController.gd
```diff
@@ -0,0 +1,50 @@
...
+
+	# Add it to the active scene, as child of root.
+	get_tree().root.add_child(current_scene)
+
+	# Optionally, to make it compatible with the SceneTree.change_scene_to_file() API.
+	get_tree().current_scene = current_scene
+
+
+func _combat_end_change_scene(combat_result: Enums.CombatResult):
+	PhaseManager.call_deferred("set_phase", Enums.Phase.GAME_STARTING)
```
cheesefrycook, line 43: So I'm guessing PhaseManager's `_ready` doesn't get called whenever we reload the scene which is why we need to call it manually here. Would it be problematic to move the logic from the `_ready` of PhaseManager into a new function and then call that function manually after the scene is loaded?

Core/Battler.gd
cheesefrycook: Maybe it helps to separate the status events into two separate cases?
`on_turn_start` = when the `ENEMY_ATTACKING` phase starts
`on_enemy_before_action` = before the specific enemy's action. Only applied to the enemy that is casting its action.

Also, #38 will make each enemy attack one after another. So Enemy A attacks, plays an animation. After some delay, Enemy B attacks and plays its animation.

Global/EnemyAction.gd
Tomzkk: https://docs.godotengine.org/en/stable/tutorials/best_practices/node_alternatives.html

Per my understanding the main difference is Resource adds the ability to be serialized/deserialized which allows you to export properties into/from the inspector. If I'm not misunderstanding or missing something then I do not really think we need anything from Resource specifically ?

Core/SceneController.gd
Tomzkk: I don't really see a problem with it in PhaseManager. 
We might not even need to touch phases at the start of this function and switching the phases at the end of it (by calling the function like current PhaseManager's ready) might be enough ?
 
I think the reason I did it like this was I wanted to switch to a Phase that doesn't have any events happening inside of it to prevent things happening while the scene is being cleaned and reloaded - although I'm not sure if that is actually necessary or even fulfills what I intended 🤔

Global/EnemyAction.gd
cheesefrycook: Yeah, I guess if we don't need any export properties then this is fine

Core/SceneController.gd
cheesefrycook: I'm thinking it may be sufficient to switch phases after the load is complete. However if we run into problems and want to prevent stuff from happening during the load process, we could potentially add a new phase like `GAME_END` or something

Core/Battler.gd
Turtyo: The reset stat will reset the temp stats, which need to be reset anyway
Doing this would allow for proper testing of stat reset at combat end (though I already tested them in unit tests)

It also makes sense to leave it as is for when we have multiple combats I guess

Core/SceneController.gd
Turtyo: exactly as Cheesy said

Core/SceneController.gd
Turtyo: `GAME_END` would be confusing, `SCENE_END` probably better
but anyway, when you load a new scene it will call the tree that will await for ready signals from it's children nodes, so I think it should all setup properly when reloading a new scene :thinking:

Tests/test_death_handling.gd
```diff
@@ -0,0 +1,118 @@
...
+		node.queue_free()
+
+func after_each():
+	_free_if_valid(_player)
+	_free_if_valid(_enemy)
+	_free_if_valid(_enemy_2)
+	_battler.queue_free()
+
+func test_check_and_handle_battle_end_player_death():
+	_player_health_component._set_health(0.)
```
Turtyo, line 63: should also have a function that does when player is killed by an enemy attack

Tests/test_death_handling.gd
```diff
@@ -0,0 +1,118 @@
...
+	_enemy_2_health_component._set_health(0.)
+	
+	assert_eq(_enemy_list.size(), 2)
+	_battler._handle_enemy_deaths()
+	assert_eq(_enemy_list.size(), 0)
+
+
+func test_enemy_death_to_player_attack():
+	_enemy_health_component._set_health(1.0)
+	var card_damage: CardBase = load("res://Cards/Resource/Card_Damage.tres")
```
Turtyo, line 104: can you make a test that uses damage all enemies to test that we can kill both enemies through a single attack ?

Tests/test_death_handling.gd
```diff
@@ -0,0 +1,118 @@
...
+	
+	assert_eq(_enemy_list.size(), 2)
+	card_damage.on_card_play(_player, [_enemy])
+	assert_eq(_enemy_list.size(), 1)
+
+
+func test_enemy_death_to_poison():
+	_enemy_health_component._set_health(1.0)
```
Turtyo, line 112: should also make a function where the duration of the poison is set to 1 so poison expires when enemy dies
this would allow to check everything works fine on the order in which things are done

Core/Battler.gd
Tomzkk: I see, I misremembered that STS does it like that then. 
The effects that would be affected by playing them in that order are mainly damage debuffs - so that if an enemy has a damage over time on it that would do enough damage to kill it, other enemies still have a window to do something about it
Though ultimately I think it's okay either way, as long as it's consistent

The only ways I see it doable right now are basically redoing what I did with enemyAction (albeit simpler) and splitting the debuff loop - so I would have to duplicate enemy list and apply statuses while going through it rather then through the original - or by making debuff handling part of EnemyAction's execution (which would however also change when debuffs are handled)

Core/Battler.gd
Tomzkk: Per my understanding temp stats are still bound to the entity (so the player in our case I believe), which gets deleted and recreated since I reload the scene, or am I wrong there ? If that is the case, why is there a need to reset temp stats ?

Global/EnemyAction.gd
```diff
@@ -2,28 +2,28 @@ extends RefCounted
...
-var target: Array[Entity]
+var _caster: Entity
+var _action: CardBase
+var _target_list: Array[Entity]
 	
-func _init(caster: Entity, action: CardBase, target: Array[Entity]):
-	self.caster = caster
-	self.action = action
-	self.target = target
+func _init(caster: Entity, action: CardBase, target_list: Array[Entity]):
```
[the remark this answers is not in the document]
Tomzkk: I believe underscored variables are in most cases used for instance variables, so it might be a bit confusing - we could also rename the function ones to be slightly different but I didn't have an idea at the time so I opted for adding underscores to the class variables

Tests/test_death_handling.gd
Tomzkk: I believe I actually had that but deleted it when I started having issues with scene reloads. Can look into adding it back

Tests/test_death_handling.gd
Tomzkk: Same as previous comment

Tests/test_death_handling.gd
Tomzkk: Added test for this one as well as for the previous two comments

Core/Battler.gd
Turtyo: other enemies don't "still have a window to do something about it" because turn start applications are automatic things, enemies can't do something when the turn already started to mitigate those
i think we should keep it as is, because it will be consistent no matter the order of the enemies on screen and in the list

Core/Battler.gd
Turtyo: Yes it gets deleted for now, but later we will copy the stats, so we will need the reset stat function to be connected to the end of combat signal anyway

As I said, it's fine to leave it as is for now, since it is indeed related to stats carrying over, so not yet implemented

Global/EnemyAction.gd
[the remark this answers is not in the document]
Turtyo: In GDScript convention, variables using an underscore are "private" and are not meant to be accessed from outside the scope of the current script, just called by this script and children of this script
Thus it would be more logical to use `¢aster`, `ŧarget` and `target_list` for the class variables and `_caster`,`_target` and `_target_list` for the `_init` variable
As you can see, `_ready` functions and `_init` functions are a good example of that. They use an underscore because you don't call them directly (`_init` gets called by new() and `_ready` gets called when the tree is building the nodes)

Tests/test_death_handling.gd
```diff
@@ -116,3 +134,16 @@ func test_enemy_death_to_poison():
...
+func test_enemy_death_to_expiring_poison():
+	_enemy_health_component._set_health(1.0)
+	var card_poison: CardBase = load("res://Cards/Resource/Card_Poison.tres")
+	card_poison.on_card_play(_player, [_enemy])
+	_enemy.get_status_component().current_status[0].status_turn_duration = 1
+
+	assert_eq(_enemy_list.size(), 2)
+	assert_eq(_enemy.get_status_component().current_status.size(), 1)
+	_battler._on_enemy_start_turn()
+	assert_eq(_enemy.get_status_component().current_status.size(), 0)
```
Turtyo, line 148: does that even work ? isn't the enemy dead thus the resource is freed and you can't check his status component ?

Core/SceneController.gd
Tomzkk: What do we think the action here should be then ? 🤔 

Should I make the new "COMBAT_END / SCENE END / ..." phase, switch to it at the beginning of this and then call the equivalent of _ready function of phaseManager after the switch like @cheesefrycook suggested ? Or ?

Tests/test_death_handling.gd
Tomzkk: freeing happens deferred via queue_free so it doesn't actually get freed until the next frame - so at the end of the test the enemy is still a valid node and we can check it - at least that's my understanding and the test has been consistently passing but let me know if you think I'm wrong in something

Actually I don't even see a way to check if enemy has actually really been freed in a test (which I wanted to assert for as well) because I would have to await for the next frame in the test - which I tried to do but it then broke things as it seemed to me that the "_before_each_" was executed when the next test was already in progress and it freed the node midrun

Global/EnemyAction.gd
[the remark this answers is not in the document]
Tomzkk: Yes, I'm saying the same thing as you are - it's just that the underscore in name is overall usually reserved for instance variables or methods that are private rather then for function arguments. Even the examples you mentioned are class functions (_init, _ready). Although I'm not sure how closely godot follows it, but from what I've seen it does seem to be the case

Overall it's just a convention though and not a big problem as the init should be pretty obvious anyway, so I swapped them around 👍

Core/SceneController.gd
Turtyo: Well imo I think it should work without that
Meaning you shouldn't have to set the `GAME_STARTING` for the game to be in this specific phase, as I explained it should naturally be set like that when the scene starts

Global/EnemyAction.gd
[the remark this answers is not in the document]
Turtyo: well for now you have class variables with an underscore, (meaning private), and function parameters without an underscore (meaning public)
I would expect it to be other way around

Tests/test_death_handling.gd
Turtyo: `before_each` is executed at the start of each test, so if you await in a specific test, it shouldn't just jump to the next test and launch the `before_each`

`_on_enemy_start_turn` calls `_handle_deaths` which calls `_handle_enemy_deaths`, and in there you add everything to a list to remove, erase and queue free. `queue_free` actually frees at the end of the current frame. If that's the only thing you need to wait you can either:
- `await get_tree().process_frame` to wait for the deferred calls
- `is_queued_for_deletion()` doesn't wait for deferred calls but you can check the enemy is indeed waiting to disappear
Option one is probably better as it makes sure everything works instead of assuming the `queue_free` works directly

Core/SceneController.gd
Tomzkk: PhaseManager is autoloaded, so just reloading the scene doesn't do anything with phases. 
If I remove the phase settings I am doing here, what happens is enemy turn finishes running (as scene reload is deferred) which switches phase to PLAYER_ATTACKING. After the reload, the phase stays on PLAYER_ATTACKING. The issue with this is that the cards are not served to the player (as cards are served when it detects change from some phase into PLAYER_ATTACKING) - game gets into a state where it is player's turn but he has no cards and can only end turn

Tests/test_death_handling.gd
Tomzkk: When I try to do it via option one, which is essentially the same as I did before which I described above, it results in tests throwing error "Nonexistent function in base 'Nil'" at some point during test's execution - which the only way I can see happening is if the player was freed in the middle of the test run and the only way we are freeing it is at the end of the previous test 🤔

Option 2 works fine

Core/Battler.gd
Tomzkk: What I meant was if we resolved those things at the start of enemy action, there could be enemy that would have it's turn before affected enemy which could cleanse it / heal it / whatever.

Core/SceneController.gd
Turtyo: Aaaaah i forgot that PhaseManager is autoloaded, you are right ! So you indeed need to call the `_ready`
I think doing a `SCENE_END` phase is probably the best here then

Tests/test_death_handling.gd
Turtyo: Did you try to put some breakpoints in godot to check when things are being freed ? That might allow you to see why option one doesn't work

Core/Battler.gd
Turtyo: ah yes I see
we will keep it as is anyway I think, because it's more consistent that way

Tests/test_death_handling.gd
Tomzkk: It seems the issue is PhaseManager is being recreated at a wrong time somehow and switches phase to Player turn which tries to apply player statuses at a time when player doesn't exist
image


Also it looks like the issue is happening if I use both await inside of tests AND queue_free in _after_each_. I tried switching the freeing after tests to _free_ and it seems like it's passing without issues

Turtyo: well as long as it's passing consistently it's fine

Turtyo (approved):

Turtyo:
Well I think there is only the comment for the `SCENE_END` phase that needs to be done and it should all be good

cheesefrycook (approved):

Core/SceneController.gd
Tomzkk: I wasn't sure if it would be better to do SCENE_END or call the _ready at the end (rather then just changing phase to player's turn) - in the end I did both. This way the SCENE_END is currently redundant I believe, but makes it more clear what's happening and might have use in the future ? [See here](https://github.com/Saplings-Projects/1M_sub/pull/60/files#diff-c6159dde6e380cb517c924628c3146eff48ca4e4b4d2cdc89af1c5adf9290290R47-R55)

Tests/test_death_handling.gd
Tomzkk: Added asserts into tests - I kept both queued_for_deletion before await and checking if node was freed after

cheesefrycook (approved):

Tests/test_death_handling.gd
```diff
@@ -50,15 +50,15 @@ func before_each():
 		PhaseManager.disconnect("on_combat_end", SceneController._combat_end_change_scene)
 
 	
-func _free_if_valid(node: Node):
+func _free_if_valid(node):
```
Turtyo, line 53: missing typing, you removed it when it was already there; any reason ?

Turtyo (changes requested):

Core/SceneController.gd
Turtyo: yes i read it, it's clear what happens; a good idea I think :+1:

Tests/test_death_handling.gd
Tomzkk: If the node is invalid (as it was freed when the enemy died), it is no longer a node type. 

I don't think godot has option for multi-typing and even if it did, I'm not sure what type would even cover invalid node if any 🤔

Turtyo: mmh I see the problem indeed;
could you put a comment so we don't forget about this exception ?

Tests/test_death_handling.gd
```diff
@@ -0,0 +1,172 @@
+extends GutTest
```
Turtyo, line 1: after pulling the branch, I realized that this test doesn't use the structure that was made in #63 
Normal, since this PR was started first, but could you update this to extend from `TestBase` ? You will be able to remove all variable declarations, most of the `before_each` (except after the `watch_signals`, but you can call `super` for everything before)
You will need to leave the the `after_each`, and could you also add an `assert_no_new_orphans` as in done in other tests ? This is to check for runtime errors mainly (cf #51 and #62 )

Turtyo (changes requested):
Something to change in test that I noticed by pulling the branch. I think we'll be good to merge after this is done :+1:

Tests/test_death_handling.gd
Tomzkk: Added

Tests/test_death_handling.gd
Tomzkk: Done

Turtyo (approved):

cheesefrycook (approved):

PR #63, opened 2024-02-08
Make a common setup function for all tests

### LINKED ISSUE #61: Make a common setup function for all tests

Watching at our tests, most do the same thing (initialize scene tree, manage enemies, get some components for checks...). It would be good to have a way to write this setup once and not have to copy paste it into each test tile.

Make a parent test file extending from `GutTest` with variables that are often used, `before_each` and`after_each` functions copying general template that can be found at the start of test files.
All other tests extend from this base setup file. If they want more specific setup, they can overwrite the base setup, and they can also call `super` if they want the base setup but just want to add a bit to it in the overwrite.

Only problem is that I'm not sure how well Gut handles its tests being child of other "tests". It should work, since by transitivity property, if test extends base test which extends Gut, then test should extend Gut thus it should all be working (but we need to make sure).

## Describe alternatives you've considered
We could have static helper functions from a base setup file. This allows for test to not be extending from a base test, in case the previous solution doesn't work when tests don't directly extend from `GutTest`. The problem with that is that no overwrite function means files that need something specific would need to rewrite the entire thing (unless there is a more elegant solution ? Like having a `custom_before_each` in the helper, and calling that in the `“before_each` of each file, and eventually adding stuff to it ?)

Turtyo:
# Description

Add common test setup files for tests that need similar setup. This removes the need to rewrite the setup each time.

## Related issue(s)
Closes #61 

## List of changes

- Add `TestBase` as parent class for health and stat tests
- Add `TestCardBase` as a parent class for card and pile ui tests

## Tests

Mentioned tests were modified to extend from their respective parent class

## Additional notes

This introduces some variables that might not always be needed by every sub-test (for example health doesn't need the stat component), but I don't think that will be a problem in itself.


Turtyo:
I didn't know github would interpret the title of a commit as an actual @ for someone (especially considering i wrote Override and not OverRide...)
Well, sorry for the person i pinged by mistake

Tests/test_pile_ui.gd
```diff
@@ -1,49 +1,31 @@
...
-		_card_container.default_deck[card_index] = CardBase.new()
-		_card_container.default_deck[card_index].card_title = "Card" + str(card_index + 1)
-	
-	get_tree().root.add_child(_card_container)
-	
 
+# @Override
 func after_each():
-	_card_container.free()
+	_card_container.queue_free()
```
Tysterman74, line 18: Since you inherit from TestCardBase, we should be able to call the super to free up the _card_container instead of calling it here, right?

We can also free the `_card_scroll` before we call the super, and remove that assert_no_new_orphans() to let the parent call that assert.

Tysterman74 (changes requested):

Tests/test_pile_ui.gd
Turtyo: the reason for that is that in `TestCardBase` you have:
```GDScript
func after_each():
	_card_container.queue_free()
	assert_no_new_orphans("Orphans still exist, please free up test resources.")
```
and In `test_pile_ui`:
```GDScript
func after_each():
	_card_container.queue_free()
	_card_scroll.free()
	assert_no_new_orphans("Orphans still exist, please free up test resources.")
```
If I was to call `super` in `test_pile_ui`, the `assert_no_new_orphans` would be called before I have a chance to call the `_card_scroll.free()` which is relevant only to this test. Which is why I need to rewrite the `after_each` instead of calling `super`

Tysterman74: Ah I get ya now. We'd have to recall it again. Sounds good!

Tysterman74 (approved):

cheesefrycook (approved):

PR #66, opened 2024-02-11
Proper targeting system

### LINKED ISSUE #42: Proper targeting system
## Feature Description

A way for a card to simply add targets to its effects.

## Motivation

Currently we can basically do the following targeting: self, target one enemy, all members of a party and everyone. The problem with the current implementation is that it would require to modify the battler / the card base or effect every time we want to introduce a new way of targeting enemies, because there is no structure that delegates the task to a function made for it.

## Acceptance Criteria

- The new targeting still makes use of the insta-cast defined in PR #39 
- The targeting functions can be added as an element of the data of an effect
- In the battler, the targeting is given to a function so that we don't need to modify the battler every time
- The targeting makes use of the new EnemyManager (similar to PlayerManager) that is done in this PR too

## Proposed Solution

We make a class (let's say `TargetingBase`). We then define sub-class which are all different targeting (like solo-target, all-enemies, targeted enemy + the one on its right, random targeting of X enemies, etc...).

The base has a targeting type (to be able to work with the insta-cast), set to targeting by default but can be overridden by children class that can be insta-cast (like targeting self or targeting all enemies). It also has a blank `generate_target_list` function (name can be discussed ofc); which is overridden by children class

`EffectData` now has a new export variable which is the targeting chosen for the effect we want to apply (so with that you can apply the same effect but in different ways, like poison all or poison target both use the poison effect but with different targeting).

## Additional Context

This is a WIP because it needs #39 to be merged before anyway, and it also seems to be a point that could be open to discussion depending on how the team wants to do this.

Turtyo:
# Description

This pull request rewrites the targeting system of the game so effects are now separate from who they target. Meaning an effect can be applied in multiple different ways depending on the chosen targeting.

## Related issue(s)

Closes #42 

## List of changes

- Create `TargetingBase` as the default one target function, all other targeting extend from this
- Rework `on_card_play` to just go through all the effects and targets (which honestly looks way easier to understand now)
- Remove unneeded `TargetType` enum and associated references
- Update all cards

## Tests

The previous test should cover most of the cases. I will write some more tests for the more esoteric targeting functions that don't have use case for now (like a random targeting)

## Additional notes

I'll just write some more tests and play it a bit longer to make sure there are no bugs and i will open this for merge.


JonaLam (dismissed):

Cards/Targeting/TargetAll.gd
```diff
@@ -0,0 +1,13 @@
+class_name TargetAll extends TargetingBase
+
+func _init() -> void:
+    cast_type = Enums.CardCastType.INSTA_CAST
+    application_type = Enums.ApplicationType.ALL
```
cheesefrycook, line 5: Is `application_type` being used anywhere? We still have the `application_type` on `CardBase` that is being used in `can_play_card`, but it doesn't seem like the ones in the targeting functions are being used.

If we want to store `application_type` in the targeting function, then I guess we will need to loop through each effect and consider if the card can be played based on all of the `application_type`

Cards/CardContainer.gd
cheesefrycook, line 1: Can we remove `battler_refrence` now that we can get the enemy list via `EnemyManager.enemy_list`?

Core/Battler.gd
```diff
@@ -110,22 +111,7 @@ func _try_player_play_card_on_entity(entity: Entity) -> void:
...
-		Enums.ApplicationType.ENEMY_ONLY:
-			all_target = _enemy_list
-		Enums.ApplicationType.FRIENDLY_ONLY:
-			all_target = [PlayerManager.player]
-			
-	return all_target
+			CardManager.card_container.play_card(entity)
 
   
 func _handle_enemy_deaths() -> void:
```
cheesefrycook, line 117: Should we remove the enemy from `EnemyManager.enemy_list` too when they die?

cheesefrycook:
We might want a `TargetCaster` targeting type so enemies could heal themselves

Cards/Targeting/TargetAll.gd
Turtyo: I think it's more useful as a way to check that everything is working correctly (that we target the correct thing)
but indeed i don't think I used them explicitly in the targeting

Core/Battler.gd
Turtyo: list is passed by reference, meaning if you remove an enemy from the original list, it will also be removed in the copy
i can add assert in the tests to make sure the two lists stay equal if you would like

Turtyo:
I can maybe do a TargetSelf yeah

Cards/CardContainer.gd
Turtyo: yes I removed it, it was only declared as variable

Turtyo:
Ah nevermind, a TargetSelf would be pretty hard to do :thinking: Because the targeting only depends the card and not who is playing it. I can do a `TargetPlayer`, but for enemies they need to use a `TargetBase` with themselves as the target. Which we can totally do when we build their attack, we call `targeted_entity` as self

Cards/Targeting/TargetAll.gd
Turtyo: I also realized when reading this again that it wouldn't really make sense to go through all the effects to look at the application type.
Take a card that does 3 heal to self and 3 damage to an enemy. The `application_type` of one is `FRIENDLY_ONLY` and the other is `ENEMY_ONLY`, but the card is totally valid

Core/Battler.gd
cheesefrycook: Oh cool, didn’t know it did that. An assert would be good

Turtyo: i added a test for that

Turtyo:
Correction on my previous message, the card depends on who plays it (for damage calculations with stats).
But **the targeting** doesn't depend on who plays the card, which is why a target self is hard to do

cheesefrycook (dismissed):

Cards/CardBase.gd
```diff
@@ -32,47 +32,19 @@ func parse_card_data(card_data: Dictionary) -> void:
 	pass
 
 
-func _apply_all_effects(caster: Entity, targets: Array[Entity], effects: Array[EffectData]) -> void:
-	for target : Entity in targets:
-		for effect : EffectData in effects:
-			effect.apply_effect_data(caster, target)
+# func _apply_all_effects(caster: Entity, targets: Array[Entity], effects: Array[EffectData]) -> void:
```
Tysterman74, line 35: Do we still need this function since it's been commented out?

Cards/Effects/EffectData.gd
```diff
@@ -4,19 +4,18 @@ class_name EffectData extends Resource
 
 @export var effect: EffectBase = null
 var caster: Entity = null
-var list_targets: Array[Entity] = []
 @export var value: int = 0
-@export var target_type : Enums.TargetType = Enums.TargetType.ALL_TARGETS
+@export var targeting_function: TargetingBase = null
 
-func _init(_effect: EffectBase = null, _caster: Entity = null, _list_targets: Array[Entity] = [], _value: int = 0) -> void:
+func _init(	_effect: EffectBase = null,
```
Tysterman74, line 10: Extra whitespace here on the `_effect` argument

Turtyo: it was to have everything line up i think

Cards/CardBase.gd
Turtyo: i forgot to remove it

cheesefrycook (approved):

Tysterman74 (approved):

Issue #45, opened 2024-01-18
Keeping informations related to the player between fights


Turtyo:
## Is your amelioration related to a problem? Please describe.

Some data currently attached to entities is supposed to be carried between fights (like stats with permanent modifications, or HP). Since the `Entity` object is destroyed at the end of the combat, this information is not kept around.

## Describe the solution you'd like
Have a reference in the `PlayerManager` (that says around between combats if i'm not mistaken ?) to stats that are to be kept.
We can have a `PlayerHP` variable and a `PlayerStats` variables. Once the combat ends, the values of the entity are copied there before the entity is destroyed.
When entering the next room, if we spawn a player entity, the data is copied from the `PlayerManager`. 

## Describe alternatives you've considered
An alternative would be to reword the entire system to move all this system in the `PlayerManager` directly, but that seems like it would be a bigger rework for not much. Since the global structure is already done, there would be a lot references to change probably

## Additional context
This can be started before #43 is done, but it can't be closed before as it will need to reference Stats anyway

Tysterman74:
I wonder if this can be an introduction to a basic save functionality. We can save the player stats to some external file then just read from it and populate the PlayerManager with those suggested changes you mentioned.

Turtyo:
Mmh, I think we should do the player manager first before wanting to do a save system.
The save system would not only need to keep the HP and stats from the player manager, but it would also need item list, card list, torches positions on the map and player position.

The save system is something bigger than just the player manager whose only goal is to enable the player to keep things between fights in the same play session. And directly going to save all this data externally just to load it back at each combat just doesn't sit right to me. Like yes sure, the computer can do it fast you won't even notice, but it seems rather inefficient anyway.

cheesefrycook:
That's a good point, we will want to start figuring out some save/load functionality for all the data we're storing in global AutoLoads PlayerManager, CardManager, MapManager. But I agree that we shouldn't be saving and loading data during a run, seems a bit strange.

For a save system, I think it makes sense to export all the relevant data from each manager to an external file whenever you quit the game, then reload the data from the external file into each manager when the run is started again.

Maybe it'd be good to start an issue related to saving/loading soon. I've found that it's way easier to get the save system done early

Turtyo:
Yes, getting the data of each manager into an external file seems good. The only thing with that is that you wouldn't be able to save mid fight / mid event (probably ?)
I will open an issue for the load/save.

[...]

PR #68, opened 2024-02-13
Issue-45: Data persistence between fights

cheesefrycook:
# Description

Adds a structure for saving and loading data when a scene reloads.

## Related issue(s)

#45 

## List of changes

- Persist player data when they are destroyed and recreated by storing data in PlayerManager.
  - Player's stats from `StatComponent`
  - Player's `current_health` from the `HealthComponent`
- Save our deck of cards in the CardManager instead of CardContainer so it can persist between scene loads (if we later have a shop that lets you add new cards to your deck, for example)

## Tests

I wrote a few tests that delete the player and CardContainer to make sure that the correct data is loading whenever the object is recreated. I'm open to suggestions if anyone has more ideas for what I can test.

## Additional notes

Later this system will be combined with the save system. I was going to combine the save system with this PR, but I figured it's easier to review if they are separate.


Turtyo:
Btw @cheesefrycook , you can automatically link an issue by using a keyword like "Closes" or "Resolves"
Full list of keywords [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

Core/PlayerPersistentData.gd
```diff
@@ -0,0 +1,7 @@
+extends Resource
+class_name PlayerPersistentData
+## Store player data that we want to persist between scenes here.
+
+
+var saved_health: float = -1
+var saved_stats: EntityStats = null
```
Turtyo, line 7: I would use an `_init` to set the default values
i had some problem later in other parts of the code where if you extend class from this class and you want different default value, you won't be able to do it if you don't use an `_init`. Probably not useful here, but for the sake of consistency

Entity/Components/StatComponent.gd
```diff
@@ -9,5 +9,9 @@ class_name StatComponent
 var _stats: EntityStats = EntityStats.new()
 
 
+func set_stats(new_stats: EntityStats) -> void:
```
Turtyo, line 12: I found out about the setter-getter syntax for variables after writing this, but it could be nice to implement here
Cf [this link](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#properties-setters-and-getters)

Entity/Player/Player.gd
```diff
@@ -2,11 +2,50 @@ extends Entity
...
+	
+	_load_persistent_data()
 
 
 func _exit_tree() -> void:
+	# Save player data when they are destroyed (combat end)
 	PlayerManager.set_player(null)
+	
+	if _should_save_persistent_data:
+		_save_persistent_data()
```
Turtyo, line 20: shouldn't the save happen before setting the player to null ?

Managers/PlayerManager.gd
```diff
@@ -5,12 +5,21 @@ extends Node
...
 	if player != null:
 		on_player_initialized.emit()
+
+
+func create_persistent_data() -> void:
+	PlayerManager.player_persistent_data = PlayerPersistentData.new()
+
+
+func get_persistent_data() -> PlayerPersistentData:
+	return PlayerManager.player_persistent_data
```
Turtyo, line 25: why not just `return player_persistent_data` ?

Managers/PlayerManager.gd
```diff
@@ -5,12 +5,21 @@ extends Node
...
 	player = in_player
 	if player != null:
 		on_player_initialized.emit()
+
+
+func create_persistent_data() -> void:
+	PlayerManager.player_persistent_data = PlayerPersistentData.new()
+
+
+func get_persistent_data() -> PlayerPersistentData:
```
Turtyo, line 24: could use a getter directly with the declaration of the variable (same as another comment I made)

Entity/Player/Player.gd
```diff
@@ -2,11 +2,50 @@ extends Entity
...
+
+
+# Set values from our saved persistent data
+func _load_persistent_data() -> void:
```
Turtyo, line 24: we might need to set other stuff later (like objects for example, though that might be handled separately of the player, in the same way that cards and deck is handled separately

Managers/CardManager.gd
```diff
@@ -14,8 +17,25 @@ func set_card_container(in_card_container: CardContainer) -> void:
...
 	
+	if in_card_container != null:
+		_initialize_deck()
+
+
+func _initialize_deck() -> void:
+	# Set our default deck from the card container.
+	# NOTE: We assume that if the current_deck is empty, then this is the first time the game was loaded.
+	# So we load the default deck stored in the card container.
+	# TODO: It would be better to store default_deck in a separate Resource file
```
Turtyo, line 28: indeed
and it's not really clear currently in the code where the default deck is (at least for me)

Tests/test_persistence.gd
```diff
@@ -0,0 +1,78 @@
...
+	card_damage.on_card_play(_player, [_enemy])
+	assert_eq(_enemy_health_component.current_health, 96.0)
+	
+	_player._should_save_persistent_data = true
+	
+	_player.free()
+	_create_player()
+	
+	# player should still have their permanent buff after being destroyed and recreated
+	card_damage.card_effects_data[0].value = 2
```
Turtyo, line 49: no need for this line, the card isn't destroyed in this context, so the value is still 2

Turtyo (changes requested):
This seems solid, a few changes are needed I think
i'll try to think of a few more tests you could do, but one that comes to mind is to reload the entire scene and check that everything is kept (i'm not sure this can be done, I remember some problems of the same nature with the test of death handling)

Entity/Player/Player.gd
cheesefrycook: It doesn't really matter since all PlayerManager.set_player does is remove the reference in PlayerManager

Entity/Player/Player.gd
cheesefrycook: What do you mean by objects? Like player's inventory?

Managers/CardManager.gd
cheesefrycook: The default deck is stored in the `CardContainer`, which doesn't make much sense any more. Should I go ahead and move it to a resource in this PR?

Entity/Player/Player.gd
Turtyo: yes, items
not really anything to change based on my comment, i was just commenting on that

Managers/CardManager.gd
Turtyo: Yes, it makes sense to do it now

Entity/Components/StatComponent.gd
cheesefrycook: Now that I'm looking at this I don't think it makes much sense. Because the way we call the getter would be by doing `get_stat_component()._stats` right? I think the explicit `get_stats` is more clear. Unless I misunderstand this syntax

Managers/PlayerManager.gd
cheesefrycook: In that case the syntax would be `PlayerManager.player_persistent_data` when you want to get the data right? Is there any benefit to doing that?

Entity/Components/StatComponent.gd
Turtyo: doing `get_stat_component()._stats` basically does `get_stat_component().get_stats()`
I find it pretty clear

Cards/CardSet.gd
```diff
@@ -0,0 +1,5 @@
+extends Resource
+class_name CardSet
+
+
+@export var card_set: Array[CardBase] = []
```
Turtyo, line 5: i generally don't like when the variable has the same name as the class because it can be confusing
`CardSetBase` for the class is probably better

Managers/CardManager.gd
```diff
@@ -9,6 +9,11 @@ signal on_deck_initialized
 var card_container: CardContainer = null
 # We store the current deck in CardManager because we may want to access it outside of battle.
 var current_deck: Array[CardBase] = []
+var default_deck: CardSet = null
+
+
+func _ready():
+	default_deck = load("res://Cards/CardSets/DefaultDeck.tres")
```
Turtyo, line 16: where was it loaded from before ? from the scene directly ? if so, should the scene file be edited to remove the deck from there ?

cheesefrycook: It was in CardContainer and set in the inspector using an @export. Since this is an AutoLoaded singleton, we can’t use the inspector to set it so that’s why I’m loading it like this. I already removed the default_deck from CardContainer

Entity/Components/StatComponent.gd
```diff
@@ -6,12 +6,12 @@ class_name StatComponent
...
-func set_stats(new_stats: EntityStats) -> void:
-	_stats = new_stats
+var stats: EntityStats = EntityStats.new():
+	set(value):
+		stats = value
+	get:
+		return stats
 
 
 func get_stats() -> EntityStats:
```
Turtyo, line 16: this is not needed anymore since we have the getter for stats already
if you call `get_stats` you basically cal `return return stats`

cheesefrycook: So if I remove `get_stats` I'll have to update all of our calls to `get_stats` to just `stats` right? Is that okay?

Turtyo: seems okay yes :thinking: 
you could probably do a search in all files and replace, then reload and check there are no problems
I don't think there are other functions called `get_stats`

Entity/Components/StatComponent.gd
```diff
@@ -6,8 +6,8 @@ class_name StatComponent
...
 
-var _stats: EntityStats = EntityStats.new()
-
-
-func get_stats() -> EntityStats:
-	return _stats
+var stats: EntityStats = EntityStats.new():
+	set(value):
+		stats = value
+	get:
```
Tysterman74, line 12: Godot has getter/setters? 🤯 Nice.

Entity/Player/Player.gd
```diff
@@ -2,11 +2,50 @@ extends Entity
 class_name Player
 ## Base player Entity class.
 
+var _should_save_persistent_data: bool = true
```
Tysterman74, line 5: I only see usage of this variable in the tests. Will this be used outside of testing?

Tysterman74 (approved):

Entity/Player/Player.gd
Turtyo: It'll probably be useful when implementing the external storage (not the one between fights), if I understood Cheesy's goal correctly

cheesefrycook: I'm not sure if it will be used for saving actually. We just don't want to have persistence with the player when testing because when a new Player is spawned in a new test then the data will carry over

Tysterman74: Okays. Just checkin 👍🏽

Turtyo (approved):

Tysterman74 (approved):

PR #69, opened 2024-02-13
Remove unused function parameters / variables and add missing typing

### LINKED ISSUE #65: Remove unused function parameters / variables and add missing typing

There are a lot of unused function parameters in the code (I count 7), which produces warning messages that could be avoided.
There are also quite a lot of untyped variables or functions with no return type.

Either remove the function parameter / variable if you can, or put a `@warning_ignore("unused_parameter")` before the function (only if you are overriding a function or using a built-in, which forces you to have the same signature even if you don't use all the variables).
For the untyped variables, type them properly.

## Additional context
Here is a list of all the places i found with such a problem:
- `res://Cards/CardContainer.gd`
    - `func finish_active_card_action(card: CardBase) -> void:`
    - `var move_state: Enums.CardMovementState = movement_component.current_move_state` in `func _update_card_positions() -> void:`
- `res://Cards/Movement State/CardMovementState.gd`
    - `func on_state_process(delta: float) -> void:`
    - `func can_transition_from(new_state: Enums.CardMovementState) -> bool:`
- `res://Cards/Movement State/MoveState_Discarding.gd`
    - `func can_transition_from(new_state: Enums.CardMovementState) -> bool:`
- `res://Cards/CardMovementComponent.gd`
    - `func _process(delta: float) -> void:`
- `res://UI/EndTurnButton.gd`
    - `func _process(delta: float) -> void:`

There are a lot of untyped variables so i'm not gonna list all of them

To see all of that, you can go into Project -> Project Settings, activate Advanced Settings then go to Debug -> GDScript and activate `Untyped declaration`, `Unsafe cast` as a warn. `Unused variable` and `Unused parameter` are already set to warn by default.. I think all of those options should be set as an error after the PR related to this issue. Warn can be ignored so we could potentially introduce new untyped declarations or unused variables. If we set it as an error, it won't be possible to ignore it.

Turtyo:
# Description

Removed unused code and added static typing.

## Related issue(s)

Closes #65 

## List of changes

- Add underscore to unused function parameters or a `warning_ignore`
- Remove unneeded variables
- Add typing to every missing place in the code
- Set warnings level of unused parameter and untyped variables to error level. New code won't be able to ignore untyped code from now on (which could be missed with a simple warning if one doesn't watch the log console).

## Tests

Just added typing to existing tests.

## Additional notes

Had to modify a part of GUT to be able to pass untyped var at an error level (but I tested and there doesn't seem to be any problem). This shouldn't be a problem since we won't update GUT in the project.

### DIFF
--- a/Cards/CardContainer.gd
+++ b/Cards/CardContainer.gd
@@ -398,7 +398,6 @@ func _update_card_positions() -> void:
 	for card_index: int in amount_of_cards:
 		var card: CardWorld = cards_in_hand[card_index]
 		var movement_component: CardMovementComponent = card.get_card_movement_component()
-		var move_state: Enums.CardMovementState = movement_component.current_move_state
 		
 		var card_x: float = per_card_separation * card_index
 		var card_y: float = 0.0
--- a/Cards/CardContainer.gd
+++ b/Cards/CardContainer.gd
@@ -52,7 +52,9 @@ var _discard_timer: SceneTreeTimer = null
 func _ready() -> void:
 	PhaseManager.on_phase_changed.connect(_on_phase_changed)
 	CardManager.set_card_container(self)
-	CardManager.on_card_action_finished.connect(finish_active_card_action)
+	CardManager.on_card_action_finished.connect(finish_active_card_action.unbind(1))
+	# ? all calls functions connected to on_card_action_finished don't even use the data of the card passed with the signal
+	# ? does the signal even need to pass the card data since no function connected uses it ?
 	
 	_init_default_draw_pile()
 
@@ -85,7 +87,7 @@ func set_active_card(card: CardWorld) -> void:
 	_active_card = card
 
 
-func finish_active_card_action(card: CardBase) -> void:
+func finish_active_card_action() -> void:
 	_discard_active_card()
 	_active_card = null
 
--- a/Cards/CardContainer.gd
+++ b/Cards/CardContainer.gd
@@ -75,7 +75,7 @@ func queued_for_active() -> void:
 
 
 func _remove_queued_card_from_hand() -> void:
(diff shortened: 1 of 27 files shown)


Cards/CardContainer.gd
```diff
@@ -52,7 +52,9 @@ var _discard_timer: SceneTreeTimer = null
 func _ready() -> void:
 	PhaseManager.on_phase_changed.connect(_on_phase_changed)
 	CardManager.set_card_container(self)
-	CardManager.on_card_action_finished.connect(finish_active_card_action)
+	CardManager.on_card_action_finished.connect(finish_active_card_action.unbind(1))
+	# ? all calls functions connected to on_card_action_finished don't even use the data of the card passed with the signal
+	# ? does the signal even need to pass the card data since no function connected uses it ?
```
cheesefrycook, line 57: I'd prefer to keep it just encase we need it at some point

cheesefrycook (dismissed):

Cards/CardContainer.gd
Turtyo: fair enough
I didn't remove it directly precisely because i thought it could be useful

Event/EventBase.gd
EChisholm, line 1: I agree that using an enum for this feels like a natural choice. Worth keeping in mind as we develop events in the future.

EChisholm (approved):
This all seems like some fair cleanup to me! 👍

Event/EventBase.gd
Turtyo: should I change it now ?

Turtyo: I decided against changing it now, as it's an issue for the map and out of the scope of this PR

cheesefrycook (approved):

EChisholm (approved):

PR #70, opened 2024-02-14
Issue-9: Light Mechanic

### LINKED ISSUE #9: Light mechanic
## Feature Description

Players can place a light source in the room they are in currently after the event of said room is finished, which provides light to adjacent rooms. The lit state of the room influences the event in the room by changing the difficulty / loot. Unlit rooms also don't display the icon of their event until you are just next to said room.

## Motivation

The idea of using light has multiple goals:
- It lets the player influence the map, which is an interesting mechanic
- It creates uncertainty, since you can't see the entire map
- It creates a risk/reward situation (using more torches reduces difficulty but the loot is less good)
- It's lore relevant since the story says the tree is corrupted by darkness, lighting it up makes sense. And Fauna also forgot multiple times to light the tree which led to accidents.

## Acceptance Criteria

- Light is placed after event in the current room and before leaving the room
- The rooms outside of the range are not revealed, the rooms in the range show the icon of the event in the room

## Proposed Solution

You start the map at node 0. There is a torch placed here by default and you can see everything up to 3 node in every direction you can move to, so you can plan your route. This area is lit and combats/events are normal difficulty in the light radius.

You work your way up to node 3 and can now only see node level 4 that is dimly lit since it is not within torch radius. Tutorial prompts you that these combats/events will be more difficult but with greater reward (risk reward). Node levels 5 and beyond are in darkness and cannot be seen until you move next to them and they become dimly lit.

Place a torch? Y/N

You place one of your 3 torches, not feeling overconfident, and the next 3 levels are lit up(4-6). You take a path leading to 2 combats on levels 4 and 5, but you mess up the first and are worried. You can place a second torch within the radius of the first, which does 2 things. 1: causes 3 levels away to become lit(5-7). 2: Levels 5 and 6 are now brightly lit because they are in the radius of 2 torches, which makes combat/events easier.

You now have only 1 torch left, which lets you continue up to node level 10 at normal difficulty, or less at easier difficulty. You can buy more at a shop, and combats have a chance to drop one. If you are full and/or don't want the torch drop, you can pass on it for money or something simple. If you have torches remaining at the end of a map, you get extra money or bonus points or something.

- Add a level of light for rooms (probably an enum, 0 unlit, 1 dimly lit, 2 lit, 3 brightly lit)
- Make a `light` class to hold data for the light
- Make light interact with the map ie:
    - Lights should update the visibility of rooms
    - The level of light in the room can influences enemies and random event chances

## Additional Context

This issue is in link with the map implementation #6 and should probably be developed on the `map_implementation` branch.

Tysterman74:
# Description

Add a light mechanic to the MapUI, where player can place a light down in the current room the player is in.

## Related issue(s)

Closes #9 

## List of changes

- Added a Place Torch Button, which will show a pop-up confirmation to place a torch down
- Added Enum LightLevel which will define the LightLevel for a room
- Updated RoomBase with a signal to update RoomUI with the light level
- Updated RoomUI to update the Alpha of the RoomUI scene to signify the room's light level
- Updated MapUI to hold a 2D array of the MapUI elements to help with calculating the rooms that need light.
    - Added a variable to test torches being placed in other rooms

## Tests

Coming soon TM

## Additional notes

Some examples:

https://github.com/Saplings-Projects/1M_sub/assets/5923102/768a16d0-2f14-4630-be19-800c9dec7b45


Tysterman74:
Another example of the light mechanic mimic-ing the documentation image example:

298361688-45d83f7f-f449-4270-bcb6-0b38da52b82e

https://github.com/Saplings-Projects/1M_sub/assets/5923102/42191272-6e47-4fe6-9447-dfdd8995d8f7

Tysterman74:
Not sure how to go about the "darkness" of the rooms. Right now I have a ColorText overlayed on top of the RoomUI that has an alpha set to max and I feel that's gonna change once we get assets. Some ideas on how to go about that would be helpful!

Turtyo:
I think we talked a bit about the way to show rooms changing level of light and we came up with a few ideas. One of them was to have an overlay over the map (and not in the room directly).
Using the alpha was one of the idea for this overlay over the map. The advantage of having an overlay was the ability of doing more than just changing the room, but also its surrounding. That way every room that is lit could have a lit halo of light around it and it would look better later on.

Another idea I just had was to link the level of a light to an image of the room, and to cycle through that when updating the room. So you would have the image for the room at light level 0 (which would be dark), an image for the room at light level 1, light level 2... Maybe using something as a dictionary for that, or an array, would work nicely.

I also notice from the images alone that you forgot one level of light. When the player stands next to rooms that are not lit by any light, the rooms become dimly lit (which is not the case on the videos for now).

Map/MapManager.gd
```diff
@@ -2,7 +2,9 @@ extends Node2D
 ## Class to manage backend for rooms (generation and such)
 
 var current_map: MapBase
-var map_width_array: Array[int] = [1, 3, 5, 7, 9, 11, 9, 7, 5, 3, 1]
+var map_width_array: Array[int] = [1, 3, 5, 7, 7, 7, 7, 5, 3, 1]
+var LIGHT_RANGE = 3
```
Turtyo, line 6: type int

Map/MapManager.gd
```diff
@@ -25,13 +27,16 @@ func create_map(map_floors_width: Array[int]) -> MapBase: ## Generates and Popul
 			var _room_event: EventBase = GlobalVar.EVENTS_CLASSIFICATION[_rand_type_index].new()
 			var _generated_room: RoomBase = RoomBase.new()
 			_generated_room.room_event = _room_event
+			# Test statement to test player room placement, Remove later
+			if (index_width == 4 and index_height == 3):
```
Turtyo, line 31: we should have a way as dev team to put the player where we want on the map to not have to do that precisely
I'll try to think of a way in the map movement PR

Map/RoomBase.gd
```diff
@@ -4,12 +4,26 @@ class_name RoomBase
 ##
 ## Holds basic functionality for a room on the map
 
-@export var light_level: int = 0
+@export var light_level: Enums.LightLevel = Enums.LightLevel.UNLIT
```
Turtyo, line 7: this should be done in an _init function, if you have classes that extend from `RoomBase` it seems like the _init() of those child class to change default value doesn't work

Map/RoomBase.gd
```diff
@@ -4,12 +4,26 @@ class_name RoomBase
...
+
 func _to_string() -> String:
 	return "RoomBase"
 
 func get_room_abbreviation() -> String:
 	return room_event.get_room_abbreviation()
+
+func set_torch_active() -> void:
+	has_torch = true
+	if light_level == Enums.LightLevel.DIMLY_LIT or light_level == Enums.LightLevel.LIT:
```
Turtyo, line 21: this is wrong
an area that is dimly lit means the player is standing next to it and there are no torches in the range
a room that is dimly lit becomes lit if a torch is placed

I would have set the new light level by just checking the current light level and adding 1 to take the next element of the enum, with the exception of being at the max brightness where it does nothing

Map/RoomBase.gd
```diff
@@ -4,12 +4,26 @@ class_name RoomBase
...
+		light_level = Enums.LightLevel.BRIGHTLY_LIT
+	else:
+		light_level = Enums.LightLevel.LIT
+	on_light_level_changed.emit()
+
+func set_light_level(new_light_level: Enums.LightLevel) -> void:
```
Turtyo, line 27: I would write a setter directly when declaring the variable, cf [this](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#properties-setters-and-getters)
This would allow you to update everything in the `set_torch_active` directly, and you wouldn't need to have a `on_light_level_changed.emit()` at the end of the function

Map/RoomUI.gd
```diff
@@ -1,7 +1,39 @@
...
+func _ready():
+	if room != null:
+		if !room.on_light_level_changed.is_connected(_update_room_UI):
+			room.on_light_level_changed.connect(_update_room_UI)
+		_update_room_UI()
 
 func set_label(title: String):
 	label.set_text(title)
+
+func set_player_icon(enabled: bool) -> void:
```
Turtyo, line 21: activate instead of set maybe ? this name makes it feel like i'm gonna change the player icon when what it does is make it appear on screen
activate maybe not the best, because it could be with enabled = false
`toggle_player_icon` ?

Map/RoomUI.gd
```diff
@@ -1,7 +1,39 @@
...
+	player_icon.visible = enabled
+
+func has_player() -> bool:
+	return player_icon.visible
+
+
+func _update_room_UI():
```
Turtyo, line 28: type void

Map/RoomUI.gd
```diff
@@ -1,7 +1,39 @@
...
+	var alpha = 0
+	match (room.light_level):
```
Turtyo, line 30: an enum with alpha levels ? alpha = Enums.AlphaLevel[room.light_level as int]
though that might change with the ideas I gave for the actual display before doing this review

UI/MapUI.gd
```diff
@@ -5,11 +5,17 @@ var room_ui: PackedScene = load("res://Map/RoomUI.tscn")
 var _padding_offset = 20
 var _MINIMUM_ROOM_WIDTH = 510
 var _MINIMUM_ROOM_HEIGHT = 490
+var _LIGHT_FLOOR_RANGE = 3
```
Turtyo, line 8: type int

UI/MapUI.gd
```diff
@@ -18,6 +24,59 @@ func _input(_inputevent: InputEvent) -> void:
...
+	
+
+func _on_add_torch_pressed():
+	torch_confirmation_dialog.show()
+
+func _close_torch_placement_dialog():
+	torch_confirmation_dialog.hide()
+
+func _add_torch_to_current_location():
+	var current_player_floor_index = current_player_room.floor_index
```
Turtyo, line 36: type int
and missing typing in multiple places in this file actually

UI/MapUI.gd
```diff
@@ -18,6 +24,59 @@ func _input(_inputevent: InputEvent) -> void:
 
 func _on_return_button_press():
 	queue_free()
+	
+
+func _on_add_torch_pressed():
```
Turtyo, line 29: return type void

UI/MapUI.gd
```diff
@@ -18,6 +24,59 @@ func _input(_inputevent: InputEvent) -> void:
...
+
+func _close_torch_placement_dialog():
+	torch_confirmation_dialog.hide()
+
+func _add_torch_to_current_location():
+	var current_player_floor_index = current_player_room.floor_index
+	var current_player_room_index = current_player_room.room_index
+	
+	var room_light_range = 1
+	var max_light_index = current_player_floor_index + _LIGHT_FLOOR_RANGE + 1
```
Turtyo, line 40: `max_light_height_index` ?
and why the +1 at the end ? if player is on floor 1, and the _LIGHT_FLOOR_RANGE is 1, it means the torch placed at level 1 is doing light until the level 2 (a range of 1)
unless you consider the current room as being the range of 1 instead of a range of 0 ?

UI/MapUI.gd
```diff
@@ -18,6 +24,59 @@ func _input(_inputevent: InputEvent) -> void:
...
+	torch_confirmation_dialog.hide()
+
+func _add_torch_to_current_location():
+	var current_player_floor_index = current_player_room.floor_index
+	var current_player_room_index = current_player_room.room_index
+	
+	var room_light_range = 1
+	var max_light_index = current_player_floor_index + _LIGHT_FLOOR_RANGE + 1 
+	
+	if current_player_floor_index + _LIGHT_FLOOR_RANGE + 1 > MapManager.map_width_array.size():
```
Turtyo, line 42: isn't that already `max_light_index` ? no need to recalculate it

UI/MapUI.gd
```diff
@@ -18,6 +24,59 @@ func _input(_inputevent: InputEvent) -> void:
...
+		max_light_index = MapManager.map_width_array.size() 
+	
+	for floor_light_index: int in range(current_player_floor_index + 1, max_light_index):
```
Turtyo, line 45: I feel like all of what is done after in this function is setting level level in rooms, which i feel should be done in `_add_light_to_rooms` ?
I would also have done this recursively instead of a for loop. Start with `remaining_light_range` equal to the how far the light goes from a torch (so 3 for us, idk which const does that exactly).
Look at the rooms accessibly by the player and then two things happen:
- the remaining light range is 1, change the light level and return
- the remaining light range is not 1, apply this function to all accessible rooms and change the light level of the room you are looking at

For me this is much more intuitive to look at than for loops with carefully chosen bounds at each step, which is very prone to index errors also

UI/MapUI.gd
```diff
@@ -59,15 +118,30 @@ func _ready():
...
+	for floor_index: int in range(current_map.rooms.size()):
+		var floor_array: Array = current_map.rooms[floor_index]
 		# When we're done populating a floor and we go to the next index, reset the X start position
 		position_for_next_room.x = start_position_for_next_room_x
-		for room: RoomBase in floor_array:
-			if (room != null):
-				var room_display: Control = room_ui.instantiate()
+		for room_index: int in range(floor_array.size()):
+			var room: RoomBase = floor_array[room_index]
+		#for room: RoomBase in floor_array:
```
Turtyo, line 130: is this comment still needed ?

Turtyo (changes requested):
I think the logic of setting the light level in a room could be wrote much more simply using recursion.
i detailed the logic in a related comment

Turtyo:
Regarding my comment about the rooms you can access to know which rooms to put light in, I'm gonna write a function that gives you all the rooms you have access to when from a given position on the map, so you will be able to use that

UI/MapUI.gd
Turtyo: i'm realizing you will also need to have a boolean to know if a room had its light changed already by another function call during this light change.
Because you could access the same room in multiple ways, like going left-right or right-left will lead you to the same room (when i say right I mean up-right in diagonal, we don't stay on the same floor)
so something like that:
```GDScript
func _add_light_to_rooms(floor_index: int = current_player_room.floor_index, room_index: int = current_player_room.room_index, remaining_light_range: int = TOTAL_LIGHT_RANGE, is_room_already_changed: bool = false) -> bool:
	if not is_room_already_changed:
		if remaining_light_range == 1:
			# change the light
			return true
		else:
			var accessible_rooms: Array[RoomBase] = get_all_accessible_rooms # to implement
			for current_room: RoomBase (RoomUI ?)  in accessible_rooms:
				_add_light_to_rooms(current_room.floor_index, current_room.room_index, remaining_light_range -1, false)
			# change the light in the current room
	return true
```
and writing this noticed that a boolean wouldn't work :laughing: maybe returning a list of all the rooms that are already changed (by returning a hash of the room for example, or an object ID when changing the light), and checking that the `current_room` is not in the list of already changed rooms ?

Map/MapManager.gd
Tysterman74: I'm thinking of opening up an export variable on the MapUI to force a player position. Since your PR will focus on setting what the player position is, this export I have will just override it so we can use that for testing.

Map/RoomBase.gd
Tysterman74: So to summarize:

NOT_LIT: 

1. No torch light in range, No player adjacent

DIMLY_LIGHT: 

1. Player is in the Room, but no torch
2. Adjacent rooms that the Player is in if there are no torches placed
3. Non-Adjacent rooms that the Player is in, but is in range of torch light

LIT: 

1. Player is in the room AND torch has been placed down
2. Any other room that is DIMLY_LIT and falls within the torch range

BRIGHTLY_LIT:

1. Rooms that are LIT and falls within another torch's range

Map/MapManager.gd
Turtyo: seems good

Map/RoomBase.gd
Turtyo: DIMLY_LIT:
1. Player is in the room but no torch
2. Room is next to the player to player and there are no torches reaching this room
3. I'm not sure to understand but i don't see another case where it's dimly lit except the two previous cases

LIT:
1. Player doesn't need to be in the room for the room to be lit
2. Correct
3. You forgot the case of rooms in darkness that are not lit. If they are within torch range, they become lit

To sum up:
level_of_light_explanation

UI/MapUI.gd
```diff
@@ -2,17 +2,18 @@ extends Control
...
+var _LIGHT_FLOOR_RANGE: int = 3
 
 @export var color_rect: ColorRect
 @export var scroll_container: SmoothScrollContainer
 @export var room_container: ColorRect
 @export var room_addition_node: Control
 @export var torch_confirmation_dialog: ConfirmationDialog
-@export var test_torch_indices: Dictionary
+@export var test_torch_indices: Array[Vector2]
+@export var test_player_position: Vector2
```
Turtyo, line 16: this can be a `Vector2i` since positions will be on the map and that's just integers

UI/MapUI.gd
```diff
@@ -60,7 +61,7 @@ func _add_torch_to_current_location():
 	current_player_room.room.set_torch_active()
 
 # Test function
-func _add_light_to_rooms(floor_index: int, room_index: int) -> void:
+func _test_add_light_to_rooms(floor_index: int, room_index: int) -> void:
```
Turtyo, line 64: if it's a test function, why is not in a test file ? or is more of a WIP function ?

Map/LightNode.gd
```diff
@@ -0,0 +1,32 @@
+extends Node2D
+class_name LightNode
```
Turtyo, line 2: so this light node comes over each room of the map ?

Map/LightNode.gd
```diff
@@ -0,0 +1,32 @@
...
+
+func _init(position: Vector2, room_ui: RoomUI):
+	_position = position
+	_room_ui = room_ui
+	_room_ui.room.on_light_level_changed.connect(_update_light)
+	var new_room_texture_rect: TextureRect = Helpers.get_first_child_node_of_type(_room_ui, TextureRect)
+	_room_size = new_room_texture_rect.get_size()
+
+func _draw():
+	if _room_ui.get_light_level() == Enums.LightLevel.BRIGHTLY_LIT:
```
Turtyo, line 17: maybe a match case instead of if/elif would be better

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+	room_container = _room_container
+	room_ui_array = _room_ui_array
+
+func _draw():
+	var floor_rect_packed_array: PackedVector2Array
+	var room_circles: Array[PackedVector2Array]
+	var lit_rooms_to_draw: Array[RoomUI]
+	
+	# Account for minimum map placement T____T
+	for floor_index: int in range(room_ui_array.size()):
```
Turtyo, line 21: why use a for with index instead of a for-loop on the values directly ?
I notice that you just use the index to get the actual floor, which could be summed up to:
`for floor: Array int room_ui_array`

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+		var floor_array: Array = room_ui_array[floor_index]
+		var _first_room_in_array: bool = false
+		
+		for room_index: int in range(floor_array.size()):
```
Turtyo, line 25: same question as previously, why not use a for loop over the elements directly ?

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+			if room != null:
+				var new_room_texture_rect: TextureRect = Helpers.get_first_child_node_of_type(room, TextureRect)
+				var room_size = new_room_texture_rect.get_size()
+				if (room.get_light_level() != Enums.LightLevel.UNLIT):
+					room_circles.append(_calculate_points_for_circle(room))
+					lit_rooms_to_draw.append(room)
+	floor_rect_packed_array.append(Vector2(0, 0))
+	floor_rect_packed_array.append(Vector2(0, room_container.get_size().y))
+	floor_rect_packed_array.append(Vector2(room_container.get_size().x, room_container.get_size().y))
+	floor_rect_packed_array.append(Vector2(room_container.get_size().x, 0))
```
Turtyo, line 36: what is the meaning of those 4 vectors being added ?

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+	
+	if room_circles.size() > 0:
+		var current_lit_room_polygon: PackedVector2Array = room_circles[0]
+		for room_circle_index in range(room_circles.size()):
```
Turtyo, line 40: missing int type

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+	# Account for minimum map placement T____T
+	for floor_index: int in range(room_ui_array.size()):
+		var floor_array: Array = room_ui_array[floor_index]
+		var _first_room_in_array: bool = false
+		
+		for room_index: int in range(floor_array.size()):
+			var room: RoomUI = floor_array[room_index]
+			if room != null:
+				var new_room_texture_rect: TextureRect = Helpers.get_first_child_node_of_type(room, TextureRect)
+				var room_size = new_room_texture_rect.get_size()
```
Turtyo, line 29: missing int type

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+	floor_rect_packed_array.append(Vector2(0, 0))
+	floor_rect_packed_array.append(Vector2(0, room_container.get_size().y))
+	floor_rect_packed_array.append(Vector2(room_container.get_size().x, room_container.get_size().y))
+	floor_rect_packed_array.append(Vector2(room_container.get_size().x, 0))
+	
+	if room_circles.size() > 0:
+		var current_lit_room_polygon: PackedVector2Array = room_circles[0]
+		for room_circle_index in range(room_circles.size()):
+			if (room_circle_index + 1 <= room_circles.size() - 1):
+				var merged_result = Geometry2D.merge_polygons(current_lit_room_polygon, room_circles[room_circle_index + 1])
```
Turtyo, line 42: missing typing

Map/LightOverlay.gd
Turtyo, line 40: if I understand this loop correctly, you are basically adding all the lit areas polygons between 1 and the last together ?
maybe you could use Array.slice() to do that ?

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+				current_lit_room_polygon = merged_result[0]
+
+		draw_polygon(current_lit_room_polygon, [Color(0, 0, 0, 0.25)])
+		var outer_polygons = Geometry2D.clip_polygons(floor_rect_packed_array, current_lit_room_polygon)
+		for outer in outer_polygons:
```
Turtyo, line 47: missing typing

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+			draw_polygon(outer, [Color(0, 0, 0, 1)])
+	
+	for room in lit_rooms_to_draw:
```
Turtyo, line 50: missing typing

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+
+func _draw_room_circle(room: RoomUI):
+	if room.get_light_level() == Enums.LightLevel.DIMLY_LIT:
+		#draw_circle(Vector2(room.get_center_X(), room.get_center_Y()), 60, Color(0, 0, 0, 0.5))
+		draw_rect(room.get_room_rect(), Color(0, 0, 0, 0.5))
+	elif room.get_light_level() == Enums.LightLevel.LIT:
+		draw_arc(Vector2(room.get_center_X(), room.get_center_Y()), 40, 0, TAU, 20, Color(1, 1, 0, 1), 3)
+		draw_circle(Vector2(room.get_center_X(), room.get_center_Y()), 39, Color(1, 1, 0, 0.1))
+
+func _calculate_points_for_circle(room: RoomUI) -> PackedVector2Array:
```
Turtyo, line 62: lots of typing missing in this function

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+	var rect: TextureRect = Helpers.get_first_child_node_of_type(room, TextureRect)
+	var rect_size = rect.get_size()
+	var center_point = Vector2(room.position.x + rect.get_size().x / 2, room.position.y + rect.get_size().y / 2)
+	var tick_count = 60
+	var display_angle = 360
+	var circle_points: PackedVector2Array
+	for i in range(1, tick_count):
+		var point_angle = -deg_to_rad(display_angle) / tick_count * i
+		point_angle = -deg_to_rad(display_angle) + (2 * PI / tick_count) * i
```
Turtyo, line 71: this line overrides the previous line immediately, what is the goal of having those two lines ?

Map/LightOverlay.gd
```diff
@@ -0,0 +1,77 @@
...
+		draw_rect(room.get_room_rect(), Color(0, 0, 0, 0.5))
+	elif room.get_light_level() == Enums.LightLevel.LIT:
+		draw_arc(Vector2(room.get_center_X(), room.get_center_Y()), 40, 0, TAU, 20, Color(1, 1, 0, 1), 3)
+		draw_circle(Vector2(room.get_center_X(), room.get_center_Y()), 39, Color(1, 1, 0, 0.1))
+
+func _calculate_points_for_circle(room: RoomUI) -> PackedVector2Array:
+	var rect: TextureRect = Helpers.get_first_child_node_of_type(room, TextureRect)
+	var rect_size = rect.get_size()
+	var center_point = Vector2(room.position.x + rect.get_size().x / 2, room.position.y + rect.get_size().y / 2)
+	var tick_count = 60
```
Turtyo, line 66: so basically we make a circle with 60 points on the outer edge ?

Turtyo (changes requested):
I think this implementation will work better, we can choose how we draw the shape around each room and it looks pretty good like that.
Do you think we will be able to add a fade around the edges for the lit rooms ? So it's not a circle that brutally goes to black, but the edge fades to black

UI/MapUI.gd
Tysterman74: Yeah, just wanted to see what the light would look like having torches in several places. This'll move to a test file eventually.

Map/LightOverlay.gd
Tysterman74: Ah good catch, it was some leftover code when I was trying to calculate each individual room position... Although in the latest push, I had to leave floor_index in order to help with some calculations on the darkness overlay portion.

Map/LightOverlay.gd
Tysterman74: I've left a comment in the file, but this'll also be updated as well to be split up.

These vectors are used to help find the polygon of the darkness overlay that surrounds the lit rooms.

I've updated the logic a bit from this comment, but I'll explain it in detail here:
We have two different packed arrays one for the upper half of the map at the player position (In red), and one for the bottom half of the map at the player position (In green). Anything that is in black in this example picture is the lit room polygon that is calculated in Line 60-64

Geometry2D.clip_polygons() is a function in Godot that will return the inverse polygon of two polygons that are passed in. HOWEVER, this only works if there is OVERLAP, if the polygons are fully encompassed or this is no overlap, then we won't get any polygons returned.
image

So we get the inverse of the top half, and draw that out (Line 67-69):
image

Then we get inverse of the bottom half, and draw that out (Line 70-72):
image

Tysterman74: I realized the naming's confusing, so I'll update it alongside some comments.

Map/RoomBase.gd
Tysterman74: Ok thank you for the flow chart, I'm gonna add this to the documentation_images folder.

Map/LightOverlay.gd
Tysterman74: That is the end product, but to achieve this we merge them individually and add it to the overall final polygon.

First pass through the array we merge the first circle and the second circle together:
image

This is now one polygon and I save this as the current polygon for all the lit rooms.
Second pass and subsequent passes we merge the next circle to the current polygon then save it as the new current polygon.

In terms of .slice(), I believe it should be something similarly achievable, but slice returns back another Array so there would have to be some extra querying from that new array to get the circles we want to merge. I think what I currently have and just looking at the single array may work best in this case without having to get the logic be too complicated.

Map/LightOverlay.gd
Tysterman74: Yep, we get those points of the circle in order to help with the calculations of showing the rooms that will have light. `draw_circle()` doesn't help here, because it will be overwritten by the darkness overlay and it doesn't give us the points of that circle so we need to calculate it manually.

Map/LightOverlay.gd
Tysterman74: I'll keep it 100%, I yoinked this chunk from reddit, but after messing around with the values a bit we really didn't need much of these lines, so I ended up refactoring it. Left some comments too, to help explain the logic behind the mathematics.

Tysterman74:
> I think this implementation will work better, we can choose how we draw the shape around each room and it looks pretty good like that. Do you think we will be able to add a fade around the edges for the lit rooms ? So it's not a circle that brutally goes to black, but the edge fades to black

Hmm I'm not understanding, I'll need a bit more details. An example would help.

Map/LightOverlay.gd
Turtyo: So basically you cut the map in half to draw a circle around the player ?

Turtyo:
Without fade around the edges (as it is currently):
image
With fade around the edges (so it doesn't go from clear to black directly)
image

Map/RoomBase.gd
```diff
@@ -4,12 +4,18 @@ class_name RoomBase
...
+@export var light_level: Enums.LightLevel:
+	set(value):
+		light_level = value
+		on_light_level_changed.emit(light_level)
 @export var has_torch: bool = false
 @export var room_event: EventBase
 
 signal on_light_level_changed
 
+func _init(_light_level: Enums.LightLevel):
```
Turtyo, line 16: it would be good to have a default argument if we want to create an instance of the class to use it after, without needing to supply an initial state

Map/LightOverlay.gd
Turtyo: I only see `floor_index` being used on a check that does
```GDScript
if floor_index == 0:
	first_room_rect = room.get_room_rect()
```
I think you could do a for loop directly on the array without the `floor_index` and setting `first_room_rect` to a default value before the loop over a given floor, then checking against this default value in the loop. So if the value is still the default, you know it's the first room of the floor

Map/LightOverlay.gd
Turtyo: Alright i see, both should be pretty equivalent, and if it works let's not fix it

Map/LightOverlay.gd
```diff
@@ -1,106 +1,114 @@
...
-		for outer in outer_polygons_top_half:
+		
+		# From the finished merged polygon blob from the lit room circles we created above, 
+		# Get the polygon inverse of the black overlay with the merged polygon blob we created above then draw them out.
+		var outer_polygons_upper_half = Geometry2D.clip_polygons(map_upper_half, current_lit_room_polygon)
+		for outer: PackedVector2Array in outer_polygons_upper_half:
 			draw_polygon(outer, [Color(0, 0, 0, 1)])
-		var outer_polygons_bottom_half = Geometry2D.clip_polygons(floor_rect_packed_array_bottom_half, current_lit_room_polygon)
-		for outer in outer_polygons_bottom_half:
+		var outer_polygons_bottom_half = Geometry2D.clip_polygons(map_bottom_half, current_lit_room_polygon)
```
Turtyo, line 72: missing typing
You should put the untyped var as a warning in the debug options (advanced parameters, debug, GDScript)
It's done on the main but we don't have it here yet, so that would probably help for the time being

Map/LightOverlay.gd
Turtyo: `draw_circle` doesn't give points ? does it give a polygon ?

Tysterman74: draw_circle() doesn't return back the points, it just draws the circle on the canvas :(

Map/LightOverlay.gd
Tysterman74: To draw a circle around all of the lit rooms.

Turtyo: Oh so you move the place where it's cut in half when you go to the next floor ?

Tysterman74: Not exactly. We just get the very first room. We just need need to cut it in half anywhere there's a circle in order to be able to get the inverse shape, since Geometry2D.clip_polygons will only return something if there's an intersection and not if it's fully covered.

Tysterman74:
Hmm actually setting the gradient size within RoomUI ends up messing up the polygons
image

There would need to be a bit more thought behind this to make it work. 🤔

Turtyo:
No problem, as I said we can do it later, for now let's try to merge the base implementation

UI/MapUI.gd
```diff
@@ -143,19 +144,33 @@ func _ready() -> void:
 		var floor_array: Array = current_map.rooms[floor_index]
 		# When we're done populating a floor and we go to the next index, reset the X start position
 		position_for_next_room.x = start_position_for_next_room_x
-		for room_index: int in range(floor_array.size()):
-			var room: RoomBase = floor_array[room_index]
-			if room != null:
-				var room_display: RoomUI = room_ui.instantiate()
+#<<<<<<< HEAD
```
Turtyo, line 147: you left some text here

UI/MapUI.gd
```diff
@@ -143,19 +144,33 @@ func _ready() -> void:
...
+				var room_display: Control = room_ui.instantiate()
+				var texture_button: TextureButton = Helpers.get_first_child_node_of_type(room_display, TextureButton)
+				# disable the button if the player can't access the room
+				texture_button.disabled = not (DebugVar.DEBUG_FREE_MOVEMENT or accessible_rooms_by_player.has(room))
 				room_display.room = room
+				if (PlayerManager.is_player_in_room(room)):
+					texture_button.disabled = true
+					texture_button.texture_disabled = room_with_player_texture
+				
+#>>>>>>> map-implementation
```
Turtyo, line 165: you left some text there

UI/MapUI.gd
```diff
@@ -181,11 +196,16 @@ func _ready() -> void:
 	# Hence we subtract half the size of the container from half the size of the height of the combined rooms to get the center point
 	# then subtract it from the position of the container
 	var new_room_position_y: float = room_container.position.y
+#<<<<<<< HEAD
```
Turtyo, line 199: you left text here

UI/MapUI.gd
```diff
@@ -181,11 +196,16 @@ func _ready() -> void:
...
 	# Get the offset if we had to adjust the Y position due to having to set a minimum height if the map is too small.
 	var offset_y: float = 0
-	if (_get_combined_room_height(new_room_texture_rect) < _MINIMUM_ROOM_HEIGHT):
-		new_room_position_y = room_container.position.y - room_container.get_custom_minimum_size().y / 2 + _get_combined_room_height(new_room_texture_rect) / 2
+	if (_get_combined_room_height(new_room_texture_button) < _MINIMUM_ROOM_HEIGHT):
+		new_room_position_y = room_container.position.y - room_container.get_custom_minimum_size().y / 2 + _get_combined_room_height(new_room_texture_button) / 2
 		offset_y = room_addition_node.position.y - new_room_position_y
+#=======
```
Turtyo, line 205: you left text here

Map/LightOverlay.gd
```diff
@@ -77,13 +77,14 @@ var room_circle_radius: int = 40
 func _draw_room_circle(room: RoomUI):
 	var center_point_with_offset: Vector2 = Vector2(room.get_center_X() - offset_position_x, room.get_center_Y() - offset_position_y)
 	if room.get_light_level() == Enums.LightLevel.DIMLY_LIT:
-		draw_circle(center_point_with_offset, room_circle_radius, Color(0, 0, 0, 0.5))
+		draw_arc(center_point_with_offset, room_circle_radius, 0, TAU, 20, Color(0, 0, 0, 0.1), 3)
```
Turtyo, line 80: what are the new parameters you added ?

Map/RoomBase.gd
```diff
@@ -7,13 +7,10 @@ class_name RoomBase
 @export var light_level: Enums.LightLevel:
 	set(value):
 		light_level = value
-		on_light_level_changed.emit(light_level)
```
Turtyo: why is it not needed anymore ?

Map/RoomBase.gd
```diff
@@ -29,8 +26,4 @@ func increase_light_level() -> void:
 
 func set_torch_active() -> void:
 	has_torch = true
-	if light_level == Enums.LightLevel.LIT:
-		light_level = Enums.LightLevel.BRIGHTLY_LIT
-	else:
-		light_level = Enums.LightLevel.LIT
-	on_light_level_changed.emit()
+	increase_light_level()
```
Turtyo, line 29: and if the room is unlit it will become DIMLY_LIT if we look at the enum of lights level, which is not what we want when setting a torch

Map/RoomUI.gd
```diff
@@ -32,4 +32,4 @@ func get_center_point() -> Vector2:
 	
 func _set_player_position_based_on_room():
 	PlayerManager.player_position = room.room_position
-	SignalBus.clicked_next_room_on_map.emit()
+	SignalBus.clicked_next_room_on_map.emit(self)
```
Turtyo, line 35: there is no need for it to emit self, since the signal itself doesn't emit anything if you look at the signal bus

Map/RoomBase.gd
Tysterman74: This was to handle updating the lightoverlay whenever light changes, but since light changes based on movement and putting a torch down which is done in MapUI, we don't need a signal to pick up when light is changed.

UI/MapUI.gd
Tysterman74: Everything in `_add_torch_to_current_location()` refactored to use `MapMovement.get_all_accessible_rooms_in_range()`

UI/MapUI.gd
Tysterman74: Everything in `_add_torch_to_current_location()` refactored to use `MapMovement.get_all_accessible_rooms_in_range()`

UI/MapUI.gd
```diff
@@ -55,12 +58,8 @@ func _add_torch_to_current_location() -> void:
 		for room_light_index: int in range(minimum_room_light_index, maximum_room_light_index):
 			var room_ui: RoomUI = room_ui_array[floor_light_index][room_light_index]
 			if room_ui != null:
-				if floor_light_index == current_player_floor_index + 1 and room_ui.room.light_level == Enums.LightLevel.DIMLY_LIT:
-					room_ui.room.light_level = Enums.LightLevel.BRIGHTLY_LIT
-				else:
-					room_ui.room.light_level = Enums.LightLevel.DIMLY_LIT
+				room_ui.room.increase_light_level()
```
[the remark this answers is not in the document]
Tysterman74: I realized we can't really determine light level within RoomBase since it depends if there's a torch in range or now, so I ended up scrapping it and just set the light level directly. If there's a better way to set it, I'm all ears.

Map/RoomUI.gd
Tysterman74: I've updated the signal to callback `_on_room_clicked()` in MapUI to help with setting light_level

Map/LightOverlay.gd
Tysterman74: Left a comment to help explain

Map/RoomBase.gd
Tysterman74: Same thing I found out with https://github.com/Saplings-Projects/1M_sub/pull/70#discussion_r1501738357

UI/MapUI.gd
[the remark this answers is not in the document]
Turtyo: I don't really see the problem
When you put a torch in a room, take the list of the rooms in the range of the torch (which you get via map movement).
Look at the light level and do the following:
- 0 -> 2
- 1 -> 2
- 2 -> 3
- 3 -> 3

Also when the player move, you also generate the list of new_accessible rooms and you do that on the light level:
- 0 -> 1
- 1 -> 1
- 2 -> 2
- 3 -> 3

Basically you just add one to the light level if it's unlit

i'm not sure to understand why that's not possible to do at the `RoomBase` level

Map/LightOverlay.gd
```diff
@@ -76,15 +76,24 @@ func _draw():
...
-	elif room.get_light_level() == Enums.LightLevel.LIT:
-		draw_arc(center_point_with_offset, room_circle_radius, 0, TAU, 20, Color(1, 1, 0, 1), 3)
-		draw_circle(center_point_with_offset, room_circle_radius - 1, Color(1, 1, 0, 0.01))
-	elif room.get_light_level() == Enums.LightLevel.BRIGHTLY_LIT:
-		draw_arc(center_point_with_offset, room_circle_radius, 0, TAU, 20, Color(0, 1, 1, 1), 3)
-		draw_circle(center_point_with_offset, room_circle_radius - 1, Color(0, 1, 1, 0.01))
+	# Discussions on customization with adding gradient to be done here. Currently we draw a circle around
+	# rooms that have a torch, and a rect around rooms that are lit from a torch
+	if room.has_torch():
+		# These draw calls draw an unfilled circle around the room with with a thicker width,
```
Turtyo, line 82: typo, you wrote "with" twice

Map/LightOverlay.gd
Turtyo: I see the explanation, but why the new TAU parameter and the 20 ?

Map/RoomBase.gd
Turtyo: I see

Map/RoomUI.gd
Turtyo: Did you update the connection to this signal to drop parameters ?

Map/Shaders/RoomUI.gdshader
cheesefrycook, line 1: Is this shader being used?

Map/LightOverlay.gd
Tysterman74: Ah gotcha, that parameter for TAU is the end angle that we want to finish drawing the arc. TAU is a constant in Godot that basically gives the fully circumference of a circle. 20 is just an arbitrary number to draw the arc out. The more the points, the smoother the line. I figured I just keep it a small-ish number to not overload on the amount of points drawn.

Map/RoomUI.gd
Tysterman74: Ah I didn't update the signal bus itself, the connect did connect to the function with the new parameter. I will update SignalBus

Map/Shaders/RoomUI.gdshader
Tysterman74: Ah good catch, we aren't using this right now. I'll remove.

Map/LightData.gd
```diff
@@ -0,0 +1,19 @@
...
+	has_torch = _has_torch
+	light_level = _light_level
+
+func increase_light_by_torch() -> void:
+	if light_level == Enums.LightLevel.LIT:
+		light_level = Enums.LightLevel.BRIGHTLY_LIT
+	else:
+		light_level = Enums.LightLevel.LIT
+	
+func increase_light_by_room_movement() -> void:
```
Turtyo, line 17: `increase_light_by_player_movement` maybe ? since the rooms don't move

Map/RoomBase.gd
```diff
@@ -4,15 +4,15 @@ class_name RoomBase
 ##
 ## Holds basic functionality for a room on the map
 
-@export var light_level: Enums.LightLevel:
+@export var light_data: LightData:
 	set(value):
-		light_level = value
-@export var has_torch: bool = false
+		light_data = value
```
Turtyo, line 9: it's a bit strange to have such a setter when we don't have a setter in LightData

Tests/test_light_data.gd
```diff
@@ -0,0 +1,48 @@
+extends GutTest
+var light_data: LightData
+
+func before_each():
```
Turtyo, line 4: you are missing void return type on every function

Tests/test_map_ui.gd
```diff
@@ -0,0 +1,69 @@
+extends GutTest
```
Turtyo, line 1: maybe extend `TestMapBase` ?

Tests/test_map_ui.gd
```diff
@@ -0,0 +1,69 @@
...
+	assert_eq(map_ui.room_ui_array[0][2].room.light_data.light_level, Enums.LightLevel.DIMLY_LIT)
+
+func test_player_placement():
+	PlayerManager.player_position = Vector2i(2, 0)
+	map_ui._on_room_clicked(map_ui.room_ui_array[0][2])
+	get_tree().root.add_child(map_ui)
+	
+	assert_eq(map_ui.room_ui_array[1][1].room.light_data.light_level, Enums.LightLevel.DIMLY_LIT, "Adjacent room should be dimly lit")
+	assert_eq(map_ui.room_ui_array[1][2].room.light_data.light_level, Enums.LightLevel.DIMLY_LIT, "Adjacent room should be dimly lit")
+	assert_eq(map_ui.room_ui_array[1][3].room.light_data.light_level, Enums.LightLevel.DIMLY_LIT, "Adjacent room should be dimly lit")
```
Turtyo, line 35: a for loop on the expected positions would avoid line repetition

Tests/test_map_ui.gd
```diff
@@ -0,0 +1,69 @@
...
+	assert_eq(map_ui.current_player_room.room.room_position, Vector2i(2, 0), "Player is in the wrong position")
+	
+	map_ui._add_torch_to_current_location()
+	
+	var rooms_in_range: Array[RoomBase] = MapMovement.get_all_accessible_rooms_in_range(PlayerManager.player_position, map_ui._LIGHT_FLOOR_RANGE, MapManager.current_map.rooms)
+	assert_true(map_ui.room_ui_array[0][2].room.light_data.has_torch, "Current selected room should have a torch")
+	for room: RoomBase in rooms_in_range:
+		assert_eq(room.light_data.light_level, Enums.LightLevel.LIT, str("Room position ", room.room_position.y, ",", room.room_position.x, " should be lit"))
+
+func test_torch_placement_then_movement():
```
Turtyo, line 50: `test_placing_two_torches` maybe ? i was expecting this test to check that rooms  don't become dimly lit when they are lit and there is player next to them, but that's not really what you do

Turtyo:
You should do Project -> Project Settings, toggle advanced parameters, Debug -> GDScript and use warn on untyped declarations
When we merge with main, we will need everything to be typed, so it needs to be done at one point.
Also I notice in `LightOverlay` that you declare an array but you never assign it a `[]` value, then you append on to it. I'm not exactly sure how that works, because you are appending on a `null` basically.

Having tested the function in the game, it seems to work nicely !

Tests/test_map_ui.gd
Tysterman74: I can add a check for that, but my thought process was that this is for checking what it SHOULD be changed to once two torches are down. In this case LIT or BRIGHTLY_LIT depending on where we're at. In any case, I'll update the function name.

Map/RoomBase.gd
Tysterman74: Good catch, forgot to clean it up once I changed it to light_data

Tests/test_map_ui.gd
```diff
@@ -1,22 +1,22 @@
-extends GutTest
+extends TestMapBase
 
 var map_ui_scene: PackedScene = load("res://#Scenes/MapUI.tscn")
 var map_ui: MapUI
 var test_map_array: Array[int] = [1, 3, 5, 3, 1]
```
Turtyo, line 5: since you extend from `TestMapBase`, you don't need to declare most of those variables anymore (check `TestMapBase`)

Tests/test_map_ui.gd
```diff
@@ -1,22 +1,22 @@
...
 
-func before_all():
+func before_all() -> void:
```
Turtyo, line 7: the `before_all` is already defined in `TestMapBase`
You just need to do an override to add `map_ui = map_ui_scene.instantiate` since this is not in the base file

edit well technically we recreate the map each_time with a before each, which prevents adverse effects from previous tests

Turtyo (approved):
Just a bit of cleanup left to do in the tests, but it's good for me apart from that

Tests/test_map_ui.gd
Tysterman74: I ended up removing out before_all() and put everything in before_each(), thanks for the suggestion

sethburch (approved):

cheesefrycook (approved):

Turtyo (approved):
Great work on the light, thanks @Tysterman74

Issue #56, opened 2024-01-30
Energy system


Turtyo:
## Feature Description

The energy system is the cost associated to playing cards.

## Motivation

The player has a limited amount of energy and each card costs some. This naturally limits the amount of cards that can be played in a turn, which is usually a good thing to not have players be gods on each run :laughing: 

## Concept art / Designs

The chosen design is to have players start the combat with less energy than the max (3/4 for example). Each turn, they gain 3 energy, with a total energy capped at 4. Energy carries between turns, but since it's a 3 per turn and 4 max, you can technically only carry 1 energy to the next turn. 
Of course, the numbers are given as example we should be able to edit them if we want. Note that by generating energy, the player can temporarily have more energy than the max during a turn (but the energy will go back to the max at the start of the next turn if they have more than the max).

## Acceptance Criteria

- The generation and max energy works as intended
- Each card now has an associated cost
- We should have a way to ignore the energy cost for testing purpose (if we want to play a lot of cards in a turn as devs)

## Proposed Solution

Make a class for the energy. There we define the useful functions such as calculate the energy of the player at the start of the turn, remove energy from the player when a card is played, add energy etc...
Each card now has a cost associated (should it be an instance of the energy class or a simple int ?)
The player also now has an energy attribute. They can't play cards if their energy is lower than the cost of the card (they should probably not be able to even put the card in targeting position for the cards that need a target, so it's not confusing).
The logic of a card now should include a call to reduce the energy of the player given the cost of the card.

For the devs, a way to be able to play a lot of cards in a single turn would probably be to just give a lot of energy (so the system is still running, just that if we have 1000 energy we won't run out). That might be done with a parameter in a global file ? Sure the player could go and cheat to give themselves unlimited energy too, but the code is public anyway so if they want to, then they can.

## Additional Context

Should this be broken down in a few smaller PR ? Maybe a first one for the energy system in itself with the functions and adding that to the player, then the implementation for the cards ? It's not impossible to do everything in a single PR, but it would be big.

cheesefrycook:
I think we should add a new `EnergyComponent` similar to `HealthComponent` that is only on the player entity. Then of course add an accessor function on the Player class like `get_energy_component`. For the energy itself, I feel like we can do a simple int. I'm not sure if we will need any more information in that regard. A single PR for both cards and player is fine I think.

[...]

PR #71, opened 2024-02-14
Energy system

Paper-2:
# Description

Adds a new system that adds a cost to cards.

## Related issue(s)

https://github.com/Saplings-Projects/1M_sub/issues/56

## List of changes

- The current amount of energy is displayed on the testing scene
- Cards have a cost tied to them
- A card can only be selected if the current energy meets its cost
- A new effect to add more energy
- A var to allow cheating for testing purposes

## Tests

Energy generation

Having no energy

Having energy

Adding energy

Using energy


Paper-2:
Test are failing. Reason being Player does not exist in PlayerManager. It could be solved by either not using PlayerManager to use the component or adding a player to PlayerManager in the tests.

Entity/Components/EnergyComponent.gd
```diff
@@ -0,0 +1,38 @@
+extends EntityComponent
+class_name EnergyComponent
+
+@export var max_energy: int = 4
```
Turtyo, line 4: capitalize this as it won't really change usually

Entity/Components/EnergyComponent.gd
```diff
@@ -0,0 +1,38 @@
...
+	if card != null:
+		return energy >= card.card_data.energy_cost
+
+	return true
+
+
+func use_energy(card: CardBase) -> void:
+	energy -= card.energy_cost
+
+	if energy < 0:
```
Turtyo, line 25: this might hide a problem (energy shouldn't be negative)
instead of setting the energy back to 0, we should assert the energy is not lower than 0 instead

Entity/Components/EnergyComponent.gd
```diff
@@ -0,0 +1,38 @@
...
+
+	on_energy_changed.emit(energy)
+
+
+func on_turn_end() -> void:
+	if energy >= max_energy:
+		energy = max_energy
+		
+	else:
+		energy += energy_generation
```
Turtyo, line 36: this should be the other way
add the energy generation and then set it back to the max if it's too big
because right now, if i have 3 energy i will start my next turn with 6 (since 3 is not the max of 4)

Tests/TestBase.gd
```diff
@@ -45,7 +48,8 @@ func before_each():
 	_player_stat_component.get_stats().ready_entity_stats()
 	_enemy_stat_component.get_stats().ready_entity_stats()
 	_enemy_2_stat_component.get_stats().ready_entity_stats()
-	
+
+	_player_energy_component.ignore_cost = true
```
Turtyo, line 52: why ? this is a test, you shouldn't ignore card cost

Entity/Player/Player.gd
```diff
@@ -6,7 +6,12 @@ class_name Player
...
 	PlayerManager.set_player(self)
+	get_energy_component().init_entity_component(self)
 
 
 func _exit_tree() -> void:
 	PlayerManager.set_player(null)
+
+#this PR is going to be so easy
+func get_energy_component() -> EnergyComponent:
+	return Helpers.get_first_child_node_of_type(self, EnergyComponent) as EnergyComponent
```
Turtyo, line 17: add a line at the end of your file please

UI/EnergyLabel.gd
```diff
@@ -0,0 +1,30 @@
+extends Label
```
Turtyo, line 1: i'm not familiar with label, what does it do ?

Entity/Player/Player.gd
```diff
@@ -6,7 +6,12 @@ class_name Player
 func _ready() -> void:
 	super()
 	PlayerManager.set_player(self)
+	get_energy_component().init_entity_component(self)
```
Turtyo, line 9: is this needed ? I know that for health we don't need to do it since it's not in the code anywhere, why do you need to do it for the energy component ?

Turtyo (changes requested):

Tests/TestBase.gd
Paper-2: Now that I think about it, its not doing anything right now. Players first need to queue the card and then play the card.
I'm guessing the cards that are used in the tests skip queuing the cards.

Entity/Player/Player.gd
Turtyo: if it's redundant it should be removed

Tests/TestBase.gd
Turtyo: you don't interact physically with the cards indeed, so they are not queued

Turtyo: ah the idea was the allow other tests to continue working without having to worry about energy cost ?

Paper-2: Yes, those tests are testing for other stuff and not the energy system directly.

Turtyo: yes so it's logical to add this line indeed, just need to overwrite that in the init of the energy tests

Cards/EnergyData.gd
```diff
@@ -0,0 +1,4 @@
+extends Resource
+class_name EnergyData
+
+@export var energy_cost: int = 0
```
Turtyo, line 4: add a line at the end of the file please

Entity/Components/EnergyComponent.gd
```diff
@@ -1,7 +1,7 @@
 extends EntityComponent
 class_name EnergyComponent
 
-@export var max_energy: int = 4
+@export var Max_energy: int = 4
```
Turtyo, line 4: by capitalize I meant everything
`MAX_ENERGY`

Entity/Components/EnergyComponent.gd
```diff
@@ -23,20 +23,20 @@ func use_energy(card: CardBase) -> void:
 	energy -= card.energy_cost
 
 	if energy < 0:
-		energy = 0
+		push_error("energy cannot be negative")
 
 	on_energy_changed.emit()
 
 
 func on_turn_end() -> void:
```
Turtyo, line 32: this should probably be a `on_turn_start`
reason for that all the other stuff happens at the start of turns. Also, we might have an effect that reduces energy at the start of turns, so it's more logical to put the logic of energy generation at the start
Another problem with that is that since earlier you wrote `starting_energy = 0` and the energy is generated at the end of turns, it means the player has 0 energy for the first turn
Edit : I see start energy was changed to 3, but this doesn't change the idea of putting everything at the start of the turn and start with 0 energy

TL;DR : make this a `on_turn_start`

Entity/Components/EnergyComponent.gd
```diff
@@ -23,20 +23,20 @@ func use_energy(card: CardBase) -> void:
...
+	if energy >= Max_energy:
+		energy = Max_energy
+
 	on_energy_changed.emit()
 
 func add_energy(value: int) -> void:
 	energy += value
-	print(energy)
\ No newline at end of file
+	print(energy)
```
Turtyo, line 43: is there a reason for printing the energy ?

Tests/test_energy.gd
```diff
@@ -0,0 +1,60 @@
...
+func test_add_energy():
+	card = load("res://Cards/Resource/Card_Energy.tres") # adds 70 energy
+
+	card.on_card_play(_player, [_player])
+	assert_eq(_player_energy_component.energy, 70)
+
+
+func test_using_energy():
+	_player_energy_component.add_energy(10)
+	card = load("res://Cards/Resource/Card_Damage.tres") # costs 7 by default
```
Turtyo, line 22: does it cost 7 by default or do you want to change it to 7 by default in this test ?
if it's 7 by default, it's too high and it needs to be realistic, you change it to 7 only in this test.

Tests/test_energy.gd
```diff
@@ -0,0 +1,60 @@
...
+
+	card.on_card_play(PlayerManager.player, [_enemy])
+
+	assert_eq(_player_energy_component.energy, 3)
+
+func test_no_energy():
+
+	_card_container.draw_cards(1)
+
+	var _queued_card: CardWorld = _card_container.cards_in_hand[0]
```
Turtyo, line 33: name it `card_to_queue` because this name makes it feel like it's already being queued

Tests/test_energy.gd
```diff
@@ -0,0 +1,60 @@
...
+func test_using_energy():
+	_player_energy_component.add_energy(10)
+	card = load("res://Cards/Resource/Card_Damage.tres") # costs 7 by default
+	card.energy.energy_cost = 7
+
+	card.on_card_play(PlayerManager.player, [_enemy])
+
+	assert_eq(_player_energy_component.energy, 3)
+
+func test_no_energy():
```
Turtyo, line 29: `test_not_enough_energy` would probably be a lot better
missing void return type for all test functions btw

Tests/test_energy.gd
```diff
@@ -0,0 +1,60 @@
...
+	_card_container.draw_cards(1)
+
+	var _queued_card: CardWorld = _card_container.cards_in_hand[0]
+	_queued_card.card_data.energy.energy_cost = 999
+
+	_card_container._on_card_clicked(_queued_card)
+
+	assert_false(_card_container.is_card_queued())
+
+func test_having_energy():
```
Turtyo, line 40: `test_play_card_with_enough_energy` ? to have energy doesn't mean you have enough

Tests/test_energy.gd
```diff
@@ -0,0 +1,60 @@
...
+
+	_player_energy_component.add_energy(10)
```
Turtyo, line 42: I notice that you always use energy amounts that the player won't usually go to
it would be good to have some functions that test for energy levels that would happen more usually

Tests/test_energy.gd
```diff
@@ -0,0 +1,60 @@
...
+func test_no_energy():
+
+	_card_container.draw_cards(1)
+
+	var _queued_card: CardWorld = _card_container.cards_in_hand[0]
+	_queued_card.card_data.energy.energy_cost = 999
+
+	_card_container._on_card_clicked(_queued_card)
+
+	assert_false(_card_container.is_card_queued())
```
Turtyo, line 38: could also assert that the energy didn't change

Turtyo (changes requested):
I think the tests could cover more cases.
For example, testing when you have just enough energy for a card
testing if the energy you have is just one less than the energy of the card

Cards/CardBase.gd
```diff
@@ -20,6 +20,8 @@ class_name CardBase
 
 @export var card_effects_data: Array[EffectData] = []
 
+@export var energy: EnergyData = EnergyData.new()
```
Turtyo, line 23: I would maybe name this something else than `energy`
because `energy` alone feels like this would be a cost (which is in fact `energy_cost`)
Maybe `energy_info` would be more fitting (so we know it's information and we need to query the actual things in it)

#Scenes/TestingScene.tscn
cheesefrycook, line 1: We should probably switch the energy UI display image to a TextureRect instead of a Sprite2D

Cards/CardBase.gd
```diff
@@ -46,6 +48,9 @@ func on_card_play(caster: Entity, targets: Array[Entity]) -> void:
 	#Split up targeted attacks and all attacks
 	var effects_targeting_single: Array[EffectData] = get_effects_targeting_single()
 	var effects_targeting_all: Array[EffectData] = get_effects_target_all()
+
+	#uses the energy asocieted with the card
+	PlayerManager.player.get_energy_component().use_energy(self)
```
cheesefrycook, line 53: Maybe add a `if caster is Player` before this call, since enemies use `on_card_play` too

Entity/Components/EnergyComponent.gd
```diff
@@ -0,0 +1,43 @@
...
+
+	energy += energy_generation
+
+	if energy >= Max_energy:
+		energy = Max_energy
+
+	on_energy_changed.emit()
+
+func add_energy(value: int) -> void:
+	energy += value
```
cheesefrycook, line 42: So if you add energy from a card then it goes above the `max_energy`? Is that by design?

Entity/Components/EnergyComponent.gd
cheesefrycook: There's already a place to do this too. CardContainer -> _on_player_start_turn

Entity/Components/EnergyComponent.gd
Paper-2: I forgot to delete the print.

Entity/Components/EnergyComponent.gd
Paper-2: The energy only returns to the maximum when the turn ends (or starts. after I change it)

Entity/Components/EnergyComponent.gd
Turtyo: yes but your function is `on_turn_end`
do you plan to call `on_turn_end` inside `_on_player_turn_start` ?

Cards/CardBase.gd
Turtyo: good catch

Entity/Components/EnergyComponent.gd
Turtyo: It's indeed by design. You can check the explanation on the related issue #56

Cards/Resource/Card_DrawCards.tres
```diff
@@ -14,14 +13,9 @@ effect = SubResource("Resource_2ap55")
 value = 2
 target_type = 1
 
-[sub_resource type="Resource" id="Resource_j6isw"]
-script = ExtResource("3_06oqv")
-energy_cost = 1
```
Turtyo: does that mean the DrawCard doesn't cost anything ? wouldn't it be better to have the energy cost of 0 to not try to get the energy info and getting null instead ?

Cards/Resource/Card_Energy.tres
```diff
@@ -14,14 +14,14 @@ effect = SubResource("Resource_jur0i")
 value = 70
 target_type = 0
 
-[sub_resource type="Resource" id="Resource_1100t"]
-script = ExtResource("3_2mfb5")
-energy_cost = 0
```
Turtyo: so the card that gives you energy costs 2 energy ?

Tests/test_energy.gd
```diff
@@ -8,42 +8,67 @@ func before_each():
...
+	card.energy_info.energy_cost = 0
 	card.on_card_play(_player, [_player])
 	assert_eq(_player_energy_component.energy, 70)
 
-
 func test_using_energy():
 	_player_energy_component.add_energy(10)
-	card = load("res://Cards/Resource/Card_Damage.tres") # costs 7 by default
-	card.energy.energy_cost = 7
+	card = load("res://Cards/Resource/Card_Damage.tres")
```
Turtyo, line 21: missing type of the card (CardBase)

Turtyo (commented):
Good changes, I have a few more questions

Tests/test_energy.gd
Paper-2: Line 3

Cards/Resource/Card_DrawCards.tres
Paper-2: if the value isn't set then it will use the one set in cardbase. (when I renamed energy to energy info I had to reset the values for every card I may or may not have skipped it)

Turtyo: I see, default is 0

Cards/Resource/Card_Energy.tres
Paper-2: Can you rephrase this?

Turtyo: Nvm I saw the card, it gives you 70 energy for a cost of 2
I was just surprised that the card that adds to your energy costs energy, but since it gives a lot it doesn't matter
this will be changed later to a more reasonable value

Cards/Resource/Card_Energy.tres
```diff
@@ -0,0 +1,27 @@
...
+[ext_resource type="Script" path="res://Cards/CardBase.gd" id="3_opvsg"]
+
+[sub_resource type="Resource" id="Resource_jur0i"]
+script = ExtResource("1_3yh17")
+
+[sub_resource type="Resource" id="Resource_6gvxm"]
+script = ExtResource("2_dr7f0")
+effect = SubResource("Resource_jur0i")
+value = 70
+target_type = 0
```
Turtyo, line 15: this is not insta-cast
things will have to change with the targeting PR anyway, i will rewrite this card once everything is merged I think

Paper-2: Just for reference, what is insta-cast?

Turtyo: Insta-cast is the mechanic that makes it so you don't need to target someone when it's clear who the card is for (buff, heal, damage everyone etc...)
This changed due to the targeting PR that was just merged, ask me if you don't understand the new system

UI/EnergyLabel.gd
```diff
@@ -0,0 +1,20 @@
...
+
+var energy_component: EnergyComponent
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
```
Turtyo, line 6: All the functions in this file are missing a return type (which causes errors when loading the game, since missing typing is treated as an error now)

UI/EnergyLabel.gd
```diff
@@ -0,0 +1,20 @@
...
+
+# Called when the node enters the scene tree for the first time.
+func _ready() -> void:
+	energy_component = PlayerManager.player.get_energy_component()
+	energy_component.on_energy_changed.connect(on_energy_changed)
+	CardManager.on_card_action_finished.connect(on_card_played)
+	update_label()
+
+
+func on_card_played(_card: CardWorld) -> void:
```
Turtyo, line 13: I was also notice that you have the function with a card parameter that is not used
I understand it is because the `on_card_action_finished` returns card. Instead of having a parameter that you don't use, I suggest doing:
`CardManager.on_card_action_finished.connect(on_card_played.unbind(1))` which will basically ignore the value that is given by `on_card_action_finished`
Because right now I was wondering if `on_card_player` was an override or something like that

Turtyo (approved):

cheesefrycook (approved):

PR #72, opened 2024-02-15
Movement on the map

### LINKED ISSUE #67: Movement on the map
## Feature Description

Be able to move on the map by clicking the display on the map.

## Motivation

Now that we have a map display to go with our map structure, we need the player to be able to move between the rooms in order to have a progression in the game

## Concept art / Designs

I'm thinking of having the room have a border that lights up a bit when you hover it to make it clear where you are going.

## Acceptance Criteria

- Player can only move to the room directly above, up-left or up-right (they can't jump a floor, they can't move in the same floor either)
- A little icon to show in which room the player is currently

## Proposed Solution

I'm thinking that add buttons in the map display. For now I think we are just displaying an image for each room (you confirm @Tysterman74 ?), so in the loop that's building the position of each image, we also put a button with the same dimensions as the room.

We have two possibilities:
- either we had a button to every room on the map when we display it, and we check when a player clicks if they can access the room from where they are, if so we move them there
- when building the layout of the buttons on top of the map display, we only add the button if the player can access the room

I like option 2 more because it will use less resources and is also probably less error-prone. One point we need to be careful of is to move the buttons when the map is moved by scrolling

We can have a hover state for the map display to highlight the rooms we have the mouse over.

Maybe an even better solution is to directly have room in the map display be a button, so it will always scroll with the map without problem, but i don't know how much we can custom the button to display images.

Turtyo:
# Description

Allow the player to move by clicking on the map

## Related issue(s)

Closes #67
Related to the display of the map and #9 (done at the same time in parallel)

## List of changes

- Add possible movements as vector
- Function to know accessible places
- Replace room texture on the map display with buttons with a texture instead

## Tests

- Test the room positions accessible directly next to a given position
- Same but with a given range instead

I'm not exactly sure of what more to write, because writing a test on the UI seems pretty difficult for the map

## Additional notes

Opened as a draft for now to help @Tysterman74 by providing the function to get the accessible rooms needed in PR #70 to light the rooms


Map/MapMovement.gd
```diff
@@ -0,0 +1,47 @@
+class_name MapMovement extends Resource
+
+
+static func get_accessible_room_positions_by_player(position: Vector2i) -> Array[Vector2i]:
```
Tysterman74, line 4: This function doesn't explictly get the positions based on the player since we pass in the position in the arguments, so I think a better name would be 
`get_accessible_room_positions_by_given_position()`

Map/MapMovement.gd
```diff
@@ -0,0 +1,47 @@
...
+	var accessible_positions_in_range: Array[Vector2i] = get_all_accessible_room_positions_in_range(position, remaining_range)
+	var accessible_rooms_in_range: Array[Variant] = []
+	for current_position: Vector2i in accessible_positions_in_range:
+		accessible_rooms_in_range.append(current_map[current_position.y][current_position.x])
+	
+	return accessible_rooms_in_range
+
+
+# Helper function to be able to call without any parameters
+static func get_accessible_rooms_by_player(position: Vector2i = PlayerManager.current_position, remaining_range: int = 1, current_map: Array[Array] = MapManager.current_map.rooms) -> Array[RoomBase]:
```
Tysterman74, line 46: I'm wondering if having a custom data class to hold the floor_index and room_index can help with readability here. Something like:
```
class_name RoomData

var floor_index: int
var room_index: int

func _init(new_floor_index: int, new_room_index: int):
	floor_index = new_floor_index
	room_index = new_floor_index
```

Then we can instantiate it like so and assign it.
`var roomData: RoomData = RoomData.new(0, 0)`

Tysterman74 (changes requested):

Map/MapMovement.gd
Turtyo: That just feels like a fancy Vector2i
I was thinking something like that, but I think it would be more confusing if we go with the implementation you gave
Reason being, when the player start, we say they are at floor 0 or 1, depending on what we choose
But the matrix is in the opposite way, meaning that when the player start they are the "floor" 4 on a map of size 5
so when you move up the floor goes down
And even if the vector is confusing, at least you can keep that in mind, when it might become more confusing if we use the RoomData
Or it needs a special structure where it returns the size of the map minus the floor to do a symetry

Map/MapMovement.gd
Turtyo: yes, will do

Managers/PlayerManager.gd
```diff
@@ -4,13 +4,42 @@ extends Node
...
+
+func _get_starting_position() -> Vector2i:
+	var current_map_width_array: Array[int] = MapManager.map_width_array
+	var starting_position: Vector2i = Vector2i(0,0)
+	
+	var padding_size_first_floor: int = (current_map_width_array.max() - current_map_width_array[0])/2
+	
+	starting_position.x = padding_size_first_floor
+	
+	return starting_position
```
Tysterman74, line 36: If I understood this correctly, this returns back the starting position as the mid point of the first floor. I believe design team mentioned that there can be a first floor where we can select where we can start, should we address that in this Issue or make a new one for it?

Tysterman74 (changes requested):
Two small comments! Lookin' good. The MapMovement functions are gonna be extremely helpful for the Light Mechanic. I'm going to continue ironing out the small comments until this PR is merged then I'll sync your changes in with the light mechanic.

Map/MapMovement.gd
```diff
@@ -0,0 +1,64 @@
+class_name MapMovement extends Resource
+
+
+static func get_accessible_room_positions_by_given_position(position: Vector2i) -> Array[Vector2i]:
```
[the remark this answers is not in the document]
Turtyo: will do

Managers/PlayerManager.gd
Turtyo: This returns the first room of the floor (so the most left room)
Indeed player could select their room at start. I thought of having a special state were you can have access to all the rooms of the first floor.
I can do that with a boolean like `selecting_first_floor`, with an associated room selection function that returns all the rooms of the first floor

Tysterman74: Ok, just wanted to make sure we have that in mind. I'll resolve this for now, but if we add this new state in this PR, just go ahead and reopen this comment and I can take another look.

Turtyo: i'll do it now, it's not too hard to do and it makes sense to do it now

Tysterman74 (approved):

Tests/TestMapBase.gd
```diff
@@ -0,0 +1,20 @@
...
+
+var test_width: Array[int] = [1,3,5,3,1]
+
+var test_map: MapBase = null
+
+func before_each() -> void:
+	map_manager = test_generator.new()
+	MapManager.map_width_array = test_width 
+	# need to change the manager since it's globally loaded
+	test_map = map_manager.create_map(test_width)
```
Paper-2, line 14: Why is `test_width` being passed to `MapManager `and then used as an argument in `map_manager.create_map` ?

Global/SignalBus.gd
```diff
@@ -0,0 +1,6 @@
+extends Node2D
+
+### This is meant for having signals that can come from multiple instance of the same class
+### Only add signals in this file
+
+signal clicked_next_room_on_map
```
Paper-2, line 6: no line at the end.

Tests/TestMapBase.gd
Turtyo: why not ?
if you look at MapManager, it has a `map_width_array` argument, so I give it the test width
then I need the map so use the `ŧest_width` again
I'm basically doing the same thing as the `_ready`

Helpers/StaticHelpers.gd
```diff
@@ -32,3 +32,11 @@ static func convert_from_range(value: float, from_min: float, from_max: float, t
 		to_value = (((value - from_min) * to_range) / from_range) + to_min
 	
 	return to_value
+
+
+# ! works in place
+static func remove_duplicate_in_array(array: Array) -> void:
+	var dict: Dictionary = {}
+	for value: Variant in array:
+		dict[value] = 1 # value doesn't matter, we just need the keys
```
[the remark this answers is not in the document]
Turtyo: I prefer using an int than null
As I said, it doesn't matter, and yes I'm using it as a set

Tests/TestMapBase.gd
Paper-2: What I'm saying is that `create_map `has a redundant argument.
`map_width_array `already stores what the argument, `map_floors_width`, is passing to the function.
I can understand it if it serves the purpose of making `create_map ` just use local variables

Turtyo: Here is the code of `create_map`
```GDScript
func create_map(map_floors_width: Array[int]) -> MapBase: ## Generates and Populates a map with rooms that have random room types. More in depth algorithms will be added in the future
	var _map: MapBase = MapBase.new()
	var _grid: Array[Array] = [] ## 2d array to return. this will be populated with rooms
	var _max_floor_size: int = map_floors_width.max()
	for index_height: int in range(map_floors_width.size()): ## loop through the floor and append null to create a square grid
		
		var _floor_width : int = map_floors_width[index_height]
		var _padding_size : int = (_max_floor_size - _floor_width)/2
		
		var _padding:Array = []
		_padding.resize(_padding_size)
		_padding.fill(null)
		_grid.append([])
		_grid[index_height].append_array(_padding)
		
		for index_width: int in range(map_floors_width[index_height]):## loop through elements of the grid and assign a room type
			var _rand_type_index: int = randi_range(0, GlobalVar.EVENTS_CLASSIFICATION.size() - 1)
			var _room_event: EventBase = GlobalVar.EVENTS_CLASSIFICATION[_rand_type_index].new()
			var _generated_room: RoomBase = RoomBase.new()
			_generated_room.room_event = _room_event
			_generated_room.room_position = Vector2i(index_width + _padding_size, index_height)
			_grid[index_height].append(_generated_room as RoomBase)
		_grid[index_height].append_array(_padding)
	_map.rooms = _grid
	return _map as MapBase
```

`create_map` doesn't have redundant arguments, you *need* to give it an array of width
I could write that instead:
```GDScript
MapManager.map_width_array = test_width 
# need to change the manager since it's globally loaded
test_map = map_manager.create_map(MapManager.map_width_array)
```
But it's longer to write and does exactly the same thing

Tysterman74 (approved):

PR #83, opened 2024-03-02
 Move all art images into a dedicated folder

### LINKED ISSUE #57: Move all art images into a dedicated folder

The `Card` folder is starting to get a bit disorganized because of the images for the cards being in the Card folder

Move all the images inside a `card_images` folder inside the `Card` directory

## Additional context
I was about to do that on my current targeting PR but I thought it might be better as a small PR of its own because people that merge this small PR will have to tinker with their imports and .tscn files that reference those images.

## Edit : proposed solution after discussion
```
Root/
    | - - Art/
	  | - - Background
	  | - - Cards
	  | - - Characters
	  | - - Enemies
	  | - - Items
	  | - - Menus
	  | - - NPC
```
All arts within an art folder

With naming convention:
- all names start with art-<something> with the something being:
    - background : `bg-<area_number>-<number>`
    - NPC :` npc-<name>` ; name should be written using underscore if there are spaces
    - Mob :`enemy-<name>-<pose>` with pose being the name of a particular pose the enemy can have (defeated, full life, took damage, etc...)
    - Items : `item-<type>-<name>` with type being either accessory or consumable
    - Cards : `card-<name>`
    - Character : `player-<pose>`
    - Menus : `menu-<name>-<number>`

Tests should be added to ensure that the naming convention is respected.

Turtyo:
# Description

Move all files into an Art folder and make a naming convention

## Related issue(s)

Closes #57 

## List of changes

- Move all images in the `Art` directory
- Make sub-directories for all those files
- Add a `naming_convention.txt` so that files are correctly named

## Tests

Added tests to check for correct naming convention. This does a parse of `naming_convention.txt` so that the tests are always up to date with the actual naming convention, without needing to rewrite the tests.
This also means the area that has the naming convention in the txt file must be properly written for it to work with tests.

## Additional notes

Hoping to merge this quickly to prevent merge conflicts when other PR come in.


JonaLam (commented):
Maybe you could move some of the UI element from the "Card" folder into it's own folder you could call "Battle UI" or just "Battle". This could make it a bit easier to find, but I could understand that the clutter it would make would be more annoying than the slight ease to find.

JonaLam:
I would suggest making a "Card Art" folder, as that folder will probaby get pretty big as the prodject gets bigger

JonaLam:
If you wish to reduce the size of each folder layer, you could put the "NPC", and "Enemy" folders into the "character" folder, and then maybe creat a new "Player" folder for those assets

Tysterman74 (dismissed):

Turtyo:
@JonaLam 
> I would suggest making a "Card Art" folder,

This is the goal of the `Cards` folder already, to put everything related to cards and decks etc... If you want to differentiate between the art of the card, and the layout of the card in different folders, that's doable. Just tell me, it won't take me long to make.

> If you wish to reduce the size of each folder layer, you could put the "NPC", and "Enemy" folders into the "character" folder

I think it's fine as is. The Enemy folder in particular will get big, so keeping it separate is probably a good thing.

JonaLam:
@Turtyo 
Yeah, I did mean that you should differentiate the art on the card, as I think there wil be a lot of those.

The other part, what I meant was put the "Enemy" and "Npc" folders inside the "character" folder, I feel like there are a lot of folders and it could be nice to further compress the, but it's fine either way

Turtyo:
I prefer having everything in the same level as it will make path length rather consistent.
Also it would complexify the test of naming convention, as I would have to recurse through folders.

I will make a folder for card interface and one for the embedded image on the cards

Turtyo:
like that @JonaLam ?
I'm not too sure about the names, layout seemed ok-ish but maybe you have a better idea ?

JonaLam (approved):
Yeah, that's good

Tysterman74 (approved):

PR #84, opened 2024-03-13
Card casting animations

### LINKED ISSUE #38: Card casting animations

cheesefrycook:
# Description

Adds an animation system to cards. Creates a few test animations.

## Related issue(s)

Closes #38 

## List of changes

- Adds 3 test animations (Poison, Slap, Fauna Sweep)
- Adds CastPosition Resource which defines how an animation is created and placed in the world
- Modifies CardContainer to block card interaction while effect is active

## Tests

No tests written. If anyone has any ideas for tests let me know.

## Additional notes

I added two CastPosition types: CastPos_AllTargets and CastPos_CentralPosition
- CastPos_AllTargets - creates a new instance of the animation scene for each target
- CastPos_CentralPosition - creates one instance of the animation scene and places it at a central position between all targets

These are the only two I added for now. We might want more in the future.

I considered having a delay between each animation for the CastPos_AllTargets so the animation would have a "staggered" type effect. However I found that this complicated the animation system a lot, so I decided to leave it out for now.

Each animation scene is required to have an AnimationPlayer node. This node allows you to create the actual animation, as well as call the function that triggers the hit. For the slap animation, it looks like this:

image

The `trigger_cast_hit` applies the CardEffect. If no `trigger_cast_hit` is called during the animation, then it will be triggered when the animation completes.

https://github.com/Saplings-Projects/1M_sub/assets/151570126/bb872d0e-c350-40ca-b510-34c926ecdbd5


Cards/Animation/CastAnimation.gd
```diff
@@ -0,0 +1,40 @@
...
+func trigger_cast_hit() -> void:
+	if _was_animation_hit_triggered:
+		push_error("Tried to trigger multiple hits in CastAnimation " + get_name())
+		return
+	
+	on_animation_hit_triggered.emit()
+	_was_animation_hit_triggered = true
+
+
+func _finish_casting(_anim_name: StringName) -> void:
```
Tysterman74, line 34: Will `_anim_name` be used eventually? Looks like we're not using it within this function.

Cards/CardBase.gd
```diff
@@ -42,8 +45,46 @@ func on_card_play(caster: Entity, base_target: Entity) -> void:
...
+
+
+func _handle_effects_queue(caster: Entity, base_target: Entity) -> void:
+	var card_effect: EffectData = _card_effects_queue[0]
+	var animation_data: CastAnimationData = card_effect.animation_data
+	var list_targets: Array[Entity] = card_effect.targeting_function.generate_target_list(base_target)
+	
+	var can_use_animation: bool = animation_data != null and\
+		animation_data.cast_animation_scene != null and\
+		animation_data.cast_position != null and\
```
Tysterman74, line 63: I think we can encapsulate these animation data checks within the CastAnimationData itself, makes it a bit easier if we're using this in other places so we're not copypasta-ing the logic.

Tysterman74 (changes requested):
Everything looks good so far! I think with the addition of the animations, it would warrant updating some of our card unit tests to account for the animations.

Cards/Animation/CastPosition/CastPos_CentralPosition.gd
```diff
@@ -0,0 +1,30 @@
...
+## If false, creates the animation at 0,0
+@export var use_central_position: bool = true
+@export var offset: Vector2 = Vector2.ZERO
+
+
+# @Override
+func initialize_animation(cast_animation_scene: PackedScene, list_targets: Array[Entity]) -> CastAnimation:
+	var cast_animation: CastAnimation = cast_animation_scene.instantiate()
+	
+	assert(list_targets.size() > 0, "Tried to create a CastPos_CentralPosition with no targets")
```
Turtyo, line 15: why is this check made here and it is not made in CastPos_AllTargets ?
Should this be a push_error instead of an assert ? Since assert get removed when the game is compiled iirc

Cards/Animation/CastPosition/CastPos_CentralPosition.gd
```diff
@@ -0,0 +1,30 @@
...
+	
+	list_targets[0].get_tree().root.add_child(cast_animation)
+	
+	if use_central_position:
+		var target_positions: Array[Vector2] = []
+		for target: Entity in list_targets:
+			target_positions.append(target.global_position)
+		
+		cast_animation.global_position = Helpers.get_mean_vector(target_positions)
```
Turtyo, line 24: should we differentiate between player and enemies for animation casting ?

Cards/CardBase.gd
```diff
@@ -42,8 +45,46 @@ func on_card_play(caster: Entity, base_target: Entity) -> void:
 	if caster is Player:
 		PlayerManager.player.get_energy_component().use_energy(self)	
 	
-	for effect_data: EffectData in card_effects_data:
-		var list_targets: Array[Entity] = effect_data.targeting_function.generate_target_list(base_target)
-		for current_target in list_targets:
-			effect_data.apply_effect_data(caster, current_target)
-	CardManager.on_card_action_finished.emit(self)
+	if _card_effects_queue.size() > 0:
```
Turtyo, line 48: so this is basically to prevent the player from playing too many cards in a row ? Wouldn't it be better to either:
- disable cards being played when another card is played
- add animations to the queue and play them when the associated card becomes the active card ?

I think solution two would fit better with your current code

Cards/CardBase.gd
Turtyo: agreed

Cards/CardBase.gd
```diff
@@ -42,8 +45,46 @@ func on_card_play(caster: Entity, base_target: Entity) -> void:
...
+	for current_target in list_targets:
+		card_effect.apply_effect_data(caster, current_target)
+	
+	# Wait for animation to complete
+	if created_cast_animation != null:
+		await created_cast_animation.on_animation_cast_complete
+	
+	# Handle next effect in the queue or finish casting
+	if _card_effects_queue.size() > 0:
+		_handle_effects_queue(caster, base_target)
```
Turtyo, line 88: is there any specific reason for making this a recursion instead of a for loop ?

Helpers/StaticHelpers.gd
```diff
@@ -34,6 +34,14 @@ static func convert_from_range(value: float, from_min: float, from_max: float, t
 	return to_value
 
 
+static func get_mean_vector(positions: Array[Vector2]) -> Vector2:
+	if positions.size() <= 0:
+		return Vector2.ZERO
+	var mean_vector: Vector2 = Vector2.ZERO
+	for position in positions:
```
Turtyo, line 41: missing typing for position. Isn't Godot complaining since we have typing as error ?

Managers/CardManager.gd
```diff
@@ -11,6 +11,8 @@ var card_container: CardContainer = null
 var current_deck: Array[CardBase] = []
 var default_deck: CardSetBase = null
 
+var disable_card_animations: bool = false
```
Turtyo, line 14: I see this is used as a check later in handle_effect_queue, and in tests.
Should this be moved to a global file ? I think we can keep it here, but it just felt a bit strange

Turtyo (commented):
As Ty said, card tests should probably be updated. 
Is there a way to test for timing related stuff in tests ? For animation related tests, we could test that some animations get played properly given the time it takes for example.

Anyway, that's a great work ! I feel like Cheesy is always the one bringing UI improvement, which is a good thing because otherwise the game would still feel very rigid :laughing:

Cards/CardBase.gd
cheesefrycook: Well ideally this code will never get reached, because our UI doesn't allow you to play a card more than once. I just added this as a check to make sure we never hit this

Cards/CardBase.gd
cheesefrycook: Good call, i'll move it

Cards/CardBase.gd
cheesefrycook: Because we are using `await` in the above code, so this function will not run instantly. You can't use `await` inside a for loop as far as i know

Helpers/StaticHelpers.gd
cheesefrycook: I think that error doesn't work for for loops specifically. I'll fix

Managers/CardManager.gd
cheesefrycook: Well CardManager is global in a way. I think it kind of fits because it's card related

Cards/Animation/CastAnimation.gd
cheesefrycook: No i guess not, I can unbind it

Cards/Animation/CastPosition/CastPos_CentralPosition.gd
cheesefrycook: Yeah that's a good point, i haven't tried this for enemies. I will test that

Cards/Animation/CastPosition/CastPos_CentralPosition.gd
cheesefrycook: Yeah true, i can change that

Cards/Animation/CastPosition/CastPos_AllTargets.gd
```diff
@@ -0,0 +1,24 @@
+extends CastPosition
+class_name CastPos_AllTargets
+## CastPos_AllTargets creates an animation for each target
```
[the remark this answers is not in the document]
cheesefrycook: Yeah CentralPosition chooses a central point between all the targets, so an explosion should work good for this i think.

My intention with the CastPosition is it just creates the animation and sets the initial position. For movement during the animation, we can potentially override the `play_animation` in CastAnimation for a specific animation to give it custom movement. I can add that as a test so we can test that functionality?

cheesefrycook:
What exactly do we need to update about the card tests? I set the flag `disable_card_animations` to true so the animations should be disregarded for our existing tests. I can see about writing some tests specifically for animation though (like Turtyo mentioned, testing if the animation gets created after a set delay)

Cards/Animation/CastPosition/CastPos_AllTargets.gd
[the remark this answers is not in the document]
Turtyo: Yeah if we can test that ! If it doesn't work, it's fine, because we are not even sure we will do that, so let's not try too hard to solve problems we don't have. But if it works out of the box then great

Cards/CardBase.gd
Turtyo: you are right :+1:

Cards/CardBase.gd
Turtyo: I see, i'm not too familiar with using the await

Managers/CardManager.gd
Turtyo: Yeah it's a singleton, but it was more to have all our variables used globally in a same place
Choosing to keep them in their related files also works

Turtyo (changes requested):
ArtSource was deleted in the latest Art PR, the aseprite should be moved into its relevant folder I think

cheesefrycook:
I added a banana throw animation to show how we can override `CastAnimation` for specific animations. Had to do a good bit of rearranging to make it work, but I think the code is better now

https://github.com/Saplings-Projects/1M_sub/assets/151570126/d02b69e2-95b4-442b-8e60-aba259a159b0

cheesefrycook:
Also did a few changes to make animations work for enemies

https://github.com/Saplings-Projects/1M_sub/assets/151570126/1f6e94ec-de6d-4b71-9337-059f1492d4fc

Core/Battler.gd
```diff
@@ -79,21 +79,35 @@ func _on_enemy_start_turn() -> void:
...
+		_enemy_action_list.append(enemy_action)
+	
+	# start the attack queue
+	assert(_enemy_action_list.size() >= 0, "Enemy tried to attack while there were attacks queued!")
+	_handle_enemy_attack_queue()
+
+
+func _handle_enemy_attack_queue() -> void:
+	var enemy_action: EnemyAction = _enemy_action_list[0]
+	_enemy_action_list.remove_at(0)
```
Turtyo, line 94: is the `EnemyAction` passed as reference or as copy ?

Tests/test_card_animation.gd
```diff
@@ -0,0 +1,65 @@
...
+
+
+func test_player_animation_created() -> void:
+	var card: CardBase = load("res://Cards/Resource/Card_Damage.tres").duplicate()
+	var cast_anim: CastAnimationData = CastAnimationData.new()
+	cast_anim.cast_animation_scene = load("res://Cards/Animation/Anim_Slap.tscn")
+	cast_anim.cast_position = CastPos_Caster.new()
+	card.card_effects_data[0].animation_data = cast_anim
+	
+	card.on_card_play(_player, _enemy)
```
Turtyo, line 30: is there a reason for playing the card ? isn't the animation already supposed to play the card after it's done ?

Tests/test_card_animation.gd
```diff
@@ -0,0 +1,65 @@
...
+	cast_anim.cast_animation_scene = load("res://Cards/Animation/Anim_Slap.tscn")
+	cast_anim.cast_position = CastPos_AllTargets.new()
+	card.card_effects_data[0].animation_data = cast_anim
+	card.card_effects_data[0].targeting_function = TargetAllEnemies.new()
+	
+	card.on_card_play(_player, _enemy)
+	
+	# CastPos_AllTargets should create an animation for all enemies
+	var created_cast_nodes_enemy_1: Array[Variant] = Helpers.get_all_children_nodes_of_type(_enemy, CastAnimation)
+	var created_cast_nodes_enemy_2: Array[Variant] = Helpers.get_all_children_nodes_of_type(_enemy_2, CastAnimation)
```
Turtyo, line 75: so basically each of those arrays should be of size 1 right ?
wouldn't it be better to test that rather than testing that the sum of the size of the two array is 2 ?

Tests/test_card_animation.gd
```diff
@@ -16,43 +16,57 @@ func after_each() -> void:
...
+	var card: CardBase = CardBase.new()
 	var cast_anim: CastAnimationData = CastAnimationData.new()
 	cast_anim.cast_animation_scene = load("res://Cards/Animation/Anim_Slap.tscn")
 	cast_anim.cast_position = CastPos_Caster.new()
+	
+	card.application_type = GlobalEnums.ApplicationType.ENEMY_ONLY
+	card.card_effects_data.append(EffectData.new())
 	card.card_effects_data[0].animation_data = cast_anim
+	card.card_effects_data[0].targeting_function = TargetingBase.new()
+	card.card_effects_data[0].effect = EffectDamage.new()
```
Turtyo, line 28: any reason for creating the card this way instead of using the already made `Card_Damage.tres` ?

Turtyo (commented):
This seems very good to me
I'll hold on on the approve for now, as i want to pull the branch and test things in local too, to be sure everything works correctly.
But at a first glance, it seems good :+1:

Cards/CardBase.gd
```diff
@@ -57,30 +57,35 @@ func _handle_effects_queue(caster: Entity, base_target: Entity) -> void:
...
+			card_effect.apply_effect_data(caster, current_target)
 	
-	_card_effects_queue.remove_at(0)
-	
-	# Apply effects to all targets
-	for current_target in list_targets:
-		card_effect.apply_effect_data(caster, current_target)
+	# Wait for last animation to complete
+	if not created_cast_animations.is_empty():
+		await created_cast_animations[created_cast_animations.size() - 1].on_animation_cast_complete
```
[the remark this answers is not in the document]
cheesefrycook: Yes that is correct

Cards/CardBase.gd
```diff
@@ -57,30 +57,35 @@ func _handle_effects_queue(caster: Entity, base_target: Entity) -> void:
...
+	_card_effects_queue.remove_at(0)
 	
 	# Handle next effect in the queue or finish casting
 	if _card_effects_queue.size() > 0:
 		_handle_effects_queue(caster, base_target)
 	else:
 		CardManager.on_card_action_finished.emit(self)
+
+
+func animation_hit(hit_target: Entity, card_effect: EffectData, caster: Entity) -> void:
```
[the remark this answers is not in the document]
cheesefrycook: This function does apply the card effect yes, but the idea is you can use the signal `on_animation_hit_triggered` in `CastAnimation` to apply the effect before the animation actually ends. This is what I did for the slap animation - I call `on_animation_hit_triggered` in the middle of the animation and the effect applies right when the slap connects with the target

Cards/Animation/CastPosition/CastPos_Caster.gd
```diff
@@ -0,0 +1,18 @@
+extends CastPosition
+class_name CastPos_Caster
+## CastPos_Caster creates a single animation at the caster's location
```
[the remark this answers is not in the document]
cheesefrycook: The animation starts at the caster's position. `CastPosition` is intended as an initialization for the animation. So it handles creating the animation and placing it at the correct starting position. The actual movement is handled later in CastAnimation

Core/Battler.gd
cheesefrycook: It should be a reference

Tests/test_card_animation.gd
cheesefrycook: Yeah good point, i'll change

Tests/test_card_animation.gd
cheesefrycook: No, calling `on_card_play` is what triggers the animation

Tests/test_card_animation.gd
cheesefrycook: I kept having a strange issue when trying to create a `Card_Damage` where it would affect other loaded `Card_Damage` resources from other tests causing the tests to fail. It did this even when calling `duplicate()` on the Resource. I couldn't figure it out so I just went with this

Cards/CardBase.gd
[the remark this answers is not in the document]
Turtyo: so could this lead to potentially triggering the card effect twice by mistake if the `animation_hit` function is called on an animation that uses `on_animation_hit_triggered` ?

Core/Battler.gd
Turtyo: if it's a reference, what happens when you remove it from the list ? I think the list just drops the reference to the enemy action from its list, but it shouldn't set it to null I think ?

Tests/test_card_animation.gd
Turtyo: oh I see

Tests/test_card_animation.gd
Turtyo: oh I know this issue, it's a problem of loading cached resources
the solution is to force the card to be loaded without using the cache for the next test at least (or actually after any test that modifies the card)

Core/Battler.gd
cheesefrycook: Yeah it should just drop the reference from the list, shouldn't be set to null

Cards/CardBase.gd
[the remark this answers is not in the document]
cheesefrycook: Yeah that could potentially happen. I can add a `push_error` for that cases

Tests/test_card_animation.gd
cheesefrycook: How do you force the card to be loaded?

Cards/CardBase.gd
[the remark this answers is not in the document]
Turtyo: that would be a good idea

Tests/test_card_animation.gd
Turtyo: you need to do `load(path, cache_mode=2)`
here is [the documentation](https://docs.godotengine.org/en/stable/classes/class_resourceloader.html#enum-resourceloader-cachemode) on that

Turtyo:
I was thinking earlier, I think the animation folder should be in the art folder, or at least have a similar structure and test to check that we are following a proper naming convention.
There might be a lot of card animations, so I think the animation folder should be its own folder in the art folder. But later, we might have other animations that are not card related, we might need to separate each folder in art between static and animation.
I think that's a problem for another time, because that would mean rewriting the test to work on depth 2.

But anyway, for now I think the animation folder should go into the art folder.

cheesefrycook:
Do you mean just the animation .tscn files and the art .png files? Or the scripts too

Tests/test_card_animation.gd
cheesefrycook: Unfortunately this doesn't work for me. The problem arises with the `test_cast_position_all` test and `test_persist_strength_status` test. If I have both of these tests load `Card_Damage`, then the `test_persist_strength_status` test fails like this.
image
I've found out it's because both tests use `_enemy_2`, which for some reason causes problems with the loaded card. If I replace the reference in `test_persist_strength_status` to `_enemy` instead of `_enemy_2` it works fine

Turtyo:
at least the png files
the .gd should not go into art
I'm wondering about the .tscn though

Tests/test_card_animation.gd
Turtyo: that's strange, the enemy shouldn't matter since they are built the same way iirc

cheesefrycook:
I rearranged some things, let me know if that's good. I can also change the naming_convention too if we want

Tysterman74:
> I added a banana throw animation to show how we can override `CastAnimation` for specific animations. Had to do a good bit of rearranging to make it work, but I think the code is better now
> Fauna-RPG.DEBUG.2024-03-16.11-57-08.mp4

lmao this made me HOLLER. I love this so much

Tests/test_card_animation.gd
Turtyo: Well anyway let's leave it as is for now, trying to figure out something later on

Cards/CardBase.gd
```diff
@@ -89,4 +93,9 @@ func _handle_effects_queue(caster: Entity, base_target: Entity) -> void:
 
 
 func animation_hit(hit_target: Entity, card_effect: EffectData, caster: Entity) -> void:
+	if _targets_triggered_hits.has(hit_target):
```
Turtyo, line 96: doesn't that mean we can't play more than one effect on a target ? since the list is cleared in the `on_card_play` only

cheesefrycook: Oh yeah forgot about that, i'll fix it

Cards/CardBase.gd
```diff
@@ -84,6 +83,7 @@ func _handle_effects_queue(caster: Entity, base_target: Entity) -> void:
 			push_error("Did not trigger a hit on all targets for effect " + card_effect.resource_path)
 	
 	_card_effects_queue.remove_at(0)
+	_targets_triggered_hits.clear()
```
Turtyo, line 86: this still might be a problem later on if we decide we want an effect that applies the same effect multiple times (for whatever reason), depending on how we do it. We can probably make it work still if the effect when triggered just files the list of effects with the repetition of the desired effect
I will approve and keep this in mind

Turtyo (approved):

PR #85, opened 2024-03-21
Issue-76: Show dialog when player set torch down but Fauna isn't in a room yet

### LINKED ISSUE #76: Player can put a torch down while they are not on the map
## Describe the bug
The first time the player opens the map they can select a room to go on the first floor. But the add torch part is not disabled, meaning a player can add a torch to the map while not being on the map yet

## To Reproduce
1. Open the map
2. Don't select a room on the map
3. Click on Add Torch on the left
4. Click yes
5. Game freezes, because the player position is not yet set

## Expected behavior
Add torch should not be visible if the player has not gone into a room yet

## Additional context
There will be more changes, later on, with disabling the button if the player doesn't have torches anymore for example. The button shouldn't be invisible then, but should have a different texture when disabled (to show it's there but not available).
It would also be good to have a popup when the button is hovered to show the reason (not enough torches for example).

Tysterman74:
# Description

Show a dialog if player tries to set a torch down if Fauna isn't in a room yet.

## Related issue(s)

Closes #76 

## List of changes

- Added new AlertDialog in MapUI for handling not being able to set a torch
- Added exports for text string of what text to display if Fauna isn't in a room yet
- Added a TODO cause we don't have torch count yet

## Tests

N/A

## Additional notes

image

### DIFF
--- a/UI/MapUI.gd
+++ b/UI/MapUI.gd
@@ -16,6 +16,8 @@ var _LIGHT_FLOOR_RANGE: int = 3
 @export var room_container: ColorRect
 @export var room_addition_node: Control
 @export var torch_confirmation_dialog: ConfirmationDialog
+@export var cant_set_torch_dialog: AcceptDialog
+@export var player_position_not_set_text: String
 
 var room_ui_array: Array[Array]
 var current_player_room: RoomUI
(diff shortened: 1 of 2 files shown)


UI/MapUI.gd
```diff
@@ -31,7 +33,12 @@ func _on_return_button_press() -> void:
 	
 
 func _on_add_torch_pressed() -> void:
-	torch_confirmation_dialog.show()
+	if !PlayerManager.is_player_initial_position_set:
+		cant_set_torch_dialog.dialog_text = player_position_not_set_text
+		cant_set_torch_dialog.show()
+	#TODO: Add in torch count here
```
Turtyo, line 39: This should be done later, i was thinking it should either be a player property, or an inventory property (which doesn't exist yet but would be used to put things like money, torches and items)

Turtyo (approved):
I would just use "put" instead of "set" in the dialogue text

cheesefrycook (approved):

PR #86, opened 2024-03-21
Global scene loader (map scene loader excluded)

### LINKED ISSUE #74: Global scene loader (map scene loader excluded)
## Feature Description

Load a different scene when changing rooms.

## Motivation

The only scene currently in the game is the combat scene and moving on the map doesn't change anything. It would be good to have the associated event with each room be loaded when we go in.
Note that this scene loader should also be able to stuff outside of the events of each room, for example the start menu, the end credits, etc...
This second point is up to discussion, we could decide to have a single loader for all the game, or make a general loader and a another loader for the map in particular

## Acceptance Criteria

- Load the event associated with the room you move into
- Don't allow the player to move to a new room if the event in the current room is not finished
- A DEBUG const to be able to ignore the previous point for devs only (see `DEBUG_VAR.gd` after alpha 1.0)

## Proposed Solution

The scene switcher would be based off of [this example in the Godot documentation](https://docs.godotengine.org/en/stable/tutorials/scripting/singletons_autoload.html#custom-scene-switcher), which uses a singleton as scene loader.

The block of the scene switch can be done using a `is_map_movement_allowed` boolean set to true by default when the game starts. We have a function named `toggle_map_movement`. We make two signals, `event_started` and `event_ended`. We connect `toggle_map_movement`to those signals.
Upon receiving a signal, the toggle function will put the boolean to true if it's false, and false if it's true. When we start, the map movement is set to true and the player can choose a room on the first floor. We enter an event, a signal is emitted, the map movement is not allowed anymore. When the event ends, the end signal is emitted and `is_map_movement_allowed` becomes true again, thus allowing the player to move again.
The DEBUG const can easily be done by having the getter for `is_map_movement_allowed` return `DEBUG_MAP_MOVEMENT_ALLOWED or is_map_movement_allowed`. If the DEBUG is set to true, the getter will always return true since we use a or expression.

## Additional Context

This needs map to be merged into main (so after alpha 1.0)

Edit: This PR will deal with making the main switcher, the map switcher will be done in a later PR

Turtyo:
# Description

Create a global loader that allows for scene switching

## Related issue(s)

Closes #74 

## List of changes

- Add a main menu, an option menu, a quit button and a button to start the game
- Add a button to go back to main menu from the game ( @cheesefrycook i would like your opinion on this to integrate it with player save, as currently it will reset the combat)
- Make the main menu the new main scene

## Tests

None

## Additional notes

This is the option menu with the button to go back to the main menu in the middle, and it made me laugh when I made it
gscreenshot_2024-03-21-212932


Turtyo:
I considered pre-loading the scenes that are accessible from a given scene, but I found it's not necessary for now as they all load directly.
If later on we have scenes that take a lot of time to load, we can load them in a background thread when entering a scene that can access such long loading scene. It should not be too hard to implement with current structure.

#Scenes/SceneScripts/TestingScene.gd
```diff
@@ -0,0 +1,7 @@
+extends Node2D
+
+## * We do this to properly init the battle state
+## Before, the testing scene was the main scene so it was properly started by the phase manager
+## Now that the main scene is the main menu and we load the testing scene from there, we init to properly init the game state 
+func _ready() -> void:
+	PhaseManager.initialize_game()
```
cheesefrycook, line 7: Since we have this now, should we remove the `initialize_game` call in PhaseManager _ready?

cheesefrycook:
This looks good! For the saving, I think that is something we can address once we have the map scene switching fully integrated. Right now it saves any time the player is destroyed. But in the actual game the plan is to save the state whenever you enter a map node. That way if you exit the game in the middle of a combat, your current progress will be reverted to the start of that combat.

#Scenes/SceneScripts/TestingScene.gd
Turtyo: I think we could yes... since it will be called upon scene loading
It just won't work if you directly load the scene from the start of the game instead of loading it from the global loader

Turtyo:
Oh yeah I forgot leaving a fight early actually resets the fight
Later on, I'm thinking of putting an option button on the top left (instead of go back to main menu), so you could adjust things like volume during the run without having reload problems
And yeah players could save scum by resetting fights that don't go well, but if that's how they want to have fun, who are we to decide

Turtyo:
I noticed a little something that will also need to be dealt with later with the save system
When going back to the main menu and going back in the fight again, enemies get back to full HP, but the player keeps the damage they already took during the fight (due to the fact player data is stored globally whereas enemies are instantiated)

cheesefrycook:
This is why I mentioned above that we only save the game when you enter a map node. So in this case, if you exit the combat in the middle then the whole combat is reset to the starting state when you re-enter. This will simplify the saving a lot for us because we won't have to worry about saving the state of the entire combat. All we need to know is which map node to reload. We will just have to provide a warning pop up like "progress in this combat will not be saved"

cheesefrycook (approved):

Turtyo:
Yes totally, just need to reset player data to before the fight too

cheesefrycook (approved):

Managers/PhaseManager.gd
```diff
@@ -14,7 +14,8 @@ var current_phase: GlobalEnums.Phase = GlobalEnums.Phase.NONE
 
 
 func _ready() -> void:
-	initialize_game()
+	# initialize_game()
```
Tysterman74, line 17: Is there a reason this is commented out?

Tysterman74 (approved):

Managers/PhaseManager.gd
Turtyo: The phase manager is a singleton, meaning it's loaded first
It was fine to use `initialize_game` (which should probably be renamed to `init_combat` instead)
This is now commented out, since we init the combat when entering the scene (which is done via the scene switcher, and not at game start)

PR #89, opened 2024-03-24
Auto-focus map display on player position 

JonaLam:
# Description

Scrolls the map to fauna when entered or to the bottom if fauna is not instanced

## List of changes

Added function to SmoothScrollContainer to scroll_to_fauna
MapUi script calls SmoothScroll on _ready

## Additional notes

The scroll_to_fauna function uses the MapManager's current_map's room array size to figure out how many rooms are in the current_map, which might be problematic if that is changed significantly in the future

### DIFF
--- a/addons/SmoothScroll/SmoothScrollContainer.gd
+++ b/addons/SmoothScroll/SmoothScrollContainer.gd
@@ -636,13 +636,11 @@ func scroll_to_fauna(duration:float=0.5) -> void:
 	
 	var room_distance: float = 66
 	var scroll_to_pos:float = -room_distance
-	var rooms_between_fauna_and_top = MapManager.map_width_array.size() - 8
+	var rooms_between_fauna_and_top : int = MapManager.map_width_array.size() - 8
 	
 	scroll_to_pos -= (rooms_between_fauna_and_top - PlayerManager.player_position.y) * room_distance 
 	
 	scroll_y_to(scroll_to_pos)
-	
-	pass
 
 
 func is_outside_top_boundary(y_pos: float = pos.y) -> bool:
--- a/addons/SmoothScroll/SmoothScrollContainer.gd
+++ b/addons/SmoothScroll/SmoothScrollContainer.gd
@@ -630,6 +630,21 @@ func scroll_to_left(duration:float=0.5) -> void:
 func scroll_to_right(duration:float=0.5) -> void:
 	scroll_x_to(self.size.x - content_node.size.x, duration)
 
+func scroll_to_fauna(duration:float=0.5) -> void:
+	if(!PlayerManager.is_player_initial_position_set):
+		return
+	
+	var room_distance: float = 66
+	var scroll_to_pos:float = -room_distance
+	var rooms_between_fauna_and_top = MapManager.map_width_array.size() - 8
+	
+	scroll_to_pos -= (rooms_between_fauna_and_top - PlayerManager.player_position.y) * room_distance 
+	
+	scroll_y_to(scroll_to_pos)
+	
+	pass
+
+
 func is_outside_top_boundary(y_pos: float = pos.y) -> bool:
 	return y_pos > 0.0
 
--- a/addons/SmoothScroll/SmoothScrollContainer.gd
+++ b/addons/SmoothScroll/SmoothScrollContainer.gd
@@ -634,11 +634,16 @@ func scroll_to_fauna(duration:float=0.5) -> void:
 	if(!PlayerManager.is_player_initial_position_set):
 		return
 	
-	var room_distance: float = 66
-	var scroll_to_pos:float = -room_distance
-	var rooms_between_fauna_and_top : int = MapManager.map_width_array.size() - 8
+	var map_UI : MapUI = get_parent()
+	var current_map_texure_button : TextureButton = Helpers.get_first_child_node_of_type(map_UI.current_player_room, TextureButton)
 	
-	scroll_to_pos -= (rooms_between_fauna_and_top - PlayerManager.player_position.y) * room_distance 
+	var max_scroll = content_node.size.y - self.size.y
+	var room_distance : float = map_UI.get_combined_room_height(current_map_texure_button) / MapManager.current_map.rooms.size()
+	var scroll_to_pos : float = -max_scroll
+	
+	var rooms_behind_seen : int = 1
+	
+	scroll_to_pos += room_distance * (PlayerManager.player_position.y - rooms_behind_seen)
 	
 	scroll_y_to(scroll_to_pos)
 
--- a/addons/SmoothScroll/SmoothScrollContainer.gd
+++ b/addons/SmoothScroll/SmoothScrollContainer.gd
@@ -635,7 +635,7 @@ func scroll_to_fauna(duration:float=0.5) -> void:
 		return
 	
 	var map_UI : MapUI = get_parent()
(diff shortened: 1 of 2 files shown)


addons/SmoothScroll/SmoothScrollContainer.gd
```diff
@@ -630,6 +630,21 @@ func scroll_to_left(duration:float=0.5) -> void:
...
+	
+	var room_distance: float = 66
+	var scroll_to_pos:float = -room_distance
+	var rooms_between_fauna_and_top = MapManager.map_width_array.size() - 8
+	
+	scroll_to_pos -= (rooms_between_fauna_and_top - PlayerManager.player_position.y) * room_distance 
+	
+	scroll_y_to(scroll_to_pos)
+	
+	pass
```
Turtyo, line 645: return instead of pass would be better

addons/SmoothScroll/SmoothScrollContainer.gd
```diff
@@ -630,6 +630,21 @@ func scroll_to_left(duration:float=0.5) -> void:
 func scroll_to_right(duration:float=0.5) -> void:
 	scroll_x_to(self.size.x - content_node.size.x, duration)
 
+func scroll_to_fauna(duration:float=0.5) -> void:
+	if(!PlayerManager.is_player_initial_position_set):
+		return
+	
+	var room_distance: float = 66
```
Turtyo, line 637: is this the distance in pixel / unit between floors ?
if yes, you can get it without hard-coding in two different ways:
- either use `_get_combined_room_height` (which will need to be renamed to `get_combined_room_height` to show it's not private anymore) and divide by the number of floors
- replicate the logic of `_get_combined_room_height` to calculate the height of a room at a floor index y

It's better to not have hard-coded values

addons/SmoothScroll/SmoothScrollContainer.gd
```diff
@@ -630,6 +630,21 @@ func scroll_to_left(duration:float=0.5) -> void:
...
+	var scroll_to_pos:float = -room_distance
+	var rooms_between_fauna_and_top = MapManager.map_width_array.size() - 8
```
Turtyo, line 639: why 8 ?

Turtyo (changes requested):
The hard-coded values will need to be changed, as we can totally decide to change MapUI representation, textures, height etc...

addons/SmoothScroll/SmoothScrollContainer.gd
JonaLam: It's that many slots which are shown when you are scrolled to the top, I'll put it in a variable

addons/SmoothScroll/SmoothScrollContainer.gd
JonaLam: Ye, I'll fix that I didn't know about these variables

addons/SmoothScroll/SmoothScrollContainer.gd
Turtyo: why are you using the map width array instead of the map height then ?
and yes, naming numbers is better

addons/SmoothScroll/SmoothScrollContainer.gd
Turtyo: I think it would be better to replicate the logic of `_get_combined_room_height` to just get the current height
because the function is already doing texture size * total height, but we can just do texture size * index (+ some offset probably)

addons/SmoothScroll/SmoothScrollContainer.gd
JonaLam: I couldn't find the map height

Turtyo: ah yes sorry, the size of the `map_width_array` is indeed the height of the map, since it's an array of the width of all the floors

addons/SmoothScroll/SmoothScrollContainer.gd
```diff
@@ -634,11 +634,16 @@ func scroll_to_fauna(duration:float=0.5) -> void:
 	if(!PlayerManager.is_player_initial_position_set):
 		return
 	
-	var room_distance: float = 66
-	var scroll_to_pos:float = -room_distance
-	var rooms_between_fauna_and_top : int = MapManager.map_width_array.size() - 8
+	var map_UI : MapUI = get_parent()
+	var current_map_texure_button : TextureButton = Helpers.get_first_child_node_of_type(map_UI.current_player_room, TextureButton)
```
Turtyo, line 638: I tested and  `var current_map_texure_button : TextureButton = map_UI.current_player_room.button;` also works, it's more explicit and there won't be a problem if we have multiple texture buttons on this later on

Turtyo (commented):
Seems better
When I tested it, I noticed that we see the scroll happening when opening the map; Could we maybe add a little delay before showing the map to let the smooth scroll the time to focus on the correct position ? So we don't have this impression of the map jumping around
Not sure if it's actually possible, but I think it should be tried

addons/SmoothScroll/SmoothScrollContainer.gd
JonaLam: Thanks, didn't know about that variable.

Regarding the scrolling, that is a built in thing to scroll smoother, so I can change that by just by giving a parameter

Turtyo: the problem isn't how hard it scrolls, the problem is that we see the movement
even if you set the smoothness to 0, if we spawn the map directly I'm not sure the display won't feel like it's jumping to the position, thus why i said we should add a little delay
if it works without the delay, then good
but we also need the smooth scroll after the map loaded for the player

JonaLam: I don't see any problems like you're describing, at least not after the changes I made, so I hope it's fine

Turtyo: Mmh I'm seeing a little quick jump around floors 3 or 4, but it's not very noticeable
It might become more noticeable if we add more textures, or maybe it won't i'm not sure
Let's keep it like that for now

Turtyo (approved):

Tysterman74:
> Seems better When I tested it, I noticed that we see the scroll happening when opening the map; Could we maybe add a little delay before showing the map to let the smooth scroll the time to focus on the correct position ? So we don't have this impression of the map jumping around Not sure if it's actually possible, but I think it should be tried

I believe if you don't pass any arguments in MapUI, there should be a delay of 0.5 so it won't look as jarring.

Turtyo:
> I believe if you don't pass any arguments in MapUI, there should be a delay of 0.5 so it won't look as jarring.

Which function to not pass arguments to ?

Tysterman74:
> > I believe if you don't pass any arguments in MapUI, there should be a delay of 0.5 so it won't look as jarring.
> 
> Which function to not pass arguments to ?

The function Jona made we're adding a default for the duration
```
func scroll_to_fauna(duration:float=0.5) -> void:
```
But where we're calling it in MapUI.gd we pass in 0
```
if(!PlayerManager.is_player_initial_position_set):
		scroll_container.scroll_to_bottom(0)
	else:
		scroll_container.scroll_to_fauna(0)
```

If we were to call
```
if(!PlayerManager.is_player_initial_position_set):
		scroll_container.scroll_to_bottom()
	else:
		scroll_container.scroll_to_fauna()
```
The duration variable will be set to 0.5 instead of 0, so that will cause the scroll to have a bit of a duration.

Tysterman74 (approved):

Turtyo:
Humu I see

Issue #75, opened 2024-02-21
Implementation of the event Dialogue


Turtyo:
## Feature Description

The event Dialogue is on where the player is faced with a scenario and they have multiple choices to choose from, potentially gaining or losing things from their choices.

## Motivation

This event is important since it makes it so that we don't only have combats, shops and rest in the game.

## Concept art / Designs

An example in Slay the Spire
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/d7697a18-aa86-4f79-b62a-9d925b26d050)

An example in Faster Than Light, where you don't have the consequence of each choice written. Result is always the same, but you need to know by playing enough or using your logic
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/c80b65da-5a8d-4643-8d7d-e15ad8410b2d)

Note that it doesn't need to look like that, nor do we need to indicate the consequence of each action.

## Acceptance Criteria

[Define the criteria that need to be met for this feature to be considered complete.]

- A text box and multiple choices
- Possibility to have colored text
- Animations can play during the event

## Proposed Solution

I was thinking we could start by using [this nice addon](https://github.com/nathanhoad/godot_dialogue_manager) that already seems to do most of the heavy lifting for the dialogue part.
The image and animation part can be an animation player probably

Edit : using the addon we might have a problem with licenses not being compatible, need to check that first

JonaLam:
The license seemed fine, but someone should also read it as I'm not sure I fully understood it.

Turtyo:
I think each event should have its associated dialogue options. We could have a database (which could just be a singular file with all the possible events and related dialogues), to hold the dialogue linked to each event. This would be loaded upon entering a room with the related event, along all other necessary resources (such as art or music for example, which could also be listed in this event file;
this would allow us to have a standard format for events, which allows us to just have a single scene for events, that loads different data; and not have to do a scene per event (which would quickly start to add up).

I will try to read the license in more details to be sure it's compatible. If it is not, we can check what they did but we won't be able to copy, we will have to do our own stuff basically.

Tysterman74:
Should we potentially talk with art and design on how we want to how the event dialogues UI look like? The add on looks like it has a lot of customization, I think it would be helpful to get a first pass like a wiremock or something of how we want it to look.

[...]

Issue #87, opened 2024-03-23
Scene loader for the map


Turtyo:
## Feature Description

Load all information related to the event of the room you move in.

## Motivation

This is to link the map to the actual scene displayed to the player, meaning there is actual progress, instead of being stuck on the same scene forever.

## Acceptance Criteria

- Clicking on a room of the map loads the related event of the room (it can be a single default scene for now, we'll deal with loading random scenes later)
- Moving to the next room can only be done when the event of the current scene is over (combat finished, dialogue event, etc...)
- Have a debug const to allow dev to still move between rooms, even if the event in the current scene is not over
- Finishing an event shows a reward screen (if the event gives a reward), and shows the map.

## Proposed Solution

The block of the scene switch can be done using a `is_map_movement_allowed` boolean set to true by default when the game starts. We have a function named `toggle_map_movement`. We make two signals, `event_started` and `event_ended`. We connect `toggle_map_movement`to those signals.
Upon receiving a signal, the toggle function will put the boolean to true if it's false, and false if it's true. When we start, the map movement is set to true and the player can choose a room on the first floor. We enter an event, a signal is emitted, the map movement is not allowed anymore. When the event ends, the end signal is emitted and `is_map_movement_allowed` becomes true again, thus allowing the player to move again.
Some events such as the shop can directly emit the event ended signal after the event started, but it shouldn't show the map directly I think;
The DEBUG const can easily be done by having the getter for `is_map_movement_allowed` return `DEBUG_MAP_MOVEMENT_ALLOWED or is_map_movement_allowed`. If the DEBUG is set to true, the getter will always return true since we use a or expression.

When clicking on a room, we check the room we are moving to and the associated event.
We send the associated event information to the GlobalLoader, which will then load the related scene event (which can later be changed to a random selection of the events of the given type).
Sending the information to the global loader also means it can keep track of all the events that have been loaded, and eventually try to not repeat events that have already been selected previously (like to not randomly select the same discussion event for example). That's not to be done now, but it's a consideration for it for later, and the reason I think we should use the global loader for that.

We could also have a common event_init function for all the scenes, that all take the same information, and that is always called by the _ready() function of such scene (which is triggered upon being loaded by the Global Loader).

## Additional Context

This is a follow up of the issue #74

Tomzkk:
I would say overall the proposed solution makes sense. 
[...]

PR #92, opened 2024-04-05
Issue-75: Event Dialogue System

Tysterman74:
# Description
Creation of Dialog Screen for showing events.

## Related issue(s)

Relates to #75 

## List of changes

- Addition of Add-on Dialogue Manager
- Creation of custom dialog scene
- DebugFlag to show test dialog screen when clicking on an eligible room

## Tests

Not added yet.

## Additional notes

https://github.com/Saplings-Projects/1M_sub/assets/5923102/6920700c-b72f-4551-9a42-0eb4513c1f67


Art/Event/rescued_saplings_comfy_sleepy.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
Turtyo, line 1: the persona should be in NPC
also they don't seem to respect the naming convention

Art/Event/test_background.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
Turtyo, line 1: There is no newline for the Event folder in the naming convention which makes tests fail
I would suggest adding the line:
Event : <name>
But this is a background, so maybe it should go in Art/Background
Or maybe we want to have a Background folder in the Event folder, in which case I need to modify the test for the naming convention

Turtyo (commented):
It seems to look nice
Did you write everything in balloon.gd or did you copy from an example ?

It's good to have a separate scene so we can implement scene switching on the map once #87 has been discussed

Tests are failing, for the convention it's normal since no line has been added for events.
For the map maybe we should have a boolean to not load dialogues ? i'm not sure, would need to look at the tests in mapUI again

Dialog/balloon.gd
```diff
@@ -0,0 +1,176 @@
...
+			balloon.focus_mode = Control.FOCUS_NONE
+			responses_menu.show()
+		elif dialogue_line.time != "":
+			var time: float = dialogue_line.text.length() * 0.02 if dialogue_line.time == "auto" else dialogue_line.time.to_float()
+			await get_tree().create_timer(time).timeout
+			next(dialogue_line.next_id)
+		else:
+			is_waiting_for_input = true
+			balloon.focus_mode = Control.FOCUS_ALL
+			balloon.grab_focus()
```
Turtyo, line 86: Did you find this yourself or copied from an example ? I'm curious

Dialog/balloon.gd
```diff
@@ -0,0 +1,176 @@
...
+		flavor_image_3.texture = flavor_image
+
+# Custom function that is interacted via the .dialogue file to add some flavor text based on the index passed in
+func add_flavor_text(flavor_text: String, flavor_number: int) -> void:
+	if (flavor_number == 1):
+		flavor_text_1.add_text(flavor_text)
+	elif (flavor_number == 2):
+		flavor_text_2.add_text(flavor_text)
+	elif (flavor_number == 3):
+		flavor_text_3.add_text(flavor_text)
```
Turtyo, line 173: I think it would be best to make this a match statement instead, as it will set the example for other files

Does it also mean we need one such .gd file for each new dialogue ?

Turtyo (commented):
I have a few questions which I left in comments.
My main concern is that I feel like we will need one script file for each new dialogue, but maybe I misunderstood

Dialog/balloon.gd
Tysterman74: image

With the addition of the add-on, there's a way to make a copy of the dialogue balloon which generates this balloon.gd file.

Dialog/balloon.gd
Tysterman74: I'll update to match.

And no, balloon.gd is used in balloon.tscn which is the packed scene that is used to bring up the dialog screen. 
[This change](https://github.com/Saplings-Projects/1M_sub/pull/92/files#diff-b7e8404905efe7afd6c8148ccc42e6f2cc45a58bc44ed0b5007a74bc3398c5daR199) will pull the packed scene and gets the dialog file we specify. In this case we get the test.dialogue. So we don't need multiple .gd scripts, we need multiple dialogue files.

Dialog/balloon.gd
Turtyo: Humu, interesting

Dialog/balloon.gd
Turtyo: okay, but it seems to me that since we only have match against number 1 to 3, we can only have 3 flavor text. What does each flavor text represent ?
Also it would be best to rename this from balloon to something that shows it's the file managing the dialogues

Tysterman74: image

Each flavor text corresponds to the fixed location on the balloon. I marked it on the screenshot. This is what I was alluding to in terms of the fake "randomness" of the doodles and writings that was discussed with design. Within the .dialogue file we can call `add_flavor_image()` or `add_flavor_text()` and add whatever image/text to supplement the event we're trying to display. This also isn't limited to three, those were just three places I set those Controls on, we can very well add more if we want and update this balloon.gd file.

Turtyo: oh i see, it makes more sense now

Turtyo (approved):

Turtyo (approved):

Turtyo:
Yeah I think that was the idea

JonaLam (approved):
Seems good. customizing the event and adding functionality like adding inv items and interacting with the player seems easy enough so I have complain about

PR #103, opened 2024-05-23
Issue-101: Update Display Size to 1920x1080

### LINKED ISSUE #101: Resize scenes to 1920x1080

Current display size is `1280x720` but we are making our assets for `1920x1080`, so we are compressing to 720p to display back in 1080 on most monitors, which is not ideal.

- Change the size in the project settings (Display -> Window)
- Resize all the scenes

## Describe alternatives you've considered

As we resize the scenes, it would also be good to change the fill type of images from a given size to a "fill with same aspect ratio", that way if we resize later we don't need to change all the scenes again. It might not be possible everywhere, but getting it on as much scenes as possible would be good.

## Additional context
This would be good to do when there aren't too many PR open as that will probably lead to quite a lot of merge conflicts (and merge conflicts on .tscn files are never fun to solve)

Tysterman74:
# Description

Update Display Size to 1920x1080 and updated scenes to match size

## Related issue(s)

Closes #101 

## List of changes

Updated the following scenes to fit the new display size
- CardScrollUI.tscn
- MainMenu.tscn
- Mob/0.tscn
- MapUI.tscn
- EventDialogWindow.tscn
Other changes:
- Update MapUI.gd to have the rooms fit more dynamically if the screen size does change again
- Update project.godot to new display size of 1920x1080

## Tests
No tests updated/created since it was size changes and no changes to logic

## Additional notes

Accidentally modified TestingScene.tscn, but it looks like we may not need it anymore. Should I just remove it from this PR?


Turtyo (changes requested):
Launching the game, there are a few more things that would need resizing probably:
- the map, as well as the map buttons
- the overlay icons (ie the map and deck icon, they need to be changed in the overlay scene)

Apart from that it seems good to me !

Tysterman74:
>     * the map, as well as the map buttons

In terms of the map, it's dynamically sized based on the rooms that are generated. There's a minimum width of 510 and minimum height of 490 to set it to if the maps we generate are too small. Do we want to adjust that?

Turtyo:
Yeah I think we probably need to generate that

UI/MapUI.gd
```diff
@@ -13,16 +13,15 @@ var balloon_scene: PackedScene = load("res://Dialog/EventDialogueWindow.tscn")
 var test_dialog: DialogueResource = load("res://Dialog/test.dialogue")
 
 var _padding_offset: int = 20
-var _MINIMUM_ROOM_WIDTH: int = 510
-var _MINIMUM_ROOM_HEIGHT: int = 490
+var _MINIMUM_ROOM_WIDTH: int = 910
+var _MINIMUM_ROOM_HEIGHT: int = 890
```
Turtyo, line 17: actually, isn't that the minimum map width and height ? because room_width and height makes it sound that a single room is at least 910x890
Regarding this, I would make the map a bit bigger than what we already have. The rooms are dynamically sized to fit inside the map container, so a bigger map would give us bigger rooms, which might be better than to click on small rooms

Turtyo (commented):
Apart from making the map container bigger, all seems good to me
I'll approve once we finish discussing about the map

Turtyo (approved):
The light radius might be just a bit too big, because we can see some pixels of the underside of letters, but it doesn't really matter right now because we will change the way the light overlay works later anyway (by making it a gradient, as we discussed in a previous issue)

Paper-2 (approved):
Tested in a 2k monitor and a 1080p,  looks good to me overall but the end turn button and the cards in the CardScrollUI could be bigger.

PR #105, opened 2024-06-02
Fix status are carried between fights

TripleCubes:


Turtyo:
The PR is not using the PR template to describe the fix, closing

Issue #97, opened 2024-04-25
Change the way event skip button works to spawn them on any scene


Turtyo:
## Is your amelioration related to a problem? Please describe.

The skip button just exists in placeholder scenes, but we could imagine wanting to skip real scenes for testing purpose. This behaviour is already partially controlled by the `DEBUG_VAR` named `DEBUG_SKIP_EVENT` 

## Describe the solution you'd like
Instead of enabling and disabling a button in some scenes, a solution would be to instantiate through code a button in all the scenes if the skip event debug var is set. It doesn't auto-skip the scene, but you can do it if you want. That way we can skip any scene without needing to update all the different scene files we have

Scenes that don't have a clear end condition yet should have the skip button on by default. I'm not sure how we would do that, as it would mean we need to have a way to say if a certain scene has a proper end condition. Any idea is welcome for this.

## Additional context
This needs to be done after #87

Turtyo:
You can put a script on the top node. All events are already child of `EventBase`, maybe we can put the check for the debug variable inside EventBase ? With an eventual override per event (ie mob, dialogue, heal and shop) if we want to customize the position.
I never tried spawning a button directly with code, but that might be doable, we'll need to check

TripleCubes:
I dont see how any event is related to EventBase. I checked the #Event folder, all events there have top node be a Control, there is no EventBase there

Spawning nodes from code is like one of the most basic thing to do in Godot. I dont 100% remeber the syntax but it literally just

```
const btn_scene = preload("PATH TO BTN SCENE")

func _ready():
    var btn: Btn = btn_scene.instantiate()
    self.append(btn)
end
```

TripleCubes:
Relevant link:
https://docs.godotengine.org/en/stable/tutorials/scripting/nodes_and_scene_instances.html#instancing-scenes

As for the checking if event have a proper end, we can just give the event a self.dont_have_proper_end variable, and then we check if this variable exist and is it true, and after than create the skip button?

TripleCubes:
This also mean you have to check if the variable exist every time you want to access the variable tho, which you might not want to do. We can also add the have_proper_end variable to BaseEvent (I realized we should name it have_proper_end instead of dont_have_proper_end because dealing with negative bool is annoying) and set it to true by default, and events that dont have proper end can set it to false

Sorry for multiple messages

Turtyo:
[...]

PR #109, opened 2024-06-08
Skip event button self delete

TripleCubes:
# Description

Skip event button self delete

## Related issue(s)

#97

## List of changes

Skip event button will delete itself if DEBUG_SKIP_EVENT is false. SkipEventButton is now available in the node selection menu. SkipEventButton doesnt need to connect its pressed() signal anymore since the _pressed() function is called directly.

Also add SkipEventButton to mob and dialogue scene, altho dialogue scene still have bug that prevent any button other than the dialogue button from being pressed.

## Tests

If you wrote test scripts to use with GUT, provide a summary of those tests here.

## Additional notes

If you have anything more to add.

### DIFF
--- a/#Scenes/Events/dialogue/0.tscn
+++ b/#Scenes/Events/dialogue/0.tscn
@@ -251,10 +251,10 @@ autowrap_mode = 2
 
 [node name="SkipEventButton" type="Button" parent="."]
 offset_left = 64.0
-offset_top = 987.0
-offset_right = 111.0
+offset_top = 983.0
+offset_right = 208.0
 offset_bottom = 1021.0
-text = "Skip
+text = "Skip this event
 "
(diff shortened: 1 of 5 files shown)


Turtyo (changes requested):
I would make the skip button bigger for the mob event and the dialogue event (as it is now, I think it's a bit too small).
I also notice that the button is not enabled by default for the shop and the heal. We can either decide:
- those are temporary screen anyway, if people want to skip them they just need to activate the debug option, and it wont' stay like that for long
- it's better if they are activated by default (even if the debug var is set to false), since you can't do anything if it's not activated

I think we can go for solution 1 honestly, it might add more code that we don't really need, we can already skip by activating the debug var anyway

PR #110, opened 2024-06-13
Issue-81: Splash Art Loading Scene

### LINKED ISSUE #81: Game splash art loading
## Feature Description

An introduction animation for the game when it is launched.

## Motivation

Currently the splash art loading time is not constant and is dependent on the speed at which your computer can launch the game. This is pretty short, and we don't have much time actually to see the splash art (which is a shame because it looks nice).

## Concept art / Designs

image

The actual loading could be a black screen, then we do a fade from black to the splash art, old it for a few seconds and load the first game scene (menu or testing scene depending on when this is implemented compared to other PR).

## Acceptance Criteria

[Define the criteria that need to be met for this feature to be considered complete.]

- We have time to see the splash art
- The first scene then loads normally

## Proposed Solution

I'm not exactly sure how to do this as I haven't looked into solutions yet. I know there are timers in Godot, so we can probably load this intro animation as the first scene with a default timer, and when the timer expires we switch the menu.
The fade from black could be done using shaders but that seems overkill, there might be a simpler solution.

## Additional Context

This issue might need the scene switcher #74

Tysterman74:
# Description

Short splash animation of the Sapling Game Studio logo before navigating to main menu

## Related issue(s)
Closes #81 

## List of changes

- Added SplashScene.tscn
- Added SplashScene.gd
- Added fauna_uuu_test.mp3
- Updated project settings to change boot screen load color to black
- Updated project settings to have the SplashScene be the first scene to load instead of Main Menu

## Tests

N/A

## Additional notes
Sound on.
https://github.com/Saplings-Projects/1M_sub/assets/5923102/0893f4ba-3edc-4db9-ae16-1ecee60cb8b8


#Scenes/SceneScripts/SplashScene.gd
```diff
@@ -0,0 +1,25 @@
...
+var tween: Tween
+
+func _ready() -> void:
+	tween = create_tween()
+	var original_scale: Vector2 = splash_image.scale
+	# Animation to slingshot the logo, scale low the image with a spring like animation 
+	# then scale it back to original size with an elastic like animation
+	# Play Fauna uuuu-ing in-between, then fade splash away and navigate to main menu when done
+	tween.tween_property(splash_image, "scale", Vector2(0.30, 0.30), 1.5).set_trans(Tween.TRANS_SPRING).set_ease(Tween.EASE_IN)
+	tween.tween_callback(_play_sfx)
```
Turtyo, line 15: not sure there is a need to repeatedly call the SFX (since if I understood correctly, that's what the callback does)

#Scenes/SceneScripts/SplashScene.gd
```diff
@@ -0,0 +1,25 @@
...
+	tween.tween_property(splash_image, "scale", original_scale, 0.5).set_trans(Tween.TRANS_ELASTIC).set_ease(Tween.EASE_OUT)
+	tween.tween_property(splash_image, "modulate:a", 0, 1)
+	tween.tween_callback(_on_tween_finish)
```
Turtyo, line 18: same here, probably no need to repeatedly load the scene

Turtyo (changes requested):
Regarding the audio files, we will probably change the way the files are organized later on (basically have them all inside the asset folder, with art on one side and audio on the other side, SFX and music folders).

Visually speaking, there is a little jump to the left on the logo when loading the game. Do you know how to fix that ? Because looking at the code, it shouldn't do that, maybe we'll need to adjust the position manually ? I think it would be good to keep the image at the same size at the start, ie not directly play the animation of the logo changing size.

From an aspect of the animation, ok for the size change but I would keep the lower border of the logo on the lower border of the screen (ie scale relative to an anchor on the bottom and not on the center, I think there is a way to change the anchor point)

#Scenes/SceneScripts/SplashScene.gd
Tysterman74: Both callbacks shouldn't be repeatedly calling? Where are you reading that at?

Tysterman74:
https://github.com/Saplings-Projects/1M_sub/assets/5923102/c9719b11-d2ad-46cd-a554-f65ef9f9115c

#Scenes/SceneScripts/SplashScene.gd
Turtyo: Ah nvm I misread [the doc](https://docs.godotengine.org/en/stable/classes/class_tween.html#class-tween-method-tween-callback)
image
didn't read the example exactly and I assumed the callback was to do a repeated call but that's not the case. What is callback used for, sequential execution ?

Turtyo (commented):
Not exactly sure to understand the transformations you do at the start of the _ready() for the image, i'm guessing it's to change the scaling point of the image ?

#Scenes/SceneScripts/SplashScene.gd
```diff
@@ -1,12 +1,30 @@
...
 @export var uuuu_sfx: AudioStreamPlayer2D
 
 var tween: Tween
 
 func _ready() -> void:
+	var splash_image_size: Vector2 = splash_image.size
+	var splash_image_scale: Vector2 = splash_image.scale
+	
+	# Set the position to the pivot offset, which would be (0,0) on global position
+	splash_image.position = Vector2(splash_image.position.x - splash_image.pivot_offset.x, splash_image.position.y - splash_image.pivot_offset.y)
```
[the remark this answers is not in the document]
Tysterman74: I've reverted this. I ended up setting the position in the scene itself rather than dynamically setting it in th e script.

#Scenes/SceneScripts/SplashScene.gd
```diff
@@ -1,12 +1,30 @@
...
+	
+	# Sets the splash image to what it would be like if it was (0,0) without a pivot offset
+	# Can't use size directly since we scaled it down a bit, so need to make sure we multiply by the scale before hand
+	var splash_image_pivot_x: int = splash_image.position.x + (splash_image_size.x * splash_image_scale.x) / 2
+	var splash_image_pivot_y: int = splash_image.position.y + (splash_image_size.y * splash_image_scale.y)
+	
+	# Set the splash image in the center of the screen
+	var splash_image_x: int = splash_image_pivot_x + color_rect.size.x / 2 - (splash_image_size.x * splash_image_scale.x) / 2
+	var splash_image_y: int = splash_image_pivot_y
```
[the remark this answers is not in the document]
Tysterman74: We're not changing the pivot point, that's set in the scene already. This is just programmatically setting the position to be in the center of the screen in order to not have the jarring image movement when we switch from the boot screen to the splash scene.

Tysterman74:
Pushed up changes to clean up the math of setting position since it got a bit confusing.

#Scenes/SceneScripts/SplashScene.gd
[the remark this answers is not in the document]
Turtyo: i see

Turtyo (dismissed):

#Scenes/SplashScene.tscn
```diff
@@ -0,0 +1,42 @@
...
+offset_right = 709.0
+scale = Vector2(0.682, 0.682)
+pivot_offset = Vector2(709, 1581)
+texture = ExtResource("1_7smrv")
+
+[node name="AudioStreamPlayer2D" type="AudioStreamPlayer2D" parent="."]
+stream = ExtResource("3_uytcb")
+volume_db = 3.043
+
+[node name="TextureRect2" type="TextureRect" parent="."]
```
Lann-hyl, line 36: Just a cleanup question, is this second TextureRect used somewhere? I see it is set to invisible and not used in the code so maybe it could be removed 
If it requires approval again we can just ignore it and I'll approve the PR

#Scenes/SceneScripts/SplashScene.gd
Tysterman74: Callback is to fire some function to call when we hit that section of the tween. In this case we call the _play_sfx() function after we do the initial scale. Similar to how signals work and firing a signal's callback when it's been emitted.

Lann-hyl (approved):

Turtyo (approved):

Issue #108, opened 2024-06-05
Better map event distribution


Turtyo:
## Is your amelioration related to a problem? Please describe.

The distribution of events is truly random currently (1/5 to get a certain event on any given room), which could end up with 4 heal / shop event in succession. This is clearly not what we want. it worked until now for testing, but we need to get closer to what it should actually feel like

## Describe the solution you'd like
We need a way to have a better control on what event spawn to what event. For example it shouldn't be possible to have a succession of shop/heal/shop or shop/shop or heal/heal for example.
In a way, what's in the room around has to influence the possibilities of what you get in the current room. I'm sure there exist plenty of algorithms to do exactly this kind of thing, but I have not researched any yet.

## Describe alternatives you've considered
A more simple alternative is simply have weighted probabilities on what rooms can appear, like 60% mob, 20% dialogue, 10% shop, 5% heal and 5% random.
But this causes more problem, for example it could end up with not having any shop or heal on the entire floor due to bad luck, which is probably something we want to avoid. We want something that looks random but has a set of rule to it (like for example at least 3 shops on the floor put in certain positions to ensure the player will always find at least one) 

## Additional context
The kind of maps we can have currently, only a single mob event in the first three floors, but two heals and 3 shops
![image](https://github.com/Saplings-Projects/1M_sub/assets/49365680/ab6b5e4c-f43b-4e8f-b6e9-7f7fe38e3427)

TripleCubes:
Get all room positions in the first 3 floors, shuffle it, the first 3 in array is heal, next 3 is shop, next 3 is ... you get the idea

And then, get the minimum number of room player need to go through to get through 3 floors (which is, 3), one of them must be heal

So like, for example, room number 2 that player go through will be guaranteed to be heal. It can be room 1 or 3 too, choosen randomly

The way we choose it is simple, get a list of steps [1, 2, 3], shuffle the list, the first index will be the heal room. If we want 2 heal rooms get the first two indexes

There is a problem tho, a forced event (like shop) can override a heal room. The solution for this is, or we can just leave it like that and add more heal rooms, or if a forced event overlap with a heal room, we place that heal room on a non-important room, like a battle room

Turtyo:
Shuffling like that might also result in unwanted distributions if you are unlucky, it doesn't change much from randomly selecting one for each room.
As I said, it will be better to check if an algorithm for distributions such as this already exists (which I'm pretty sure it does, considering how often this kind of stuff needs to be done in games).
[...]

PR #113, opened 2024-07-15
Better map event distribution

Lann-hyl:
# Description

Changes Map Event generation based on set probabilities instead of just randomness. 
There are also rules to prevent cases such as consecutive shop rooms, and floors where events are fixed (heal rooms before boss room).

## Related issue(s)

Issue: https://github.com/Saplings-Projects/1M_sub/issues/108

## List of changes

Room events are generated with the following probabilities:
- Mob: 45%
- Random: 16%
- Shop: 5%
- Heal: 12%
- Dialogue: 22%

The probabilities are set in `GLOBAL_VAR.gd`, and the implementation is in `MapManager.gd`

We then apply some generation rules that are inspired by this [blog](https://steamcommunity.com/sharedfiles/filedetails/?id=2830078257):
- Rule 1: No Heal rooms before half of the map
- Rule 2: No consecutive Heal and Shop rooms
- Rule 3: There must be at least 2 room types among destinations of rooms that have 2 or more Paths going out.
- Rule 4: No Heal rooms two floors before Boss (two before last floor)

Also some fixed rooms:
- Last room is set to Mob (placeholder to be replaced by Boss at some point)
- Rooms before boss are Heal rooms (before last floor)

## Tests

Tests are written in `test_map.gd`, they test that the 4 rules described earlier are applied correctly to the test map.

## Additional notes
Code in `MapManager.gd` that counts and prints the number of rooms for each event to verify if we get the number of rooms we want, with randomness. Set `DEBUG_PRINT_EVENT_COUNT` in `DEBUG_VAR.gd` to `True`.


Managers/MapManager.gd
```diff
@@ -51,27 +51,162 @@ func create_map(map_floors_width: Array[int] = map_width_array) -> MapBase:
...
+	## since they could be the only destinations of a room from the previous floor (happens when current floor's size [br]
+	## is smaller or equal than previous)
+	var _neighbors_events: Array[String] = []
+	var delta_xs: Array[int] = []
+
+	# If rightmost room of floor
+	if current_room.room_position.x == map.rooms[current_room.room_position.y].size() - 1 or map.rooms[current_room.room_position.y][current_room.room_position.x+1] == null:
+		delta_xs = [-1, 0]
+	else:
+		delta_xs = [-2, -1, 0]
```
Turtyo, line 113: i'm not sure i see why this is done
I understand why we need to check -2, -1 and 0
but if a room is the rightmost of the floor, someone on the floor below could still access -2, -1 and 0.
Basically, when the map is getting smaller, if you are not already on the rightmost side of the map you can choose between -2, -1 or 0 for the next floor. If you are already on the rightmost/leftmost side of the map, you only have a single room choice

Managers/MapManager.gd
```diff
@@ -51,27 +51,162 @@ func create_map(map_floors_width: Array[int] = map_width_array) -> MapBase:
...
+		
+	# Get the neighbor and current rooms types
+	for delta_x: int in delta_xs:
+		var _neighbor_x: int = current_room.room_position.x + delta_x
+		var _neighbor_y: int = current_room.room_position.y
```
Turtyo, line 118: this could be defined outside of the loop

Managers/MapManager.gd
```diff
@@ -51,27 +51,162 @@ func create_map(map_floors_width: Array[int] = map_width_array) -> MapBase:
...
+		sum += proba
+	var _rand_type_number: int = randi_range(0, sum-1)
+
+	for _rand_type_index: int in range(GlobalVar.EVENTS_CLASSIFICATION.size()):
+		if _rand_type_number < GlobalVar.EVENTS_PROBABILITIES[_rand_type_index]:
+			return GlobalVar.EVENTS_CLASSIFICATION[_rand_type_index].new()
+
+		_rand_type_number -= GlobalVar.EVENTS_PROBABILITIES[_rand_type_index]
+	
+	# This shouldn't be reached
```
Turtyo, line 157: use the `push_error` to show an error in the console, and return a monster event by default
better have an event we didn't want rather than no event at all in a reachable room

Managers/MapManager.gd
```diff
@@ -51,27 +51,162 @@ func create_map(map_floors_width: Array[int] = map_width_array) -> MapBase:
...
+			if index_height == map.rooms.size() - 2:
+				# Rooms before boss are Heal rooms
+				_current_room.room_event = EventHeal.new()
+
+			elif index_height == map.rooms.size() - 1:
+				# Boss room, TO BE ASSIGNED
+				_current_room.room_event = EventMob.new()
+
+			else:
+				while not is_room_event_correct(_current_room, map):
```
Turtyo, line 181: are we sure this will end ? i think testing how long this takes could be good
would it be possible to end with a map distribution such as for a given room, there will always be one of the rule that can't be upheld ?

Tests/test_map.gd
```diff
@@ -12,3 +12,125 @@ func test_map_gen() -> void:
...
+	var expected_exists_array: Array[Array] = [[0,2],[1,1],[1,2],[1,3],[2,0],[2,1],[2,2],[2,3],[2,4],[3,1],[3,2],[3,3],[4,2]]
+
+	for couple: Array[int] in expected_exists_array:
+		var _room_event: EventBase = test_map.rooms[couple[0]][couple[1]].room_event
+		assert_not_null(_room_event, "Expected not null event at %s but got %s" % [couple, _room_event])
+
+## Rule 1: No Heal room before half of the map
+func test_rule_1_map_gen() -> void:
+	var current_room: RoomBase
+	for index_height: int in range(int(test_map.rooms.size()/2.0)):
```
Turtyo, line 27: if you modify the `int` to a `ceil` / `floor` as I mentioned in the map generation, don't forget to change it here too

Tests/test_map.gd
```diff
@@ -12,3 +12,125 @@ func test_map_gen() -> void:
...
+			
+			# If current room is not a Heal or Shop room, ignore
+			if current_room == null or (not current_room.room_event.get_event_name() in no_consecutive_room_event):
+				continue
+			print(index_height, " ", index_width)
+			
+			# Get child room type and compare
+			for delta_x: int in [-1, 0, 1]:
+				var child_x: int = current_room.room_position.x + delta_x
+				var child_y: int = current_room.room_position.y + 1
```
Turtyo, line 55: could be defined outside the `delta_x` loop

Tests/test_map.gd
```diff
@@ -12,3 +12,125 @@ func test_map_gen() -> void:
...
+				if child_y >= 0 && child_y < test_map.rooms.size() && child_x >= 0 && child_x < test_map.rooms[child_y].size() && test_map.rooms[child_y][child_x] != null:
+					var consecutive: bool = (current_room.room_event.get_event_name() in no_consecutive_room_event &&
```
Turtyo, line 57: i'm not sure this first part is useful since at line 48 you already test that the room event of current room is no in the `no_consecutive_room_event`

Tests/test_map.gd
```diff
@@ -12,3 +12,125 @@ func test_map_gen() -> void:
...
+				var child_y: int = current_room.room_position.y + 1
+				if child_y >= 0 && child_y < test_map.rooms.size() && child_x >= 0 && child_x < test_map.rooms[child_y].size() && test_map.rooms[child_y][child_x] != null:
+					var consecutive: bool = (current_room.room_event.get_event_name() in no_consecutive_room_event && 
+											 current_room.room_event.get_event_name() == test_map.rooms[child_y][child_x].room_event.get_event_name())
+			
+					assert_false(consecutive, "Should not have two consecutive Heal rooms or two consecutive Shop rooms but found %s at %s and %s at %s" % 
+								 [current_room.room_event.get_event_name(), current_room.room_position, test_map.rooms[child_y][child_x].room_event.get_event_name(), test_map.rooms[child_y][child_x].room_position])
+
+
+## Rule 3: There must be at least 2 room types among destinations of Rooms that have 2 or more Paths going out.
```
Turtyo, line 64: According to what you wrote in the rule checking function of map generation, we also make sure the two leftmost and rightmost events are different, but here you only say that it counts only for rooms that have 2 or more paths going out, which one is it ?

Tests/test_map.gd
```diff
@@ -12,3 +12,125 @@ func test_map_gen() -> void:
...
+			unique = []
+
+			# Remove duplicates
+			for event in queue:
+				if not unique.has(event):
+					unique.append(event)
+			
+			assert_true(unique.size() != 1, "The two rightmost rooms of the floor should not have the same event but found %s at %s and %s at %s " % 
+					[test_map.rooms[rightmost_room_position.y][rightmost_room_position.x-1].room_event.get_event_name(), test_map.rooms[rightmost_room_position.y][rightmost_room_position.x-1].room_position, 
+					 test_map.rooms[rightmost_room_position.y][rightmost_room_position.x].room_event.get_event_name(), test_map.rooms[rightmost_room_position.y][rightmost_room_position.x].room_position])
```
Turtyo, line 121: isn't that already covered by the main checking loop ? when you get to the right, you end up on null rooms, so you stop appending but you pop the one at x-2, so you just check the two rightmost rooms at the end of the loop, no ?

Turtyo (changes requested):
A few things to change, but it's mainly details
The comments explain well what you are doing, that's great

Would it be possible to add a test to check the time it takes to generate a map ? Ran multiple times, to know the distribution of times, maybe 100 times ?

Managers/MapManager.gd
Lann-hyl: From my quick testing (just printing when a rule is not respected and the event is picked again), it usually picks a room after 0 to 2 attemps, so it does end pretty quickly, because except Rule 3, the rest of the rules apply mostly to Heal rooms. 
So technically, a whole map with only Mobs, Random and Dialogue upholds all rules (e.g. M,R,D,M,R,D,... repeat for each floor)
It's obviously not what we want, but is ensures that the generation does end.

Tests/test_map.gd
Lann-hyl: I've put a `continue` so it just skips when it reaches a null room and doesn't pop the queue, so the queue stays at size 3 until the end of the `for` loop. I'll check to be sure but I remember that's why I put that part

Managers/MapManager.gd
Turtyo: oh yeah mb i was only thinking of the rightmost room, but the room before that can only access two rooms indeed, not three
All good :+1:

Managers/MapManager.gd
Turtyo: does the % of each event stay stable / close to target after changing due to rules ?

Tests/test_map.gd
Turtyo: Putting the pop before the continue might solve this problem

Managers/MapManager.gd
Lann-hyl: image
image

We usually end up with less Shop and Heals rooms since those are the most restricted, resulting in more of the rest

Lann-hyl: Actually 2 shops out of 61 rooms is not a lot yeah, we could increase probabilities to adjust or something else

Turtyo: A few things:
- printing the target % next to the actual % could be nice to quickly know how far away we are
- we can adjust % ofc
- maybe you should add some boundaries on the actual % for each event (lower and upper bound), and generate the map again if the actual % are too far off from the target

Lann-hyl: With boundaries, won't we get a case where the map generation never ends from being very unlucky? It'll depend on how strict we are with boundaries ofc, I'll write the map generation time test.

Also i'm stupid, for shops 5% out of 61 is still 3 rooms, so it is close to what we want

Lann-hyl: image
image
image

Just some more generations, but yeah it does fluctuate quite a bit
I'll implement the changes you've pointed out and we can see again

Turtyo: it indeed seems to fluctuate a lot
To change that, maybe we can adjust the probabilities of a given type being chosen given the amount of each room type already select ? The actual math to do that might be a bit complex, but we might be able to eyeball it

Lann-hyl: We can try that, I get the idea but don't have the implementation in mind at all, it's a bit late rn 

variable probabilities based on what has been picked vs how much we expect for each event, no need of the exact number everytime we generate the map, while respecting the rules

Turtyo: I think you can do this without needing too much code changes (there will be code added though)
I'm thinking of making a copy of the original % array / dictionary, and modify the values inside it as we go
For how much should be removed / added i'm still trying of a good way to do it (just the difference between target and actual seems a bit too harsh, especially when there are few events selected at the start)

Lann-hyl: Could be a sort of priority system, a "what we need most".

Also small note: in the actual implementation, the probabilities don't need to be normalized or add to 100 for it to work. Just to say that if we increase only one value, it won't break. Normalizing is still better for understanding ofc

Turtyo: yes i saw that it doesn't need to sum up to 100, we can normalize, or we don't, it doesn't really matter since it's internal

Tomzkk (dismissed):
The generation and the rules seem good. I had to take time with some of them (3rd one specifically) but then realized it's probably because they are made for the specific map layout we decided on. 

If the distributions look good (which is being discussed in other message from turtyo) then it should be good

Global/GLOBAL_VAR.gd
```diff
@@ -13,8 +13,17 @@ var MODIFIER_KEYS: Dictionary = {
 # Keys can be accessed with the syntax: GlobalVar.MODIFIER_KEYS.ADD_PERMANENT
 # this is useful if we need to change the reference value in multiple places
 
-## A list of all the possible events
-var EVENTS_CLASSIFICATION: Array[Resource] = [EventMob, EventRandom, EventShop, EventHeal, EventDialogue]
+## A list of all the possible events, ordered the same as GlobalEnums.EventType
+var EVENTS_CLASSIFICATION: Array[Resource] = [EventRandom, EventHeal, EventMob, EventShop, EventDialogue]
```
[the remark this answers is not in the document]
Lann-hyl: Hm I thought it'd be better if things were consistent across classes, but I could change it to a Dictionary instead, so that it is linked without needing to keep the order

Managers/MapManager.gd
```diff
@@ -27,22 +27,25 @@ var current_map: MapBase
 ## - Width should not increase or decrease by more than 2 per floor (this makes it certain all rooms on the map are accessible [br]
 var map_width_array: Array[int]
 
+## Room events that we do not want to be consecutive
+const no_consecutive_room_event: Array[String] = ["shop", "heal"]
```
Turtyo, line 31: maybe an array of GlobalEnum.EventType would prevent mistyping the name in a string ?

Global/GLOBAL_VAR.gd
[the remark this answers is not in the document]
Turtyo: I'm thinking we might be able to remove this maybe :thinking: because we have the event types with the global enums already
That might require changing the places that need this array, so maybe out of scope for this PR;
But at the same time, since you are already working on events, might as well

Managers/MapManager.gd
Lann-hyl: I'm using strings here because I'm using the event name to know which event it is like this:
> `if current_room_event_name == "heal"`
> `if current_room_event_name in no_consecutive_room_event:`

So we could use the Class of the resource or GlobalEnum.EventType but I don't think we defined an GlobalEnum.EventType in the Event classes

Turtyo: Ah yes you would need to have a variable on each event indeed, because the class name is always a string :thinking: 
Mmh let's keep it like that then

Tests/test_map.gd
```diff
@@ -109,7 +109,6 @@ func test_rule_3_map_gen() -> void:
 					 test_map.rooms[index_height][index_width].room_event.get_event_name(), test_map.rooms[index_height][index_width].room_position])
 
 		# Check if the two rightmost rooms of the floor are the same type or not
-		queue.pop_front()
```
Turtyo: wasn't that needed to have correct size for the test ?

Turtyo (dismissed):

Tests/test_map.gd
Lann-hyl: Ah right you’re correct, i’ll revert it

Turtyo: it's even more surprising that the test is still passing even without it

Tomzkk (dismissed):

Managers/MapManager.gd
```diff
@@ -142,30 +146,35 @@ func is_room_event_correct(current_room: RoomBase, map: MapBase = current_map) -
...
 	
-	# This shouldn't be reached
-	push_error("Error: The random number _rand_number is bigger than the sum of probabilities _total_proba: %s > %s" % [_rand_number_error, _total_proba])
-	return EventMob.new()
+	for event_type: GlobalEnums.EventType in GlobalVar.EVENTS_PROBABILITIES:
+		var event_type_list: Array[GlobalEnums.EventType]
+		var event_count: int = floor(total_nb_rooms * GlobalVar.EVENTS_PROBABILITIES[event_type]/100)
+		# The floor before boss room is composed of heal rooms
+		#if event_type == GlobalEnums.EventType.Heal:
+			#event_count -= 3
```
Turtyo, line 163: why is this commented ? it seems to make sense to leave it, no ?

Turtyo (commented):
Seems clear, will just test it before approving

Turtyo:
What happens if we can't generate the map because of rules ? I don't see a retry of the generation

Managers/MapManager.gd
```diff
@@ -214,7 +234,7 @@ func _ready() -> void:
 					events[event] += 1
 					total_nb_rooms += 1
 			for k: String in events:
-				print("Event " + k + " has " + str(events[k]) + " rooms. (is " + str(float(events[k]) * 100 / total_nb_rooms).pad_decimals(2) + "%, expected: " + str(expected_probabilities[k]) + "%)")
+				print("Event " + k + " has " + str(events[k]) + " rooms (expected: "+ str(float(expected_probabilities[k] * 61.0/100)).pad_decimals(2) +"). (The percentage is " + str(float(events[k]) * 100 / total_nb_rooms).pad_decimals(2) + "%, expected: " + str(expected_probabilities[k]) + "%)")
```
Turtyo, line 237: ah i was also wondering, why multiply by `61.0/100` ?

Managers/MapManager.gd
Lann-hyl: ah good catch will remove it

Managers/MapManager.gd
Lann-hyl: the probability is not between 0 and 1 but 0 and 100, so it still needs to be divided by 100 to get the number of rooms we want

Managers/MapManager.gd
Lann-hyl: (misunderstood the question so I deleted my previous comment)

Ah right because while it makes sense, I saw from running that we'd usually get less heal rooms, since heal rooms are quite restricted by the rules. The expected number is 7 heal rooms, and I got 4 to 7 heal rooms on average, and commenting the lines gave 6 to 9 rooms. So it just kind of worked better by removing them, it gave a 'better' distribution I think

Managers/MapManager.gd
Turtyo: yes, for the divide by 100, but why multiply by 61 ?

Lann-hyl: oh right it was just hard coded, I will change it

Managers/MapManager.gd
Turtyo: oh then if it works better let's keep it like that, you can even remove the commented lines i think

Turtyo:
Once we have the map re-generation in case we it gets blocked, I think it'll be good for approval

Managers/MapManager.gd
```diff
@@ -195,13 +197,35 @@ func assign_events(map: MapBase = current_map) -> void:
...
+
+
+	if map_reset_flag:
+		# Reset all room events to null
+		for index_height: int in range(map.rooms.size()):
+			for index_width: int in range(map.rooms[index_height].size()):
+				if map.rooms[index_height][index_width] == null:
+					continue
+
+				map.rooms[index_height][index_width].room_event = null
```
Turtyo: instead of having this double break, can you not put the reset of room events after the `map_reset_counter >= map_reset_limit` check ?

Turtyo (commented):
Seems good, will just test that tomorrow, a thing to change regarding code structure I think, the double break with a flag is a bit strange

Managers/MapManager.gd
Lann-hyl: I thought of it but if we reset in the for loops, after the function call ends wouldn't it resume the for loops?

Turtyo: can you not do the reset then call the `generate_event` in the block ? that would recurse
alternatively, put all the room at null when you call "generate event" and just call `generate_event` if need to regenerate

Lann-hyl: I'm not sure to understand your reasoning, but the "problem" is the `generate_event` call itself:

image

If we call `generate_event` inside the for loops, doesn't it resume after the call ends like on this image? And if there are multiple recursive calls (which are very rare of course), it would then need to finish the previous loops. And with a bigger map, it could be slower

My point with the breaks are just to avoid that, like on this image:

image

Tell me if it's unclear.

Lann-hyl: image
Example of the recursive calls, I think it happens like that (?) if we don't break out of the for loops

Lann-hyl: Hm, while it doesn't become super slow or anything, I thought breaks would be better imo

Turtyo: solution: instead of a `break` in the for loop after the recursive call to `regenerate_map`, use a `return`. The problem in the current for loop is that since there are two levels loop, indeed a break would continue the outer loop, which wouldn't happen with a return. So this:
```GDScript
if map_reset_counter >= map_reset_limit:
	# Reset all room events to null
	for index_height: int in range(map.rooms.size()):
		for index_width: int in range(map.rooms[index_height].size()):
			if map.rooms[index_height][index_width] == null:
				continue

			map.rooms[index_height][index_width].room_event = null
	assign_event(map)
	return # since we return here, we do not continue the for loop we were in
```

Turtyo: I just thought the double break could lead to confusion later on

Lann-hyl: Ah right it's much simpler

Turtyo (approved):
Good job :+1:

Tomzkk (approved):
Looks good. Left one small comment but if you think it's okay to keep it there then we don't have to move it.

If the distributions appear good enough, then I think we are good to go with this.

Managers/MapManager.gd
```diff
@@ -51,27 +61,196 @@ func create_map(map_floors_width: Array[int] = map_width_array) -> MapBase:
...
-	map_width_array = [1, 3, 5, 7, 9, 11, 9, 7, 5, 3, 1]
-	current_map = create_map()
+	if not is_map_initialized():
+		map_width_array = [1, 3, 5, 7, 9, 11, 9, 7, 5, 3, 1]
+		current_map = create_map()
+
 
+		##### DEBUG #####
+		# Count number of each event
+		if DebugVar.DEBUG_PRINT_EVENT_COUNT:
```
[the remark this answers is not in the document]
Lann-hyl: You're right, I'll move it to a separate function

Tomzkk (approved):

Turtyo (approved):

Issue #111, opened 2024-06-25
Block mechanic


Turtyo:
## Feature Description

Implement the mechanic that allows a player to gain shield to block incoming enemy damage.

## Motivation

Since health is a resource that's difficult to come by, being able to block incoming damage every turn is important.

## Concept art / Designs

Show somewhere how much damage will be blocked.

## Acceptance Criteria

[Define the criteria that need to be met for this feature to be considered complete.]

- Enemy damage starts to deplete the block first, then the health
- Block is reset to 0 each turn
- If the damage is more than the block, the excess damage carries over to damage to health (if there is 5 block remaining and 7 damage, we take 2 health damage)

## Proposed Solution

We can probably copy the health component to do a block component, with a reset to 0 in a `on_turn_start` function.

JonaLam:
A lot of the damage effects in the game tend to use the 'modify_health' function, so we need to either change that or have something intercept it to make armor work. I think the best solution would be to either change the health component directly to have armot or add an armor component which the healthcomponent has a direct refrence to, and then do the calculations in the 'modefy_health' function.

As for the display, I think just a simple shield with a number would work fine, at least as a placeholder

Turtyo:
Actually, all the damage effects use `modify_health`.
I think having a block value in the health component and checking that when applying the logic of modify health would work fine; I think an even better solution is to leave `modify_health` unchanged (leave it to just do what's it meant to do).

Instead make a new function `modify_health_and_block`, which decides if it should modify health or modify block, then make a `modify_block` function similar to `modify_health`. Also need to add an `on_turn_start` to reset the block to 0.

I thought about a new class. From an organization standpoint, it would be better to have the block as a separate component; the problem I have with it is that we can't have reference to other component inside a given component (because then you lose the advantage of components, ie it being modular). So I think in the end having the block with the health still makes sense.

A simple shield with a number will do it for now, we can modify the asset later.

JonaLam:
I assume the function to be made should be in the health component, and in that case, maybe it should be called something more generic like `take_damage` as it will probably be what the damage functions call when they are going to do their thing.

 It would be possible to put the block component in it's own script which the health component always has a refrence to, so that we prevent the health component script from being too big, but that might be unnecesary complexity.

Turtyo:
`modify_health` makes more sense because it's not just for taking damage, but also to heal. Healing with a function called `take_damage` would be a bit strange.

I think that would be unnecessary complexity for two reasons:
[...]

PR #114, opened 2024-07-19
Block

JonaLam:
# Description

added block to the game

## Related issue(s)

#111 

## List of changes

Changed Health component to have block
Added a block display to all entities
Added block debug cards

## Tests

added some block tests to test_health


Entity/Components/HealthComponent.gd
```diff
@@ -7,32 +7,30 @@ class_name HealthComponent
...
+signal on_block_changed(new_block: int)
+
 ## The maximum health the entity can have; this is a hard limit
 @export var max_health: float = 100
 ## The health the entity currently has [br]
 ## Between 0 and max_health
 var current_health: float = 100 # ? change this to max health, or just unset since it's init by the ready
 
+
+var current_block: float = 0
```
Turtyo, line 19: this should be an `int` not a `float`
and actually all health should be `int`, not `float` but we never changed it since the start

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
-	if is_healing:
-		new_health = clampf(current_health + amount, 0, max_health)
-	else:
-		new_health = clampf(current_health - amount, 0, max_health)
-		
+	
+	var leftover_damage: float
+	leftover_damage = amount
+	
+	leftover_damage = block_damage(leftover_damage)
```
Turtyo, line 46: `var leftover_damage: float = block_damage(amount)`
is probably shorter and simpler

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	_health_damage(leftover_damage)
+
+## removes health without considering block [br]
+## primarily a helper function, but could be used in future for block ignoring attacks[br]
+func _health_damage(damage : int) -> void:
+	var new_health : int
+	new_health = current_health
+	
+	new_health = clampf(current_health - damage, 0, max_health)
```
Turtyo, line 55: `var new_health: int = clampf(current_health - damage, 0, max_health)` ?

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	
+	var leftover_damage: float
+	leftover_damage = amount
+	
+	leftover_damage = block_damage(leftover_damage)
+	_health_damage(leftover_damage)
+
+## removes health without considering block [br]
+## primarily a helper function, but could be used in future for block ignoring attacks[br]
+func _health_damage(damage : int) -> void:
```
Turtyo, line 51: I would add a check for negative damage here, don't want to heal with negative damage

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	var new_health : int
+	new_health = current_health
+	
+	new_health = clampf(current_health - damage, 0, max_health)
 	_set_health(new_health)
 
+## adds health to the unit[br]
+func heal(amount : int,  _caster : Entity) -> void:
```
Turtyo, line 59: adding a check for negative heal would be good here

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	var new_health : int
+	new_health = current_health
+	
+	new_health = clampf(current_health + amount, 0, max_health)
```
Turtyo, line 63: `var new_health: int = clampf(current_health + amount, 0, max_health)`

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
 func _set_health(new_health: float) -> void:
 	if (new_health == current_health):
 		return
 	current_health = new_health
 	on_health_changed.emit(current_health)
+
+## Removes block from the entity[br]
+func block_damage(damage: int) -> int:
+	var new_block : int
+	new_block = current_block
```
Turtyo, line 76: `var new_block: int = current_block`
no need for double line

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	
+	var leftover_damage : int
+	leftover_damage = max( damage - current_block, 0)
```
Turtyo, line 79: `var leftover_damage: int = max( damage - current_block, 0)`

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	
+	new_block = max(new_block - damage, 0)
```
Turtyo, line 81: `new_block = max(current_block - damage, 0)`
i know that technically `new_block = current_block` but I find it clearer that way, as the block as not been changed yet

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,67 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	_set_block(new_block)
+	
+	return leftover_damage
+
+## Adds block to the entity[br]
+func add_block(amount : int, _caster : Entity) -> void:
```
Turtyo, line 87: add a check for negative block values

Managers/PhaseManager.gd
```diff
@@ -12,6 +12,8 @@ signal on_phase_changed(new_phase: GlobalEnums.Phase, old_phase: GlobalEnums.Pha
 signal on_event_win
 ## When the player is dead (reduced to 0 health)
 signal on_defeat
+## When the phase is changed to PLAYER_ATTACKING
+signal on_turn_start
```
Turtyo, line 16: there is already `on_phase_changed` which gives you the current phase of the combat. This is basically recreating a sub-case of a signal that already exists, and it might be confusing in the long run

Tests/test_health.gd
```diff
@@ -6,23 +6,46 @@ func test_take_damage() -> void:
...
+	assert_eq(_player_health_component.current_block, amount)
+
+func test_take_damage_to_block() -> void:
+	var damage : float = 5
+	var block_amount: float = 10
+	var caster: Entity = _player
+	
+	_player_health_component.add_block(block_amount, caster)
+	_player_health_component.take_damage_block_and_health(damage, caster)
+	assert_eq(_player_health_component.current_block, 5)
```
Turtyo, line 34: i would also assert the total health is still 100

Turtyo (changes requested):
I'm not sure why but you seem to like multi-line value assignment, when it's not really necessary.
It follows what we discussed in the issue well, so that works nicely i think. i'll try to think of possible interactions with debuff that could make negative healing for example, as I remember a problem with this, but we might have added a test for it :thinking:

Entity/Components/HealthComponent.gd
JonaLam: Ye, i noticed that, but decided to just make most of the variables floats to keep it consistent. Should I just change all of them to ints?

Turtyo: if it doesn't break everything, then yes

Entity/Components/HealthComponent.gd
```diff
@@ -40,27 +40,29 @@ func take_damage_block_and_health(amount : float, caster: Entity) -> void:
 		caster = null
 
 	
-	var leftover_damage: float
-	leftover_damage = amount
-	
+	var leftover_damage: float = amount
 	leftover_damage = block_damage(leftover_damage)
```
Turtyo: why not just `var leftover_damage: float = block_damage(amount)` ?

Entity/Components/HealthComponent.gd
```diff
@@ -7,32 +7,30 @@ class_name HealthComponent
...
+
+
+var current_block: int = 0
 
 ## Initialize the health component, putting health to max health
 ## This only happens at game start for the player, the value is then tracked inside the [PlayerManager] singleton [br]
 ## For enemies, this is called when the scene is instantiated (or more accurately, when each enemy is instanciated) [br]
 func _ready() -> void:
 	_set_health(max_health)
+	PhaseManager.on_phase_changed.connect(reset_block)
```
Turtyo, line 26: `PhaseManager.on_phase_changed` has two arguments, `current_phase` and `old_phase` but `reset_block` doesn't take any argument
To work, it should ignore the `old_phase` and use the `new_phase` to check when it's the turn of the player

Turtyo (changes requested):

Tests/test_health.gd
```diff
@@ -35,6 +35,22 @@ func test_take_damage_to_block() -> void:
...
+	_player_health_component.add_block(block_amount, caster)
+	PhaseManager.on_phase_changed.emit(GlobalEnums.Phase.PLAYER_ATTACKING, GlobalEnums.Phase.ENEMY_ATTACKING)
+	assert_eq(_player_health_component.current_block, 0)
+
+func test_dont_reset_block_on_enemy_round_start() -> void: 
+	var block_amount: float = 10
+	var caster: Entity = _player
+	
+	_player_health_component.add_block(block_amount, caster)
+	PhaseManager.on_phase_changed.emit(null, GlobalEnums.Phase.PLAYER_ATTACKING)
```
Turtyo, line 51: maybe emit `GlobalEnums.Phase.ENEMY_ATTACKING` as the new phase ?

JonaLam: I did try that, but that did reset the block, which I doubt is a bug with the system, as it doesn't happen when you play the game properly. I think that maybe emiting the GlobalEnums.Phase.ENEMY_ATTACKING will lead to an enemy attack which wil then lead to the player turn starting before the assertion is done

Turtyo: that's possible indeed

Turtyo (approved):

Turtyo:
Great job on the block :+1: I don't have anything more to ask, actually that end was more me being picky about the types so we don't get conversions everywhere :laughing:

Entity/Components/HealthComponent.gd
```diff
@@ -7,32 +7,30 @@ class_name HealthComponent
...
-func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> void:
-	# Allow caster to be null, but not the target.
-	# If caster is null, we assume that the modification came from an unknown source,
-	# so status won't calculate.
-	var new_health: float
-
+## The intended way for entities to take damage.[br]
+## Removes block first and then deals excess damage to the health[br]
+func take_damage_block_and_health(amount : int, caster: Entity) -> void:
 	if amount <= 0.0:
```
Tomzkk, line 31: Should this still be 0.0 if amount is an int ?

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,69 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
-		new_health = clampf(current_health - amount, 0, max_health)
-		
+	
+	var leftover_damage: int = block_damage(amount)
+	_health_damage(leftover_damage)
+
+## removes health without considering block [br]
+## primarily a helper function, but could be used in future for block ignoring attacks[br]
+func _health_damage(damage : int) -> void:
+	if damage <= 0.0:
```
Tomzkk, line 49: same as above comment

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,69 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+		return
+	
+	var new_health : int = clamp(current_health - damage, 0, max_health)
+	
 	_set_health(new_health)
 
+## adds health to the unit[br]
+func heal(amount : int,  _caster : Entity) -> void:
+	if amount <= 0.0:
```
Tomzkk, line 58: same as above comment and the above above comment

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,69 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
-func _set_health(new_health: float) -> void:
+func _set_health(new_health: int) -> void:
 	if (new_health == current_health):
 		return
 	current_health = new_health
 	on_health_changed.emit(current_health)
+
+## Removes block from the entity[br]
+func block_damage(damage: int) -> int:
+	if damage <= 0.0:
```
Tomzkk, line 73: Should this still be 0.0 if amount is an int ?

Entity/Components/HealthComponent.gd
```diff
@@ -41,18 +39,69 @@ func modify_health(amount: float, caster: Entity, is_healing: bool = false) -> v
...
+	var leftover_damage : int = max(damage - current_block, 0)
+	
+	var new_block : int = max(current_block - damage, 0)
+	_set_block(new_block)
+	
+	return leftover_damage
+
+## Adds block to the entity[br]
+func add_block(amount : int, _caster : Entity) -> void:
+	if amount <= 0.0:
```
Tomzkk, line 85: Should this still be 0.0 if amount is an int ?

Entity/Components/HealthComponent.gd
```diff
@@ -7,32 +7,30 @@ class_name HealthComponent
...
+
+
+var current_block: int = 0
 
 ## Initialize the health component, putting health to max health
 ## This only happens at game start for the player, the value is then tracked inside the [PlayerManager] singleton [br]
 ## For enemies, this is called when the scene is instantiated (or more accurately, when each enemy is instanciated) [br]
 func _ready() -> void:
 	_set_health(max_health)
+	PhaseManager.on_phase_changed.connect(reset_block_on_round_start)
```
Tomzkk, line 26: How does GoDot decide what order to apply signals in ? 

For example, we have PhaseManager.on_phase_changed connected to this, but also to player start turn in battler. 
If a player has a buff that would increase his block on round start for example, would it be possible it happens before this reset and so it will then be wrongly reset to 0 ?

Tomzkk (changes requested):

Entity/Components/HealthComponent.gd
Turtyo: that's a good question, I'm not sure how Godot decides, and it indeed might be an issue. To solve the issue we could instead decide to reset the block on enemy turn finishing instead of player turn start.

Another question we could wonder is with effects that apply at turn start like poison. Is poison blocked by shield ? If yes, does poison trigger before or after the block reset, in which case is shield from the previous turn could block poison or not. Moving this to enemy turn end doesn't really solve the problem, because an enemy could die from poison thus finishing the fight before the player takes poison damage.

Need to check how signals are done, but my intuition tells me it's just "whatever component gets picked first", in which case we might actually need to create a new phase to handle those cases

Tomzkk: I was trying to find what ordering there is but wasn't really able to. 

I think most likely is that rather then it being random, what matters is the component's position in the tree. If that was the case (which I am not sure of) that would at least take the randomness away but it still doesn't resolve the problem. 

We would need some way to more clearly separate the specific cases. The only issue is I am not sure how many different cases that need to be resolved separated there are at the moment, and there might be more in the future ?

JonaLam: If we only need new signals, then I could just make a simple `before_phase_change` signal in the `PhaseManager`. It might be too simple, but anything more than that would probably be too complex for this PR, and we can do something more complicated if needed later

Turtyo: I think adding a phase shouldn't be too difficult (?) and would solve the problem

I'll open an issue to change the way we change phases so it's clearer later, but for now you will need to find where we call the change from phase enemy finishing to player attacking and insert a "remove block" phase between those two.

What do you think of this solution ?

Managers/PhaseManager.gd
```diff
@@ -13,6 +13,8 @@ signal on_event_win
 ## When the player is dead (reduced to 0 health)
 signal on_defeat
 
+## This is a signal to fix a bug with blocking regarding order of turn start effects, might be removed later
+signal before_phase_changed(new_phase: GlobalEnums.Phase, old_phase: GlobalEnums.Phase)
```
Turtyo, line 17: that seems like it might cause more problems later.
Implementing an actual new phase for block removal instead of playing with signal order in the same phase will be clearer and less error prone

JonaLam: I think this can be fine as a temp solution as long as we remove it when the phase system is changed. If I were to add more phases it would need a bit of reworking the battler, which I think is a bit out of scope, as well as the fact that it would be changed immediately in #118. This problem is a bit to complicated and I think it would be better if we leave it as is then wait for #118 to be finished, then we can revisit this. The only issue is that this could get very messy if we forget about it, but as long as we make a new issue, I'm sure we won't. Also, I could also change the new signal name, to make it clear that it's teporary so that no one uses it

Turtyo (approved):
I would say we merge as is and we fix the problem with order of block removal as quickly as possible with #118  and #120

Entity/Components/HealthComponent.gd
Tomzkk: Resolving this for now as there is temp solution + actual solution planned for future PR (https://github.com/Saplings-Projects/1M_sub/issues/120)

Tomzkk (approved):

Turtyo (approved):

PR #115, opened 2024-07-23
Game menu

Paper-2:
# Description

Creates a more accessible menu for player to interact and look at.
## Related issue(s)

https://github.com/Saplings-Projects/1M_sub/issues/80

## List of changes

- Arrow keys can interact with the menu buttons.
- Added a version tag and title placeholder
- Added a prototype ish buttons for the menu UI

## Tests

No test are needed

PR #116, opened 2024-07-23
Credit scene

### LINKED ISSUE #99: End  / credit screen
## Feature Description

Wow look at all the people that participated, wow !

## Motivation

We need to credit everyone properly.

## Concept art / Designs

image

## Proposed Solution

A new scene with an animation player. Order of the members is random, a few animations that are selected randomly too (the credit scene is never the same).

## Additional Context

Think of properly crediting the dialogue manager, see [this link](https://github.com/nathanhoad/godot_dialogue_manager/blob/main/docs/FAQ.md#how-do-i-credit-dialogue-manager-in-my-game)
Will also need to credit the [smooth scroll addon](https://github.com/SpyrexDE/SmoothScroll)

Turtyo:
# Description

Add a credit screen to show all contributors to the project.

## Related issue(s)

Closes #99 

## List of changes

- Credit screen with all name + roles of contributors, organized by category. Order inside categories is random
- Add a license scene, can be accessed through the options

## Tests

No new tests

## Additional notes

Did not make the final image that was shown in the issue as it did not look good. The credit scene is standalone for now, it needs to be triggered by something (either when we have the "Credits" button in the main menu from #115 or we have a clear end condition that triggers the end scene)

The license screen doesn't look very good now but it's functional at least, we can add better visuals once we have the art for it, which is clearly not the priority at the moment.


#Scenes/SceneScripts/credits.gd
```diff
@@ -0,0 +1,327 @@
...
+	# a little timer to start the scene
+	var timer: SceneTreeTimer = get_tree().create_timer(1)
+	var team_members_names: Array = _choose_member_order()
+
+	await timer.timeout
+	
+	# play first animation with animation player in
+	# this is needed since the loop also plays the out animation of the previous avatar
+	# but there is no previous avatar before the first one
+	var first_member: String = team_members_names[0]
```
JonaLam, line 143: Should probably change to `first_member_namer` or something to make it more clear what this variable is

#Scenes/SceneScripts/credits.gd
```diff
@@ -0,0 +1,327 @@
...
+	# member is now at the center of the screen, display their name + roles with a fade in of the text
+	name_role_label.text = first_member + "\n" + roles_string
+	var text_time_on_screen: float = _get_text_time_on_screen(name_role_label.text)
+	var tween: Tween = get_tree().create_tween()
+	tween.tween_property(name_role_label, "theme_override_colors/font_color", COLOR_FADE_IN, TEXT_FADE_IN_DURATION)
+	await get_tree().create_timer(text_time_on_screen).timeout
+	
+	var out_animation: String
+	
+	# now loop on all the remaining members
```
JonaLam, line 164: Is there a reason you do out anim first and then in anims?
The way you have it now means you kinda wrote the same code twice so I think it would be better if it would be possible to make in anims come first and outs come second and then remove the repeat code

JonaLam (changes requested):
Good job, Just the little thing with the `_animation_loop` function, other than that it seems good

#Scenes/SceneScripts/credits.gd
Turtyo: indeed

#Scenes/SceneScripts/credits.gd
Turtyo: I'll check the code, maybe there was a reason for the animation in itself

Turtyo: The reason we do out anim first then in anim is due the way things are handled in the animation, a little explanation:

We have two players, one that does the animation for the sprites going in, one for the sprites going out.

There is a sprite animated for the animation in, and another sprite animated for the animation out, even if it's the same image. Because if we use the same sprite to do in and out animation, then we can't have the same image going in and out at the same time. For example, if two "Cool" sapling follow each other, with a single sprite we would need to same sprite to both play the in and out animation at the same time, which is not possible, thus the separation.

Now doing it like this forces us to play the out animation before the in animation. What would happen if I played the in before the out:
1. I play the in, so I reset the position of the last sprite that got in (which was at the center of the screen) in order to play the animation of the next sprite that goes in
2. Since I reset the position, there is no more sprite at the center of the screen
3. No I play the out animation, the sprite for the out animation pops at the center of the screen

Note that steps 1-3 are done one after the other very fast. So maybe it wouldn't make a flicker on the sprite, but playing the out before doing a reset on the position of the in ensures that there is no flicker. I could invert it and check that it still works, but on slower computer that might not be the case.

#Scenes/SceneScripts/credits.gd
Turtyo: done in [f3edd60](https://github.com/Saplings-Projects/1M_sub/pull/116/commits/f3edd60c74d4a008653584ea998376a07bd3c904)

#Scenes/SceneScripts/credits.gd
JonaLam: I don't think you need to change it then. It's only a small thing, and the system is so isolated that I think it probably fine to leave it like it is.

JonaLam (approved):

#Scenes/SceneScripts/credits.gd
Turtyo: yeah, it probably won't be changed a lot, it doesn't interact with outside stuff anyway
there might be a way to change the code to make it work, but i would have to check

Tomzkk (dismissed):
Looks good - is there any reason why we don't link it to the credits button in the main menu for now though ?

Turtyo:
There isn't a reason to not link it, I thought I linked it though :thinking:

Tomzkk (approved):

Issue #48, opened 2024-01-20
Save system


Turtyo:
## Feature Description

Let's be merciful to the player and not force them to win the game in a single play session ! Let's make a save system !

## Motivation

No data is being stored currently anywhere, this save system would help with that.

## Acceptance Criteria

- All relevant data is being stored in an external file
- This file is auto-loaded at game start if it exists
- Auto-save between each room ?

## Proposed Solution

We have different managers, we save the data from all those managers (for example we save the player HP and stats from the `PlayerManager`). Later on we will probably have more managers, like a map manager, we will add those to the save too.

## Additional Context

This system would need to be updated every time a change is made to the available info in the managers, or if a new manager appears.

cheesefrycook:
This seems good, I think autosaving between each room is a good idea. 

Personally I think we should not save progress during an encounter. If you quit the game in the middle of a fight, the fight will restart when you re-enter the game. This is so we don’t have to save all the enemy data too. 

Only problem is the player could quit the fight if they’re about to die. But we could potentially save the player data when the game is closed so there’s no benefit to doing that

Turtyo:
Well if they want to save scum, so be it...
Question for the auto-save, when do we do it ?
For now, we probably auto-save after combat end (so we need #35 probably).
But ultimately, when would we save when we implement the map, we can:
- save when the player does a click to go to the next room
- save when event in current room is done but player has not chosen next room yet
- save when player enters the next room (which could be equivalent to the first option, but with option 1 you could potentially reload when entering a random event to get another random event)

If we do it with an auto-save, do we even provide a save button ?

cheesefrycook:
I'd say we save both when:
- The player first enters an event
- The player enters the map screen

So this means the player could quit the game if they're in an event or on the map screen. For this, we will have to save where the player was when they quit the game (either map or event) so we can reload that scene when the game is opened again.

If the player quits in the middle of an event, then data from that event will not be saved. The event will be restarted when the game is opened again.

If we have auto-save then a save button isn't needed I think.

Turtyo:
The map screen would after the reward screen right ?

cheesefrycook:
Yeah that's what I was thinking

Turtyo:
Did you see the comment I left in #45 about using auto-loaded singleton to hold user data instead of having a main scene load all the other scenes ? I think it might be a better approach to the problem actually, as the Godot doc also says it is preferable to do this way.
[...]

PR #117, opened 2024-07-25
Issue-48: Save Functionality

Tysterman74:
# Description

Save Functionality when we're in the middle of a game session and we exit out and go back in to the game again.

## Related issue(s)

Related to #48 

## List of changes

- Added Inventory Icon to toggle inventory to track saving items
- Add logic on Main Menu for continue and start game buttons to load and clear data
- Add SaveManager, has signals for save and load signals
- Add listening for signals in CardManager, InventoryManager, MapManager, PlayerManager, SceneManager, XpManager to save and load their data
- Add executing save when entering all scenes
- Set default torch count to 3, update inventory tests

## Tests

N/A

## Additional notes

N/A


Managers/SceneManager.gd
```diff
@@ -75,3 +79,38 @@ func on_defeat() -> void:
 ## Give rewards, allow player to move on the map
 func on_event_win() -> void:
 	PlayerManager.player_room.room_event.on_event_ended()
+
+func _save_scene_data() -> void:
+	for children: Node in current_scene.get_children():
+		children.owner = current_scene
```
Turtyo, line 85: what is the goal of this ?

Managers/SceneManager.gd
```diff
@@ -75,3 +79,38 @@ func on_defeat() -> void:
...
+	if error:
+		print("Error saving player data: ", error)
+
+func load_scene_data() -> void:
+	call_deferred("_deferred_load_current_scene_from_data")
+
+func _deferred_load_current_scene_from_data() -> void:
+	var config_file: ConfigFile = SaveManager.load_config_file()
+	if config_file == null:
+		return
```
Turtyo, line 103: should there be a warning for this ? is it possible to normally call this with a null save file ?

Managers/SaveManager.gd
```diff
@@ -0,0 +1,31 @@
+extends Node
+
+signal start_save
+signal start_load
+
+var config_file: ConfigFile
```
Turtyo, line 6: should it be called `save_file` instead ?

Managers/InventoryManager.gd
```diff
@@ -40,3 +50,23 @@ func reset_inventory() -> void:
...
+func _save_inventory() -> void:
+	var config_file: ConfigFile = SaveManager.config_file
+	config_file.set_value("InventoryManager", "gold_component", gold_component)
+	config_file.set_value("InventoryManager", "torch_component", torch_component)
+	config_file.set_value("InventoryManager", "consumable_component", consumable_component)
+	config_file.set_value("InventoryManager", "relic_component", relic_component)
+	
+	var error: Error = config_file.save("user://save_data.ini")
+	if error:
+		print("Error saving inventory data: ", error)
```
Turtyo, line 63: `push_error`

Managers/CardManager.gd
```diff
@@ -44,3 +47,21 @@ func is_discard_hand_signal_connected(callable: Callable) -> bool:
 func connect_discard_hand_signal(callable: Callable) -> void:
 	card_container.on_finished_discarding_hand.connect(callable)
 
+func save_data() -> void:
+	var config_file: ConfigFile = SaveManager.config_file
+	config_file.set_value("CardManager", "current_deck", current_deck)
+	
+	var error: Error = config_file.save("user://save_data.ini")
+	if error:
+		print("Error saving player data: ", error)
```
Turtyo, line 56: `push_error`

#Scenes/SceneScripts/MainMenu.gd
```diff
@@ -1,20 +1,38 @@
...
-	if not PlayerManager.is_player_initial_position_set:
-		SceneManager.goto_scene("res://#Scenes/MapUI.tscn")
-	else:
-		# TODO : load the last scene the player was in
-		pass
+	SaveManager.clear_data()
+	PlayerManager.clear_data()
+	MapManager.clear_map_data()
+	XpManager.clear_data()
+	CardManager.clear_data()
```
Turtyo, line 24: should they be named `init_data` instead of `clear_data` ? clear makes it sound like we are deleting things, when we are creating new things

Also why is the map the only one with a special name ?

Managers/PlayerManager.gd
```diff
@@ -15,7 +15,8 @@ var player: Player
 var player_position: Vector2i = Vector2i(-1,-1):
 	set(position):
 		player_position = position
-		is_player_initial_position_set = true
+		if player_position.x != -1 and player_position.y != -1:
```
Turtyo, line 18: is it possible to do `if player_position != (-1, -1)` ?

Managers/PlayerManager.gd
```diff
@@ -42,9 +43,34 @@ var player_persistent_data: PlayerPersistentData = null:
...
 
+func _save_player() -> void:
+	var config_file: ConfigFile = SaveManager.config_file
+	config_file.set_value("Player", "position", player_position)
+	config_file.set_value("Player", "player_room", player_room)
+	config_file.set_value("Player", "player_persistent_data", player_persistent_data)
+	
+	var error: Error = config_file.save("user://save_data.ini")
+	if error:
+		print("Error saving player data: ", error)
```
Turtyo, line 59: `push_error`

Turtyo (changes requested):
I would also put the save file in their own directory instead of putting everything at the root

#Scenes/SceneScripts/MainMenu.gd
```diff
@@ -1,20 +1,38 @@
...
+	SceneManager.goto_scene("res://#Scenes/MapUI.tscn")
+
+func _on_continue_pressed() -> void:
+	if PlayerManager.has_saved_data():
+		PlayerManager.load_player()
+		MapManager.load_map_data()
+		XpManager.load_data()
+		CardManager.load_data()
+		InventoryManager.load_inventory()
+		SceneManager.load_scene_data()
```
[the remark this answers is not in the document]
Tysterman74: In the _ready() function, I disable the continue button if we don't have save data. If that suffices then we can resolve.

Managers/SceneManager.gd
Tysterman74: Was reading a thread on saving a PackedScene and it mentioned to set the children's owner to the current scene. However, I removed this for loop and it seems to still work, so I'm removing it!

Managers/SceneManager.gd
Tysterman74: It technically isn't where we are calling it from we already do a check on config_file. I'll add a warning for it.

Managers/SaveManager.gd
Tysterman74: Good idea. I named it config_file since it's the type, but save_file is better readability for us.

#Scenes/SceneScripts/MainMenu.gd
Tysterman74: I can go either way, I'll rename it to init_data, and I just have bad naming convention sense.

Managers/PlayerManager.gd
Tysterman74: It works! Will refactor

#Scenes/SceneScripts/MainMenu.gd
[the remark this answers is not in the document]
Turtyo: Is the disabled button hidden ? Or can you just not click on it ?

Managers/SceneManager.gd
Turtyo: alright
if you still remember where you read that I would be curious to see, if you have a link

#Scenes/SceneScripts/MainMenu.gd
Turtyo: I think everyone has bad naming convention sense x)

#Scenes/SceneScripts/MainMenu.gd
```diff
@@ -17,13 +18,11 @@ func _ready() -> void:
...
-	MapManager.clear_map_data()
-	XpManager.clear_data()
-	CardManager.clear_data()
-	InventoryManager.reset_inventory()
-	SceneManager.goto_scene("res://#Scenes/MapUI.tscn")
+	if PlayerManager.has_saved_data():
+		start_new_game_dialog.show()
+	else:
+		MapManager.init_data()
+		SceneManager.goto_scene("res://#Scenes/MapUI.tscn")
```
Turtyo, line 25: previously there was an init for SaveManager, PlayerManager, Map, XP, Card, Inventory and Scene when you start a new game. I see it's still being done when pressing the confirm, but in the case where you don't have anything saved and just start normally, shouldn't everything be init ?

Managers/SceneManager.gd
```diff
@@ -80,29 +79,26 @@ func on_defeat() -> void:
...
 
 func load_scene_data() -> void:
 	call_deferred("_deferred_load_current_scene_from_data")
 
 func _deferred_load_current_scene_from_data() -> void:
-	var config_file: ConfigFile = SaveManager.load_config_file()
-	if config_file == null:
-		return
+	var save_file: ConfigFile = SaveManager.load_save_file()
+	assert(save_file != null, "Attempting to load a scene with no config file. How did you get here???")
```
Turtyo, line 99: `assert` does not stay in non debug build, see [this page](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#assert-keyword)
It would be better to keep the check with `==` and use `push_error`

Managers/CardManager.gd
```diff
@@ -48,7 +48,7 @@ func save_data() -> void:
 	var save_file: ConfigFile = SaveManager.save_file
 	save_file.set_value("CardManager", "current_deck", current_deck)
 	
-	var error: Error = save_file.save("user://save_data.ini")
+	var error: Error = save_file.save("user://data/save_data.ini")
```
Turtyo, line 51: `data` can be a lot of thing, I would name the folder `save` instead

Turtyo (changes requested):

#Scenes/SceneScripts/MainMenu.gd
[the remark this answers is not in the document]
Tysterman74: We're just not able to click on it. We would need an asset of it being disabled to fully see the change.

#Scenes/SceneScripts/MainMenu.gd
Tysterman74: Hmm yeah good point, I originally thought that we would hit this block already initialized, but it doesn't hurt to include the init here as well.

#Scenes/SceneScripts/MainMenu.gd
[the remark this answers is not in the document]
Turtyo: Can the sprite for the disabled state just be a transparent ?

Managers/SceneManager.gd
```diff
@@ -82,27 +82,28 @@ func on_event_win() -> void:
...
+		push_error("Error saving player data: ", error)
 
 func load_scene_data() -> void:
 	call_deferred("_deferred_load_current_scene_from_data")
 
 func _deferred_load_current_scene_from_data() -> void:
 	var save_file: ConfigFile = SaveManager.load_save_file()
-	assert(save_file != null, "Attempting to load a scene with no config file. How did you get here???")
+	if save_file == null:
+		push_error("Attempting to load a scene with no config file. How did you get here???")
```
Turtyo, line 100: I'm nitpicking, but shouldn't it be "Attempting to load a scene with no **save** file [...]" ?

Turtyo:
Few errors I got from playing the game:
- loading a combat, going back to main menu, then continue leads to this error:
image
And after testing a bit more, it's actually on any continue

- not exactly sure of the reason, but sometimes I get this error when using the continue:
image

I know we've talked about not saving inside events (meaning leaving the game and coming back let's you redo the event). However, for the combat it seems a bit extreme. How feasible would it be to save current HP and status of all entities in a combat when pressing the "go back to main menu" ?

#Scenes/SceneScripts/MainMenu.gd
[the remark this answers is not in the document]
Tysterman74: Uhh sure, I just increased the transparency on the new asset. We should probably ask for a better image later.

Tysterman74:
> Few errors I got from playing the game:
> 
>     * loading a combat, going back to main menu, then continue leads to this error:
>       image
>       And after testing a bit more, it's actually on any continue
> 
>     * not exactly sure of the reason, but sometimes I get this error when using the continue:
>       image
> 
> 
> I know we've talked about not saving inside events (meaning leaving the game and coming back let's you redo the event). However, for the combat it seems a bit extreme. How feasible would it be to save current HP and status of all entities in a combat when pressing the "go back to main menu" ?

First image looks to be some data that hasn't been created and we're trying to access it. I believe player_persistent_data is only created once we enter a Mob room, so going through other event rooms won't create it. I'll add some checks on all of the loading to see if the Key and Section are stored before attempting to load it.

Second image, the error code looks to be FILE_NOT_FOUND... basically saying the ConfigFile object doesn't exist. Will need to check.

Let me look into saving while in battle.

Tysterman74:
> > Few errors I got from playing the game:
> > ```
> > * loading a combat, going back to main menu, then continue leads to this error:
> >   image
> >   And after testing a bit more, it's actually on any continue
> > 
> > * not exactly sure of the reason, but sometimes I get this error when using the continue:
> >   image
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I know we've talked about not saving inside events (meaning leaving the game and coming back let's you redo the event). However, for the combat it seems a bit extreme. How feasible would it be to save current HP and status of all entities in a combat when pressing the "go back to main menu" ?
> 
> First image looks to be some data that hasn't been created and we're trying to access it. I believe player_persistent_data is only created once we enter a Mob room, so going through other event rooms won't create it. I'll add some checks on all of the loading to see if the Key and Section are stored before attempting to load it.
> 
> Second image, the error code looks to be FILE_NOT_FOUND... basically saying the ConfigFile object doesn't exist. Will need to check.
> 
> Let me look into saving while in battle.

Okay saving during a battle is turning out to be a bit more of an effort than I thought, not sure if we want to address it in another issue just to get this save mechanic working.

Turtyo:
If it's too much effort we can tackle it later as an enhancement of the save system

Turtyo (commented):
Instead of checking for each event, could we create the player persistent data at the start ? Or are we missing information ?

Tysterman74:
> Instead of checking for each event, could we create the player persistent data at the start ? Or are we missing information ?

Player Persistent Data can't be created until the Player object is created, which is only created in the Mob scene

#Scenes/SceneScripts/MainMenu.gd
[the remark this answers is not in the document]
Tysterman74: I think the best option here is to have another art asset in a disabled state so we can set the button disabled texture as that. I'll leave my asset in for now with the hopes we'll update it.

Tysterman74:
> If it's too much effort we can tackle it later as an enhancement of the save system

So diving into this more, we're able to save Player stats right now because of Player Persistent Data, but we'll need to have something similar for the Enemy as well. Something to note when we enhance it.

Turtyo:
The reason for the PlayerPersistentData was that the player object is destroyed between scenes, and we need to keep the information.
But the enemies are unique to each encounter (they don't carry over between events), so having EnemyPersistentData doesn't make sense. We could call it EnemyState instead, but it would be destroyed between each fight so I'm not really sure it's really useful.

Can we not have a section to save enemy state in case we decide to quit inside a combat instead ? Not right now ofc, for when we do it.

Turtyo (changes requested):
Checking the game, I think there was a misunderstanding with what I meant when I asked if we could make the continue button transparent when there was no save data.

Currently, the button is not invisible if you hover it in the menu (and it's actually even visible if you don't hover it). The solution I had in mind was to use `modulate` (see [this post](https://forum.godotengine.org/t/how-to-change-the-opacity-of-a-sprite/23520)).
You can basically do `button.modulate.a = 0` and the button becomes entirely transparent.

Because right now it looks like this, and it's visible even without hovering
image
Hovering clearly shows the button
image

Turtyo:
image

Also this error when starting a new game without any previous save file. Probably due to the fact we try to check if there is save data in the save file, but since we don't even have a save file it can't do it. Solution is to check if a save file exists before trying to check what's inside the save file (if the error is what I wrote that is)

Managers/SaveManager.gd
```diff
@@ -18,13 +18,16 @@ func clear_data() -> void:
...
+		if !DirAccess.dir_exists_absolute("user://save/"):
+			DirAccess.make_dir_absolute("user://save/")
+		
+		if error:
+			push_error("Error loading save_data ", error)
+			return null
+		
+		return save_file
+	else:
 		return null
```
Turtyo, line 33: a little `push_warning` maybe ? is there a situation where we would try to load a save file without the save file existing ?

Turtyo: nvm I saw we were always checking for null in the different sections that need to save to the file, maybe it's fine without the `push_warning`

Turtyo (approved):

Tysterman74:
> The reason for the PlayerPersistentData was that the player object is destroyed between scenes, and we need to keep the information. But the enemies are unique to each encounter (they don't carry over between events), so having EnemyPersistentData doesn't make sense. We could call it EnemyState instead, but it would be destroyed between each fight so I'm not really sure it's really useful.
> 
> Can we not have a section to save enemy state in case we decide to quit inside a combat instead ? Not right now ofc, for when we do it.

Yeah, I think that can work. We can utilize the EnemyManager for saving and loading. Not sure if EnemyState would be the right class to describe it, but at the end of the day we need some sort of data class to be able to save the enemy data in-between sessions and make sure it gets cleared out after battles and reset again when another battle begins.

Turtyo:
I think we should do that `EnemyState` later on, better get this done first and work to improve it later

Managers/PlayerManager.gd
```diff
@@ -43,8 +44,35 @@ func _ready() -> void:
...
+	if save_file.has_section_key("Player", "player_room"):
+		player_room = save_file.get_value("Player", "player_room")
+	
+	if save_file.has_section_key("Player", "player_persistent_data"):
+		player_persistent_data = save_file.get_value("Player", "player_persistent_data") as PlayerPersistentData
 
 
+func has_saved_data() -> bool:
+	var save_file: ConfigFile = SaveManager.load_save_file()
+	return save_file != null
```
Turtyo, line 75: This should be in the `SaveManager`

Managers/SaveManager.gd
```diff
@@ -0,0 +1,33 @@
+extends Node
+
+var save_file: ConfigFile
+
+func _ready() -> void:
```
Turtyo, line 5: we should create the `save` directory inside the `ready` function

Turtyo (changes requested):

Managers/PlayerManager.gd
```diff
@@ -54,7 +82,12 @@ func set_player(in_player: Player) -> void:
 
 func create_persistent_data() -> void:
 	player_persistent_data = PlayerPersistentData.new()
-	
+
+func init_data() -> void:
+	player = null
+	is_player_initial_position_set = false
+	player_position = Vector2i(-1,-1)
+	player_room = null
```
Lann-hyl, line 90: Need to put `is_map_movement_allowed = true` here to reinitialize the flag and allow movement whenever the game is recreated

Lann-hyl (changes requested):
Issues as we discussed:
- Need to create the `save` folder in code or saving does not work
- Cannot move when creating a second new game, because movement is blocked from quitting the previous game. Need to put `is_map_movement_allowed = true` in PlayerManager

Managers/PlayerManager.gd
Turtyo: this solves the issue for me too

Managers/SaveManager.gd
```diff
@@ -4,6 +4,8 @@ var save_file: ConfigFile
 
 func _ready() -> void:
 	save_file = ConfigFile.new()
+	if !DirAccess.dir_exists_absolute("user://save/"):
+		DirAccess.make_dir_absolute("user://save/")
```
Turtyo, line 8: this is not recursive so if we ever want a more nested structure we'll need to think of using `make_dir_recursive` instead. I mean, we could directly use it here, but idk if it's really useful

Turtyo (approved):

Turtyo:
The previously mentioned problems seem to not be there anymore, so that's nice

Lann-hyl (commented):
Opening the inventory then going back to main menu leaves the inventory window opened, might be a small issue to be fixed

Otherwise it's all good for me

Turtyo:
That might be due to the inventory not being attached to the tree of the scene; this is not a problem related to the save PR, so I think opening an issue for it might be better

Issue #120, opened 2024-08-06
Fix block to use new phases


JonaLam:
## Is your amelioration related to a problem? Please describe.

The way block is removed at the start of the turn is with a temporary signal in the phase manager.
That signal should be removed and the block should be removed in a newly implemented remove block phase

## Additional context
#114 needs to be done
#118 needs to be done

Turtyo:
I would consider that a bug since it doesn't behave as we would like it

Turtyo:
Once #124 is merged, I think the plan would be to add two block remove phases (one for player, one for enemies). When on this phase, we call the `block_reset` of `health_component`

Or we can do it via signal, but it's harder to debug than direct function calls I think (because with direct calls you have a clear stack trace)

Turtyo:
Okay, now that #124 is merged, I'm thinking of this solution:
- Add two new phases: `reset_block_ally` and `reset_block_enemy`
- In the battler:
    - `reset_block_ally`: reset the block of the player by  calling `block_reset` for the `health_component` of the player
    - `reset_block_enemy`: reset the block of all the enemies in the same way
After each phase we call the `advance_to_next_combat_phase` to go to next phase
I'm thinking of putting the reset for ally just before the player turn (so effects like poison still damage the player). And in a similar fashion for the enemies

I'm also thinking that we might want to separate the effects of start and end of turn into their own phase. This way if later we want to move stuff around, it's easier. That way, the player phase is reserved for player interaction with cards, and everything that is automatic (effects triggering) is in its own phase

Later if we decide we want the block to last until after the turn start effects (thus meaning we can block poison for example), we just need to exchange the position of the remove block and the effect of the turn start (which I don't think we can do currently, as everything is in the player turn for that matter)

PR #124, opened 2024-08-12
A new way to change phase

### LINKED ISSUE #118: A new way to change phase

The way we change phase actually is to call the next phase by name when arriving at the end of the current phase. This makes it error prone because if we somehow write the wrong phase, tracking in the entire codebase where we called the wrong phase won't easy.

Instead of calling the next phase by name, create a `advance_to_next_phase` function in the phase manager. At the end of the code of a phase, simply call this `advance_to_next_phase` function. This function changes phases in the order the phases are written in the `Phase` enum in `GlobalEnums`
Two advantages to this:
- no more error possible with calling the wrong phase (or if there is an error, it's in a single place)
- if later we want to change the order of the phases, it's simply a matter of rewriting a single enum (much simpler than to track down each phase change in the entire code)

## Additional context
This might need to wait for #114 to be finished since we are discussing adding phases in there.
Additionally, `set_phase` would not be used directly anymore by other classes outside of the `PhaseManager`

Turtyo:
# Description

The phase manager now is the one doing the phase changes for combat phase, instead of the phase being set from outside. This allows (in theory) to add new phases to the combat phase and the phase manager will automatically go through those without need for more code.

You still need to call `advance_to_next_combat_phase` at the correct places, but you don't need to worry about what the next phase actually is.

## Related issue(s)
Closes #118 

## List of changes

- Split the `Phase` enum into `GlobalPhase` and `CombatPhase`
- Rewrite `PhaseManager` to go to next combat phase by calling `advance_to_next_combat_phase`
- Replace all instances of `Phase` by either `CombatPhase` or `GlobalPhase`
- Replace all instances of `set_phase` by either `advance_to_next_combat_phase` or `set_global_phase`

## Tests

Updated to work with the new phase system.

## Additional notes

This will help solving #120

### DIFF
--- a/#Scenes/Events/mob/0.tscn
+++ b/#Scenes/Events/mob/0.tscn
@@ -14,7 +14,7 @@
 [ext_resource type="Texture2D" uid="uid://d4muqvs3etnr8" path="res://Art/Card_layout/discard_pile.png" id="12_0nlrw"]
 [ext_resource type="Script" path="res://UI/DiscardPileUISetter.gd" id="13_8nsar"]
 [ext_resource type="PackedScene" uid="uid://bam77cwf4emyr" path="res://#Scenes/TopBarOverlay.tscn" id="14_os5i4"]
-[ext_resource type="PackedScene" path="res://#Scenes/xp_bar.tscn" id="15_r7cr8"]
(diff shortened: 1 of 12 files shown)


JonaLam (approved):
Good job
It seems clean and easy to use so I have nothing to complain about

Tomzkk (approved):
👍

PR #125, opened 2024-08-18
 Fix block to use new phases

### LINKED ISSUE #120: Fix block to use new phases

Turtyo:
# Description

Separate the block removal into its own phase, using the new phase system.

## Related issue(s)

Closes #120 

## List of changes

- Create two new phase to remove block for player and enemies
- Create two other phase to handle turn start effects (such as poison)

## Tests

Changed the tests to make them work

## Additional notes

Would like @JonaLam 's approval for this if possible


Tomzkk (approved):
Looks good

Issue #122, opened 2024-08-10
Soft-locking game when exiting room selection at game start


Turtyo:
## Describe the bug
When escaping the map at game start, it results in a screen with nothing on it.

## To Reproduce
1. Start a new game
2. This opens the map, do not click on anything and just press "escape" on your keyboard
3. This closes the map
4. Now the screen is just grey, with nothing to do

## Expected behavior
No soft-lock possible from starting the game

## Possible solution
Either of those would work
1. Disable escaping the map at the start of the game
2. Add the overlay with map button

Preference for solution 1, because there is no reason for a player to escape the map upon game start

Turtyo:
@Biosquid I see you took this issue a week but haven't actually discussed how you plan to do it or approve the issue.
Which of the two solution I described would you go for ? Do you have another solution perhaps ?

Biosquid:
I was planning on going w/ the first. It seems easier to implement and has the lowest chance of causing any other issues in the future as long as we find a good spot to re-enable it.

Turtyo:
You can probably re-enable it when the player clicks a room. i'm not too sure how to disable a key mapping, you'll have to see that for yourself

Biosquid:
@Turtyo I went over this issue w/ @BishopJohnson and we got it fixed and I pushed it to my fork.

Bishop also recommended that we change the queue_free() method to be _on_return_button_pressed() instead as it can be better in the long run if we want to go somewhere specific when escape is hit (i might be wording it badly but i think thats the general idea). What do you think about this?

Biosquid:
I should note that I fixed it by just not allowing the method to run if map is the only scene, so it'll work if we change things in the future to go to the map only view after each fight/event/shop/whatever

Turtyo:
Which `queue_free` are you referring to ? There are multiple inside the code, which file are you talking about ?

Regarding what you did, it would be easier if you made a pull request, as it's easier to show what you did instead of saying what you did. If you think the PR is not ready for merge, open it as a draft PR

PR #133, opened 2024-09-25
Fixed softlock from trying to return to main menu on first map selection by not letting the function go through if its the only scene.

Biosquid:
# Description

All this changed was a single line in an if statement adding another condition. if it sees that the parent scene is root then it doesn't allow for queue_free() to run.

## Related issue(s)

https://github.com/Saplings-Projects/1M_sub/issues/122 

## List of changes

Can be found in the the file: MapUI on line 34

## Tests

No test scripts written

### DIFF
--- a/UI/MapUI.gd
+++ b/UI/MapUI.gd
@@ -32,7 +32,7 @@ var current_player_room: RoomUI
 var light_overlay: LightOverlay
 
 func _input(_inputevent: InputEvent) -> void:
-	if (_inputevent.is_action_pressed("cancel_action")):
+	if (_inputevent.is_action_pressed("cancel_action") and get_parent() != get_tree().root):
 		queue_free()
 
 


Turtyo:
Are we sure the only case where the parent of the map is the root is when we start the game ?

Biosquid:
my bad didnt see your comment, im not completely sure but even if it was the case in other points you still never wanna delete the only scene you are on because you would softlock regardless right? I guess in the future we may wanna change it to where it brings you back to the last scene you are on but depending on how we order the scenes it shouldn't be a problem

Turtyo:
Yeah you are right, even if there are other cases, we wouldn't want to delete the scene, so that's not a problem

Turtyo (approved):

JonaLam (approved):

PR #134, opened 2024-09-26
 Select a group of enemy when entering a combat depending on map position

### LINKED ISSUE #128: Select a group of enemy when entering a combat depending on map position
## Feature Description

We need to be able to select enemies for certain areas of the game, and to load some of their information (such as their sprite for example).

## Motivation

We currently only have two enemies in the game, which are the same for all combats, we need to have more for the game.

## Acceptance Criteria

- Able to select one or more enemies / a certain group of enemy for a given level of the map
- Have a list of all possible enemies 

## Proposed Solution

Make an enum with each enemy type inside global enums (we won't do it with strings, because it's always error prone).

We need some sort of table that lists possible group of enemies by level of the map. A simple solution would be a dictionary where a key is a certain group of enemy (represented as an array of enemies). The values could be which levels of the map we allow this group of enemy to appear on.
We might need a separate dictionary for mini-boss and bosses.

A question is "how do we represent the levels of the map ? We could do it in %, but it was already mentioned inside #79 to use bitmasks to code different parts of the tree, but since it is not implemented yet, the choice is still up in the air. Using % would allow it to be adaptable depending on the size of the map, as well as being easy to change if we want to change the range in which we want the group of enemy.

For example, we could have a lower bound and an upper bound %, and we can just change those if we want the group of enemy to be available in different parts of the map. We also need to say in which floor is a certain enemy available. I think a certain enemy should have fixed HP and capabilities, meaning we don't change those depending on the level inside the tree (as that would be hard for balancing later on).

We might need a new class for coding the part of the maps, or we can just make an array of values; an enemy available between 40 and 60% of the 3rd floor would be [3, 40, 60] for example. When going to check which enemy group to spawn, we go into this dictionary and check which groups are in the correct range. If we make a class, we can re-use it for other things, which is why I prefer this approach.

Another question to take into account would be if we allow re-using the same group of enemy multiple times, or if we add a boolean to say if we used this group of enemy already.

## Additional Context

This needs to be done before #127 because loading enemy decks needs to have enemies loaded. We have to create a new enemy scene for each enemy to change their sprite based on their name.

Turtyo:
# Description

Choose an enemy group when entering a new fight, depending on a given distribution of enemy groups

## Related issue(s)

Closes #128

## List of changes

- Add a function to choose an enemy group based on player position in the map
- Add an `EnemyGroup` which holds a list of enemies and their position in the scene (this is still not great, because there is no way to check for enemy position in the scene apart from launching the game, open to suggestion on how to handle positions)
- Add a background and a few enemies / enemy groups
- Save data if needed

## Tests

None yet, an idea for a test could be to launch the function to choose an enemy a few time and compare to an expected output

## Additional notes

The enemies are a bit strong now, you can use the instantWinDeck (at "res://Cards/CardSets/InstantWinDeck.tres") and use it by modifying the `_ready` of the `CardManager` (by switching out the default deck).

An idea regarding this deck switching, I found out you can use a `tscn` as an autoloaded singleton (which I did here). So we could make the default deck an `@export` and autoload a `tscn` file, so we could change the deck from the inspector to test, instead of having to change it in the code.


JonaLam (approved):
Good job, seems good

Core/Battler.gd
```diff
@@ -5,16 +5,19 @@ class_name Battler
...
 var _enemy_list: Array[Entity]
 var _enemy_action_list: Array[EnemyAction] = []
 
 func _ready() -> void:
-	_summon_enemies()
-	EnemyManager.enemy_list = _enemy_list
+	var enemy_group: EnemyGroup = EnemyManager.choose_enemy_group().instantiate()
+	add_child(enemy_group)
+	_summon_enemies(enemy_group)
+	EnemyManager.current_enemy_group.enemy_list = _enemy_list
```
Tomzkk, line 20: Is there any point in assigning the enemy_list here ?

Managers/EnemyManager.gd
```diff
@@ -2,7 +2,119 @@ extends Node
...
+func _choose_enemy_group_inner() -> PackedScene:
+	if current_enemy_group != null:
+		return current_enemy_group_packed
+	if DebugVar.DEBUG_USE_TEST_ENEMY_GROUP:
+		return load("res://Entity/Enemy/EnemyGroup/test_group.tscn") # return a test group
+	var height_percent_position: float = MapManager.get_map_percent_with_player_position()
+	if (current_lower_bound <= height_percent_position && current_higher_bound > height_percent_position):
+		# meaning we stayed in the same sub-section of the map, keep the same array of enemies
+		if shuffled_enemy_group_array.is_empty():
+			shuffled_enemy_group_array = enemy_group_array
```
Tomzkk, line 73: I believe this does not work as intended -> shuffled_enemy_group_array = enemy_group_array probably causes it to still refer to the same array, which means pop_back will remove the element from 'both'. 

Every time we reach this point, the enemy_group_array is reduced in size and eventually reaches 0 resulting in a game crash (as we return null from this function and try to instantiate it)

Tomzkk (commented):
Implementation wise looks okay, but there's an error which causes the game to crash. Details in one of the comments. Once that is resolved we should be good

Managers/EnemyManager.gd
Tomzkk: Running game with breakpoint in enemy manager, enemy group array sizes:
image
First combat encounter ^
image
Second combat encounter ^
image
Third combat encounter ^
Now the enemy group is array of size 0 and function returns null, resulting in crash

Core/Battler.gd
Turtyo: well first of all it's incorrect and I corrected it in 7593c00238bc

And it's to keep track of the current list of enemy, since you need to instantiate the enemy group to get access to the enemy list (which the enemy manager doesn't do)
I think even using the `_enemy_list` of enemy battler is just a way to not have to go search for it inside the `EnemyManager` each time

Managers/EnemyManager.gd
Turtyo: You are totally correct about this,, I often forget that array are passed by reference, solved this in dad9ad285d4c

Managers/EnemyManager.gd
```diff
@@ -2,7 +2,119 @@ extends Node
...
+			shuffled_enemy_group_array.shuffle()
+
+		return shuffled_enemy_group_array.pop_back()
+	else:
+		# we entered a new sub-section, search the corresponding array of enemy group
+		for key: Array[int] in enemy_group_distribution:
+			# include lower bound, exclude higher bound
+			if (key[1] <= height_percent_position && key[2] > height_percent_position):
+				enemy_group_array = enemy_group_distribution[key]
+				shuffled_enemy_group_array = enemy_group_array
```
Tomzkk, line 83: Don't we have potential same issue here as we had above ? 
enemy_group_array = enemy_group_distribution[key]
- would this assign the group by reference or as a new array ? Because then we are assigning it to shuffled_enemy_group_array and popping back from it. 
Does this work as intended or would it immediately remove one enemy group from the array when it's chosen ?

Tomzkk: There is a bug in map position calculation so actually we never went into this Else. What I wrote above probably applies and needs to be checked after the issue gets resolved

Managers/MapManager.gd
```diff
@@ -292,3 +292,11 @@ func debug_print_event_count() -> void:
...
 		print("Event " + k + " has " + str(events[k]) + " rooms (expected: "+ str(float(expected_probabilities[k] * total_nb_rooms/100.0)).pad_decimals(2) +"). (The percentage is " + str(float(events[k]) * 100 / total_nb_rooms).pad_decimals(2) + "%, expected: " + str(expected_probabilities[k]) + "%)")
 	print("Total number of rooms generated: " + str(total_nb_rooms))
+	
+
+## Returns the percent height of the player in the current map
+func get_map_percent_with_player_position() -> float:
+	var player_position: Vector2i = PlayerManager.player_position
+	var number_of_floors: int = map_width_array.size()
+	var player_y_floor: int = player_position.y
+	return (player_y_floor/number_of_floors) * 100
```
Tomzkk, line 302: (player_y_floor/number_of_floors) is always returning 0
Both variables needs to be floats insetad of int (or cast to floats at division)

Tomzkk (commented):
I found a new batch of issues. Some of them are in the comments, but even after I tried to fix them locally myself, I ended up running in more issues down the line so I'm just leaving this one.
Even after fixing the get_map_percent_with_player_position() on my end and testing, it seems the else statement in the _choose_enemy_group_inner doesn't work, mostly related to data types not being the same as what we expect (or at least godot thinking so). 

for key: Array[int] in enemy_group_distribution: is returning an error for me-> it has to just be for key: Array .... without specifying the int type otherwise godot complains. 
Even after fixing that, it seems there's more type mismatches down the line, for example enemy_group_array = enemy_group_distribution[key] failed on my end with godot complaining that I tried to assign Array into a variable of type Array[PackedScene]. 

I am not fully sure if (some of) these are just on my end as I customized the enemy_Group_distribution and might have done it wrong so do test on your own end

Core/Battler.gd
Turtyo: I think you are right, it's mainly an artifact from the time we didn't have the enemy manager and only the battler i think. I'll remove it, it will prevent confusion

Turtyo: After checking, it's actually used 11 times, which is starting to be a bit more. Maybe it makes sense to keep it then, instead of replacing everything by `EnemyManager.current_enemy_group.enemy_list`

Managers/EnemyManager.gd
Turtyo: yes it would cause the same problem, fixed in 4995df761b3c

Managers/MapManager.gd
Turtyo: mmh that's strange, i'll look into it

Turtyo: Fixed in 62734436aac9

Turtyo: I was also thinking we could do `(player_y_floor *100 / number_of_floors as float)` instead of casting both to float

Turtyo:
I need to fix a few things in that `else` branch of the `EnemyManager`
At times like this, I really wish Godot had static type analysis (and not just static type annotation) and not me needing to execute the code to know if the types are correct or not...

Turtyo:
Okay e2efa2e7576f should fix that `else` branch
I added a check to make sure the new list of enemy groups is not empty (and for now that second list past 20% is indeed empty); so it loads the test group as a default. I'll add the cauldron alone on that second part, it will change later anyway

Managers/EnemyManager.gd
```diff
@@ -2,7 +2,122 @@ extends Node
...
+## - Check if the player is in the same sub-section as the previous time this function was called [br]
+## - Choose an array of enemy group depending on player position if either sub-section changed, or the array to choose from is empty [br]
+## - Shuffle it, pop the last element of that array
+func _choose_enemy_group_inner() -> PackedScene:
+	if current_enemy_group != null:
+		return current_enemy_group_packed
+	if DebugVar.DEBUG_USE_TEST_ENEMY_GROUP:
+		return load("res://Entity/Enemy/EnemyGroup/test_group.tscn") # return a test group
+	var height_percent_position: float = MapManager.get_map_percent_with_player_position()
+	if (current_lower_bound <= height_percent_position && current_higher_bound > height_percent_position):
```
Tomzkk, line 70: I believe the current_lower_bound and current_higher_bound is never actually updated so we actually never go into the first IF after we get past 20% height

Managers/EnemyManager.gd
```diff
@@ -2,7 +2,122 @@ extends Node
...
+func choose_enemy_group() -> PackedScene:
+	var packed_enemy_group: PackedScene = _choose_enemy_group_inner()
+	current_enemy_group_packed = packed_enemy_group
+	return packed_enemy_group
+
+## Chooses an enemy group in the following way: [br]
+## - Check if the player is in the same sub-section as the previous time this function was called [br]
+## - Choose an array of enemy group depending on player position if either sub-section changed, or the array to choose from is empty [br]
+## - Shuffle it, pop the last element of that array
+func _choose_enemy_group_inner() -> PackedScene:
```
Tomzkk, line 64: Loading of the game is currently broken - if you exit the game on a combat node and try to reload it crashes 
(1. enter combat, 2. exit game completely, 3. start again, 4. continue game -> Cannot call instantiate on a null object)
The current_enemy_group gets loaded, but current_enemy_group_packed is null and it gets returned from the function

Tomzkk (commented):
There's two more issues left that I could find, both described in comments

Managers/EnemyManager.gd
Turtyo: Fixed with 6f33aab92219

Managers/EnemyManager.gd
Turtyo: Fixed with 162bd734e2da
I forgot a lot of little things this PR

Managers/EnemyManager.gd
```diff
@@ -2,7 +2,131 @@ extends Node
...
+		if shuffled_enemy_group_array.is_empty():
+			shuffled_enemy_group_array = enemy_group_array.duplicate()
+			shuffled_enemy_group_array.shuffle()
+
+		return shuffled_enemy_group_array.pop_back()
+	else:
+		# we entered a new sub-section, search the corresponding array of enemy group
+		for key: Array in enemy_group_distribution:
+			# include lower bound, exclude higher bound
+			if (key[1] <= height_percent_position && key[2] > height_percent_position):
```
Tomzkk, line 84: One thing came to mind but I am not sure if it should be part of this PR as we don't even have other floors implemented yet as far as I know, but currently this is only checking for the height bounds, not which floor we're on (key[0])
If we don't have floors implemented yet in any way then it probably can't really be added now - or can we have some placeholder at least so that it's not forgotten ?

Tomzkk (approved):
Seems to be working properly now, just some minor comments left to address

Managers/MapManager.gd
Tomzkk: That could also work - there would however be a difference in cases where we hit decimals like 19.853% would get rounded to 20 and fall into 20-40 distribution, whereas with current Float casting it would fall under 0-20.

Managers/EnemyManager.gd
```diff
@@ -53,6 +53,9 @@ func _ready() -> void:
 ## This calls the inner function that actually chooses the enemy group [br]
 ## This is done to save the packed scene in this EnemyManger before it is returned
 func choose_enemy_group() -> PackedScene:
+	if current_enemy_group_packed != null:
```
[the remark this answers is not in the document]
Turtyo: technically once we check `current_enemy_group` and the other time `current_enemy_group_packed`
it's fine to do the test twice I think

Managers/EnemyManager.gd
Turtyo: We check the section we are one, because we do `for key in enemy_group_distribution` and later `var new_enemy_group_array: Array = enemy_group_distribution[key]`
Since the key is of the shape `[floor, lower_bound, upper_bound]`, we check the section we are on

Managers/EnemyManager.gd
```diff
@@ -2,7 +2,131 @@ extends Node
...
+		for key: Array in enemy_group_distribution:
+			# include lower bound, exclude higher bound
+			if (key[1] <= height_percent_position && key[2] > height_percent_position):
+				# update lower and higher bound
+				current_lower_bound = key[1]
+				current_higher_bound = key[2]
+				var new_enemy_group_array: Array = enemy_group_distribution[key]
+				if new_enemy_group_array.is_empty():
+					break # go to load the test group
+				enemy_group_array.assign(new_enemy_group_array) # assign performs a copy
```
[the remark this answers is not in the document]
Turtyo: We don't need to deep copy the array, since we are not modifying the elements of the array (only instantiating them); we are only modifying the arrays themselves.
Indeed, I thought assign would perform a copy, but seems like it does not. It doesn't really matter if we do a copy of `enemy_group_array` either, since we do not modify this array; I can leave a comment for the future in case we need it

Managers/EnemyManager.gd
Tomzkk: I know that there is support for having floors in the distribution, just that we are not actually checking for the floors in any way within the function - although as they are not yet implemented best we could do would probably just be some sort of placeholder for now anyway so it shouldn't really matter

Managers/EnemyManager.gd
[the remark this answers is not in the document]
Tomzkk: One should not exist without the other - in fact if it does things would break like they did before when _packed was not being saved. It's effectively check for the same thing and for the same reason. I don't really see reason to keep both

Tomzkk (approved):

JonaLam (approved):

Biosquid (approved):

Issue #112, opened 2024-06-25
Make cards


Turtyo:
## Is your amelioration related to a problem? Please describe.

Most of the cards we have are there for testing purpose, and we don't have many actual cards that have been voted.

## Describe the solution you'd like

Start to implement some cards from our card list. We don't need to do all of them right now, but we should do enough to cover most of the functionalities, to make sure we have everything setup correctly.
This will require making some new effects for sure, and some new status. We might discover we are missing things in terms of infrastructure.

## Additional context
Card resources should be separated between testing and actual cards to make sure to not confuse them.

What cards should we choose to implement right now ?

Turtyo:
I think we should a list of cards from the card doc and list them here. We can discuss the different effects and targeting needed for each cards.
There will be repeats, actually not a bad thing, less effects to implement.

The main difficulty in this issue is to implement the effects and targeting, actually making the cards is just putting everything together. Might take some time but won't be difficult.

I'm also thinking we can use this issue to add the actual borders we plan to use for the different cards.

Tysterman74:
Doing an initial pass through the cards that were approved and the effects we currently have, I think we can divide and conquer with the following:

These cards can be made without additions to Buffs/Debuffs:
- Slap 
- Double Slap
- Banana 
- Sapling Rap 
- Ara-Ara
- Hanya?
- Uuuuu
- Chu~
- Oops Oops
- Wrench
- Whispers
- Hurricane
- Ear Pick
- Mic Bonk
- Sneeze
- Slime Girl
- Wink
- Headbutt Attack
- Motivational Quote
- Cave Echo
- Cook
- Recycle Hand
- Ancient Meme
- Kalimba Hammer
- Let Her Cook
- Too Good To Use
- Manifesto
- Treastise
- Full Send
- Sneak Attack
- Sweep
- Gaslight
- "Fun" Facts
- (Vampiric) Bite
- Heart Hands
- Green Dart Gun
- Blue Dart Gun
- Reload Me
- UwU
- pspspsps
- Backwards Driving
- Give Me the Money
- Speedrun
- Vampirism
- Fauna Freeze Ray

These cards don't have any description:
- Idol Step
- Penlight
- Idol Meeting
- French Toast <Fauna Lore>
- Pink Wool
- Jokirin
- Pass a Fist
- Ooh!
- White Wool
- Orange Wool
- You Dont Know Me

Need to implement Comfy Buff for the following cards:
- Blessed Giggles
- Hug
[...]

PR #138, opened 2024-12-06
Issue-112: Implement Cards

Tysterman74:
# Description

Added Cards and Card Art, refactored Battle scene to adjust for new art.

## Related issue(s)

Related to #112 

## List of changes

Added the following cards and their associated artwork:
- Slap
- Double Slap
- Banana
- Broom
- Ara-Ara
- Uuuuu
- Wrench
- Whispers
- Mic Bonk
- Sneeze
- Wink
- Headbutt Attack
- Ancient Meme
- Sweep
- Gaslight
- "Fun" Facts
- Stealth Crouch
- Take Cover
- (Vampiric) Bite
- Heart Hands
- Green Dart Gun
- Blue Dart Gun
- pspspsps
- Backwards Driving

Other Changes:
- Updated Card Container to fit the new Card Art
- Updated the Card packed Scene with the new Card Art, changed the heirarchy to have a child for each card type in order to freely modify the text placement on each card type since the art is different
- Updated the CardScrollUI to fit the new Card Art

## Tests

N/A

## Additional notes

N/A


UI/CardScrollUI.gd
```diff
@@ -42,12 +45,17 @@ func populate(parent_name: String) -> void:
...
-		cardui.anchors_preset = CORNER_TOP_LEFT
-		cardui.position = CARDUI_POS
+		#cardworld.custom_minimum_size = (cardui.size * CARD_SCALE) + SIZE_OFFSET
+		
+		#cardui.scale = CARD_SCALE
+		#cardui.anchors_preset = CORNER_TOP_LEFT
+		#cardui.position = CARDUI_POS
+		
+		#key_art.scale = KEY_ART_SCALE
+		#key_art.position = cardui.position - Vector2(50, 130)
```
Turtyo, line 58: is this still needed ?

UI/CardUISetter.gd
```diff
@@ -5,12 +5,10 @@ class_name CardUISetter
...
 
-
-@export var title_label: Label
-@export var description_label: Label
-@export var energy_label: Label
 @export var key_art: TextureRect
-
+@export var attack_ui: Control
+@export var skill_ui: Control
+@export var power_ui: Control
```
Turtyo, line 12: so you replace all with a specific UI for each type of card then ?

Turtyo (changes requested):
Something I notice when playing the game:
- I think the cards are bit too big
- When hovering a card, the bottom of it disappears, meaning we can't read the name of the artist
- maybe there should be a default image when no image is available for the card, like a "UUUU we forgot to put the art here"

Turtyo:
Also you moved cards used in Tests to `TestingCards` but you didn't update the path in the test, could you modify that please ?

UI/CardUISetter.gd
Tysterman74: Yes, Since the card art for Attack, Skill, and Power varies slightly it was pretty tough trying to just use one Control to set the text and have it be aligned properly. So I created three separate Control objects for each Card Type so I can set the text individually for each type. 

It does mean if we wanted to change the location of a text, I'd have to do it three times for each card type but seeing as how we're only gonna have three anyways I think we can handle that burden.

UI/CardScrollUI.gd
Tysterman74: Oop it's not, I removed.

UI/CardUISetter.gd
Turtyo: If it's only three times, and it's better to fit everything like that, then that's a good thing

Turtyo (approved):

Issue #139, opened 2024-12-10
Add music to the game


Turtyo:
## Feature Description

Use the audio tracks made by the music inside the game.

## Motivation

It will *sound* way better like that !

## Acceptance Criteria

- Each area as its related music (see below for list)
- Music loops inside a given level if it ends too early
- Music changes after a while

## Proposed Solution

I would create a global sound reader (I'm not sure of the name inside Godot). The idea behind making it global is that you don't need to recreate it and destroy it between scenes, which simplifies handling transitions.
There is a parameter to automatically loop the music on the music player inside Godot;
We can cut the map in three and play one of the three area music for each section.
So when you move, check the Y coordinate (aka the floor) to see if we are at more than 1/3 and more than 2/3.
So music are for specific areas, like boss music, shop music, heal music...

## Implementation Steps (Optional)

Everything can be done at the same time I think.

## Additional Context
Here are the musics and the area I think we should use them in:
0 to 33% of the map: Battle area 1.wav
33 to 66% of the map: Battle area 2.wav
66 to 100% of the map: Fauna_floor_3_theme_loop.wav
Boss fight: Boss 1.wav
Shop: Snail_shop_loop.wav
Heal: Sweet_Snail_Healing_loop.wav

Even though shop and heal are either being worked on or not done yet, we can still put the music for those, because we don't need to modify the event. Just need to check the event type when entering a new room (I think there is somewhere that already prints the event type, with a debug var, so we can check that to see how we did it and just copy that)

The boss fight is a normal mob fight, the way to see it's a boss fight is just that it's the last floor of the map (and it's probably the way we will do to decide to spawn a boss, just see that it's the last floor, so it makes sense to do it like that too).

PR #140, opened 2024-12-11
Issue-139: Add Music

Tysterman74:
# Description

Add music and SFX to game

## Related issue(s)

#139 

## List of changes

Added the following tracks
- area_1.wav
- area_2.wav
- area_3.wav
- boss_1.wav
- heal.wav
- shop.wav

Other changes:
- Added AudioManager and added it to Autoloader
- Play heal.wav when entering Heal Event
- Player shop.wav when entering Shop Event
- When entering Mob Event, calculate the percentage complete the player is on the Map and play either area_1, area_2, or area_3
- When reaching the top floor, play boss_1

## Tests

N/A

## Additional notes

N/A


#Scenes/SceneScripts/TestingScene.gd
```diff
@@ -5,5 +5,25 @@ extends Node2D
...
 	SaveManager.execute_save()
 	PhaseManager.initialize_game()
+
+func _play_music() -> void:
+	var number_of_floors: float = MapManager.get_floors()
+	
+	# Offset by 1 since it's reading from an array
+	var current_player_floor: float = PlayerManager.player_position.y + 1
+	
+	var percentage_complete: float = (current_player_floor / number_of_floors) * 100
```
Turtyo, line 18: you can use `MapManager.get_map_percent_with_player_position` instead to be consistent

Turtyo (changes requested):
No issue with the code as is, but I noticed some issues when playing the game:
- There is no sound if I start with something else than a mob fight (in the case I tested, with a dialogue at the start, no music is playing)
- The sound is game is not the same balance as the sound of the original music, it sounds like the music is played stronger on the left than the right ear (mostly noticeable for the area 1 music, but it's present on all musics)

Turtyo (changes requested):
No problem anymore, apart from the fact I hear the music louder in my left ear than my right ear still.
I think this might be the reason : https://forum.godotengine.org/t/why-is-audio-only-played-from-the-left-earphone/21639
Basically, the 2D audio stream plays sounds relative to the camera, meaning that if we want it to be the same on both ears, we need to instantiate the AudioManager at the center of the screen.

I have two ideas for that:
- either see if you can move the 2D audio stream with code from the AudioManager
- if that's not possible, then create an instance of the AudioManager (ie a scene with the script of the AudioManager) and move it from there

I think the first solution would be better, as it's less heavy

Turtyo (approved):

PR #144, opened 2024-12-26
Enemy attack pattern

### LINKED ISSUE #127: Enemy attack pattern
## Feature Description

Provide each enemy with a customizable set of attacks and actions.

## Motivation

The enemy attack is currently only a single attack with one damage, the same for everyone. This is neither interesting nor practical to edit.

## Acceptance Criteria

- Each enemy can have it's own attack set
- We can easily change an enemy attack pattern without having to re-code an enemy
- If the player can do it, the enemy can do it (for most attacks)
- Enemy attacks are generally random

## Proposed Solution

Use the current card system to give each enemy a set of card which is its attack pattern. Changing the proportion of each card naturally changes an enemy amount of each attack.
This way, adding a new attack to an enemy means adding a new card to a deck (which is easy to do with the current card system, using the inspector inside Godot).
This way, every time we add an effect to an enemy or the player, it becomes available to the other (so it's re-usable).

- Add *something* (to be discussed) to keep track of which deck of card for each enemy (a dictionary ?)
- When loading an enemy, load its attack deck too
- Will need to change the battler to separate player drawing with enemy drawing (since currently, drawing draws every deck, since we have only one it was not a problem, but now it is). Can use the new phase system to separate the phases.
- Make some decks for a few enemies (can add the enemies that are finished by the art team)

## Additional Context
We might add a functionality to show the next enemy attack later on. With a deck, we just need to peek at the next card in the deck. This is not to be done in this implementation, but keep it in mind.

Turtyo:
# Description

Add a way to build the deck / AI of an enemy

## Related issue(s)

Closes #127

## List of changes

- Add EnemyActionTree to choose the next action of the enemy
- Use a dictionary of probability given previous action to know next action
- Change directory structure to separate player and enemy actions (as well as cards that work for both)

Last change is due to targeting, you can have the exact same card, but one is targeting the enemy team, and one is targeting the player. The original choice was that it was easier to not have to switch sides every time (meaning "enemies" is a global truth, doesn't depend on who is attacking). It's easier to understand, but it means some cards will exist in double.

## Tests

Tests have been corrected to account for the new structure

## Additional notes

Once enemy cards have been given an enemy action name (see global enums), the order of the variants of the enum shouldn't be changed. Since this is simply a value, if we change the order in the enum, the order in the `.tres` of the cards will stay the same, meaning it will be wrong

TL;DR : don't change the order of the variants of the EnemyAction enum, you can add more though.

I'll add some demo decks for enemies tomorrow I think.

### DIFF
--- a/Entity/Enemy/ActionTrees/golden_cauldron_action_tree.tres
+++ b/Entity/Enemy/ActionTrees/golden_cauldron_action_tree.tres
@@ -5,12 +5,12 @@
 [ext_resource type="Script" path="res://Entity/Enemy/NextCardProbability.gd" id="2_5h3sb"]
 [ext_resource type="Resource" uid="uid://dluk3gyluaxe6" path="res://Cards/Resource/Both/Card_Big_Block.tres" id="3_dpwwb"]
 
-[sub_resource type="Resource" id="Resource_io36s"]
+[sub_resource type="Resource" id="Resource_nj3wp"]
 script = ExtResource("2_5h3sb")
 probability = 100
 card = ExtResource("1_7ay63")
 
-[sub_resource type="Resource" id="Resource_2cubn"]
+[sub_resource type="Resource" id="Resource_hfiqb"]
 script = ExtResource("2_5h3sb")
 probability = 100
 card = ExtResource("3_dpwwb")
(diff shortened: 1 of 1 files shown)


Turtyo:
Mmh, upstream for Godot still down, test pass in local though, feel free to run them on your end too just to be safe

Issue #146, opened 2024-12-27
Heal event interface


Turtyo:
## Feature Description

Make a scene for the heal event

## Motivation

There is currently nothing in this event, it's just empty, you can't heal.

## Concept art / Designs

There is currently no background for the heal event, but it's being worked on.

## Acceptance Criteria

- The player should be healed by a certain % of their max HP upon entering the room
- The player can remove one card from its deck before leaving (if they wish to do so)
- The player can also decide to just leave the room

## Proposed Solution

Check event type to see if its heal when entering a new event, in that case heal player. Have a button that opens the view card interface (probably needs to be fixed with #145 first)
I think in that interface, cards are built as buttons, so the way to remove cards is just to make enable the button being clickable. A confirmation window should pop up to make sure it's not a miss-click.

## Implementation Steps (Optional)

- A first PR with just the auto-heal when entering the room
- A second PR with the card removal

If we just want to make the game playable, only the healing will do the trick, so that's what should be done and merged ASAP. Since the second part depends on a bugfix, it's better to keep it separate (also it's a bit more complex).

## Additional Context

[Provide any additional context, background information, or related resources that might be helpful for understanding the feature.]

Turtyo:
Will quickly do the first part, that way we'll at least have something

PR #147, opened 2024-12-29
Heal event interface

### LINKED ISSUE #97: Change the way event skip button works to spawn them on any scene

Turtyo:
# Description

Add an actual heal event, where the player gets healed for 25% of their max HP

## Related issue(s)

In link with #146 but doesn't close it

## List of changes

- Add UI for heal event
- Add a player scene with just the nodes, no display (for easy add when needed in scenes that need the Player object without the player visual)

## Tests

Removed the `naming_convention_test` as I think it's not flexible at all. When I first wrote the test, the project was way smaller, but the naming convention is too rigid and also doesn't provide much value.

## Additional notes

The card removal in the heal event mentioned in #146 should be in another PR


Tomzkk (changes requested):
The event does not end after healing, and it is possible to press the heal button multiple times and heal to full

Tomzkk (approved):

